PR #8, opened 2021-08-16
add support for the cluster placement api

gparvin:
New placement api https://github.com/open-cluster-management-io/community/issues/63


gparvin:
Note, you can use Placement without installing the Application components, but you do need to install the `PlacementRule` CRD.  I have not investigated removing this dependency since I feel it's minor, but other feedback is welcome.

pkg/apis/cluster/v1alpha1/placement_types.go
```diff
@@ -0,0 +1,48 @@
+package v1alpha1
```
mprahl, line 1: Are these defined in a shared library that we could just import instead of redefining it for this controller?

pkg/apis/cluster/v1alpha1/placementdecision_types.go
```diff
@@ -0,0 +1,55 @@
...
+// +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
+
+// PlacementDecision is the Schema for the placementdecisions API
+// +kubebuilder:subresource:status
+// +kubebuilder:resource:path=placementdecisions,scope=Namespaced
+type PlacementDecision struct {
+	metav1.TypeMeta   `json:",inline"`
+	metav1.ObjectMeta `json:"metadata,omitempty"`
+
+	Spec   PlacementDecisionSpec   `json:"spec,omitempty"`
```
mprahl, line 40: Optional: Since this is just an empty struct, could you remove it and delete the struct definition?

My philosophy is that we shouldn't keep "dead" code around. We can always add it back if needed.

pkg/controller/propagator/placementDecisionMapper.go
```diff
@@ -0,0 +1,61 @@
...
+	client.Client
+}
+
+func (mapper *placementDecisionMapper) Map(obj handler.MapObject) []reconcile.Request {
+	object := obj.Meta
+	log.Info("Reconcile Request for PlacementDecision %s in namespace %s", object.GetName(), object.GetNamespace())
+
+	// get the placement name from the placementdecision
+	placementName := object.GetLabels()["cluster.open-cluster-management.io/placement"]
+	if len(placementName) == 0 {
```
mprahl, line 27: Optional: A bit more intuitive to read
```suggestion
	if placementName == "" {
```

pkg/controller/propagator/placementDecisionMapper.go
```diff
@@ -0,0 +1,61 @@
...
+		return nil
+	}
+
+	pbList := &policiesv1.PlacementBindingList{}
+	// find pb in the same namespace of placementrule
+	err := mapper.List(context.TODO(), pbList, &client.ListOptions{Namespace: object.GetNamespace()})
```
mprahl, line 33: Could you also add a field selector for at least the PlacementRef name? Filtering on the server side will reduce network IO consumption and should be faster.

pkg/controller/propagator/placementDecisionMapper.go
```diff
@@ -0,0 +1,61 @@
...
+	}
+
+	var result []reconcile.Request
+	// loop through pb to find if current placement is used for policy
+	for _, pb := range pbList.Items {
+		// found matching placement rule in pb
+		if pb.PlacementRef.APIGroup == clusterv1alpha1.SchemeGroupVersion.Group &&
+			pb.PlacementRef.Kind == clusterv1alpha1.Kind && pb.PlacementRef.Name == placementName {
+			// check if it is for policy
+			subjects := pb.Subjects
```
mprahl, line 45: Optional: If the if statement checked the opposite condition, you could specify `continue` and dedent the rest of the code. This makes it easier to read since the reader does not need to remember the context of the previous if statement and the code is less indented.

pkg/controller/propagator/placementDecisionMapper.go
```diff
@@ -0,0 +1,61 @@
...
+			for _, subject := range subjects {
+				if subject.APIGroup == policiesv1.SchemeGroupVersion.Group && subject.Kind == policiesv1.Kind {
```
mprahl, line 47: Optional: If the if statement checked the opposite condition, you could specify `continue` and dedent the rest of the code. This makes it easier to read since the reader does not need to remember the context of the previous if statement and the code is less indented.

pkg/controller/propagator/placementDecisionMapper.go
```diff
@@ -0,0 +1,61 @@
...
+					log.Info("Found reconciliation request from placement rule...", "Namespace", object.GetNamespace(),
```
mprahl, line 48:
```suggestion
					log.Info("Found reconciliation request from placement decision...", "Namespace", object.GetNamespace(),
```

pkg/controller/propagator/propagation.go
```diff
@@ -210,6 +201,89 @@ func (r *ReconcilePolicy) handleRootPolicy(instance *policiesv1.Policy) error {
...
 }
 
+// getApplicationPlacementDecisions return the placement decisions from an application
+// lifecycle placementrule
+func getApplicationPlacementDecisions(c client.Client, pb policiesv1.PlacementBinding, instance *policiesv1.Policy) ([]appsv1.PlacementDecision, *policiesv1.Placement, error) {
+	plr := &appsv1.PlacementRule{}
+	err := c.Get(context.TODO(), types.NamespacedName{Namespace: instance.GetNamespace(),
+		Name: pb.PlacementRef.Name}, plr)
+	if err != nil && !errors.IsNotFound(err) {
+		log.Error(err, "Failed to get plr...", "Namespace", instance.GetNamespace(), "Name",
```
mprahl, line 211:
```suggestion
		log.Error(err, "Failed to get PlacementRule..", "Namespace", instance.GetNamespace(), "Name",
```

pkg/controller/propagator/propagation.go
```diff
@@ -210,6 +201,89 @@ func (r *ReconcilePolicy) handleRootPolicy(instance *policiesv1.Policy) error {
...
+}
+
+// getClusterPlacementDecisions return the placement decisions from cluster
+// placement decisions
+func getClusterPlacementDecisions(c client.Client, pb policiesv1.PlacementBinding, instance *policiesv1.Policy) ([]appsv1.PlacementDecision, *policiesv1.Placement, error) {
+	plr := &clusterv1alpha1.Placement{}
+	err := c.Get(context.TODO(), types.NamespacedName{Namespace: instance.GetNamespace(),
+		Name: pb.PlacementRef.Name}, plr)
+	if err != nil && !errors.IsNotFound(err) {
+		log.Error(err, "Failed to get plr...", "Namespace", instance.GetNamespace(), "Name",
```
mprahl, line 231:
```suggestion
		log.Error(err, "Failed to get Placement...", "Namespace", instance.GetNamespace(), "Name",
```

pkg/controller/propagator/propagation.go
```diff
@@ -210,6 +201,89 @@ func (r *ReconcilePolicy) handleRootPolicy(instance *policiesv1.Policy) error {
...
+			pb.PlacementRef.Name)
+		return nil, nil, err
+	}
+	// plr found, add current plcmnt to placement
+	placement := &policiesv1.Placement{
+		PlacementBinding: pb.GetName(),
+		PlacementRule:    plr.GetName(),
+		// Decisions:        plr.Status.Decisions,
```
mprahl, line 219:
```suggestion
```

pkg/controller/propagator/propagation.go
```diff
@@ -210,6 +201,89 @@ func (r *ReconcilePolicy) handleRootPolicy(instance *policiesv1.Policy) error {
...
+		Name: pb.PlacementRef.Name}, plr)
+	if err != nil && !errors.IsNotFound(err) {
+		log.Error(err, "Failed to get plr...", "Namespace", instance.GetNamespace(), "Name",
+			pb.PlacementRef.Name)
+		return nil, nil, err
+	}
+	// plr found, add current plcmnt to placement
+	placement := &policiesv1.Placement{
+		PlacementBinding: pb.GetName(),
+		PlacementRule:    plr.GetName(),
```
mprahl, line 218: This code is reached if it wasn't found because of `if err != nil && !errors.IsNotFound(err) {`. Was that intentional?

pkg/controller/propagator/propagation.go
```diff
@@ -210,6 +201,89 @@ func (r *ReconcilePolicy) handleRootPolicy(instance *policiesv1.Policy) error {
...
+		// Decisions:        plr.Status.Decisions,
+	}
+	return plr.Status.Decisions, placement, nil
+}
+
+// getClusterPlacementDecisions return the placement decisions from cluster
+// placement decisions
+func getClusterPlacementDecisions(c client.Client, pb policiesv1.PlacementBinding, instance *policiesv1.Policy) ([]appsv1.PlacementDecision, *policiesv1.Placement, error) {
+	plr := &clusterv1alpha1.Placement{}
```
mprahl, line 227: Could you rename the `plr` variable? It implies it is a placement rule which is confusing to the reader.

pkg/controller/propagator/propagation.go
```diff
@@ -210,6 +201,89 @@ func (r *ReconcilePolicy) handleRootPolicy(instance *policiesv1.Policy) error {
...
+		Name: pb.PlacementRef.Name}, plr)
+	if err != nil && !errors.IsNotFound(err) {
+		log.Error(err, "Failed to get plr...", "Namespace", instance.GetNamespace(), "Name",
+			pb.PlacementRef.Name)
+		return nil, nil, err
+	}
+	// plr found, add current plcmnt to placement
+	placement := &policiesv1.Placement{
+		PlacementBinding: pb.GetName(),
+		Placement:        plr.GetName(),
```
mprahl, line 238: This code is reached if placement wasn't found because of `if err != nil && !errors.IsNotFound(err) {`. Was that intentional?

pkg/controller/propagator/propagation.go
```diff
@@ -210,6 +201,89 @@ func (r *ReconcilePolicy) handleRootPolicy(instance *policiesv1.Policy) error {
...
+		// Decisions:        plr.Status.Decisions,
+	}
+	list := &clusterv1alpha1.PlacementDecisionList{}
+	lopts := &client.ListOptions{Namespace: instance.GetNamespace()}
+
+	opts := client.MatchingLabels{"cluster.open-cluster-management.io/placement": plr.GetName()}
+	opts.ApplyToList(lopts)
+	err = c.List(context.TODO(), list, lopts)
+	if err != nil && !errors.IsNotFound(err) {
+		log.Error(err, "Failed to get plr...", "Namespace", instance.GetNamespace(), "Name",
```
mprahl, line 248:
```suggestion
		log.Error(err, "Failed to get PlacementDecision...", "Namespace", instance.GetNamespace(), "Name",
```

pkg/controller/propagator/propagation.go
```diff
@@ -210,6 +201,89 @@ func (r *ReconcilePolicy) handleRootPolicy(instance *policiesv1.Policy) error {
...
+		Placement:        plr.GetName(),
+		// Decisions:        plr.Status.Decisions,
+	}
+	list := &clusterv1alpha1.PlacementDecisionList{}
+	lopts := &client.ListOptions{Namespace: instance.GetNamespace()}
+
+	opts := client.MatchingLabels{"cluster.open-cluster-management.io/placement": plr.GetName()}
+	opts.ApplyToList(lopts)
+	err = c.List(context.TODO(), list, lopts)
+	if err != nil && !errors.IsNotFound(err) {
```
mprahl, line 247: I'm curious, have you verified if `List` returns a not found error if the API resource returned nothing? Normally, a REST API would not give you a 404 for instance. It'd just return an empty list, so this wasn't what I was expected is all.

pkg/controller/propagator/propagation.go
```diff
@@ -210,6 +201,89 @@ func (r *ReconcilePolicy) handleRootPolicy(instance *policiesv1.Policy) error {
...
+		log.Error(err, "Failed to get plr...", "Namespace", instance.GetNamespace(), "Name",
+			pb.PlacementRef.Name)
+		return nil, nil, err
+	}
+	var decisions []appsv1.PlacementDecision
+	decisions = make([]appsv1.PlacementDecision, 0, 100)
```
mprahl, line 253: Why specify a capacity when the amount is unknown? Is it really common to have 100+ placement decisions returned?

Perhaps, you could start with `len(list.Items)` or `len(list.Items) * 2` or no capacity at all and let Go handle that?

test/resources/case6_placement_propagation/case6-test-policy.yaml
```diff
@@ -0,0 +1,50 @@
+apiVersion: policy.open-cluster-management.io/v1
+kind: Policy
+metadata:
+  name: case6-test-policy
+spec:
+  remediationAction: inform
+  disabled: false
+  policy-templates:
+    - objectDefinition:
+        apiVersion: policies.ibm.com/v1alpha1
```
mprahl, line 10: Is this old API intentional?

test/utils/utils.go
```diff
@@ -33,6 +34,18 @@ func GeneratePlrStatus(clusters ...string) *appsv1.PlacementRuleStatus {
 	return &appsv1.PlacementRuleStatus{Decisions: plrDecision}
 }
 
+// GeneratePldStatus generate pld status with given clusters
+func GeneratePdStatus(placementName string, placementNamespace string, clusters ...string) *clusterv1alpha1.PlacementDecisionStatus {
```
mprahl, line 38:
```suggestion
func GeneratePldStatus(placementName string, placementNamespace string, clusters ...string) *clusterv1alpha1.PlacementDecisionStatus {
```

pkg/apis/cluster/v1alpha1/placement_types.go
```diff
@@ -0,0 +1,48 @@
...
+
+// Placement is the Schema for the placements API
+// +kubebuilder:subresource:status
+// +kubebuilder:resource:path=placements,scope=Namespaced
+type Placement struct {
+	metav1.TypeMeta   `json:",inline"`
+	metav1.ObjectMeta `json:"metadata,omitempty"`
+
+	Spec   PlacementSpec   `json:"spec,omitempty"`
+	Status PlacementStatus `json:"status,omitempty"`
```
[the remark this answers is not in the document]
gparvin: I tried to remove this code completely and instead use `open-cluster-management.io/api` but hit dependency issues.  I think it's due to our use of an old operator-sdk so I opened a jira to update the operator-sdk.  We would then need to revisit this.  Opened https://github.com/open-cluster-management/backlog/issues/15459 to help remember that.

pkg/apis/cluster/v1alpha1/placement_types.go
gparvin: Yes, but hit issues mentioned above when trying to use it.

pkg/controller/propagator/placementDecisionMapper.go
gparvin: Fixing

pkg/controller/propagator/placementDecisionMapper.go
gparvin: yes, will try this out

pkg/controller/propagator/placementDecisionMapper.go
gparvin: fixed, thanks

pkg/controller/propagator/placementDecisionMapper.go
gparvin: fixed

pkg/controller/propagator/placementDecisionMapper.go
gparvin: fixed

pkg/controller/propagator/propagation.go
gparvin: fixed

pkg/controller/propagator/propagation.go
gparvin: fixed

pkg/controller/propagator/propagation.go
gparvin: fixed

pkg/controller/propagator/propagation.go
gparvin: It doesn't seem right does it.  Trying a fix here too

pkg/controller/propagator/propagation.go
gparvin: fixed

pkg/controller/propagator/propagation.go
gparvin: trying the same fix here too

pkg/controller/propagator/propagation.go
gparvin: fixed

pkg/controller/propagator/propagation.go
gparvin: right, not sure about this

pkg/controller/propagator/propagation.go
gparvin: Right, I'll take that out and I expect we won't see any performance issues, but if we do then a comment about why a size is there will be added at that time.

test/resources/case6_placement_propagation/case6-test-policy.yaml
gparvin: It's a copy paste from the case 1 policy.  So it was intentional sort of.

test/utils/utils.go
gparvin: fixed

mprahl:
@gparvin is this ready for another round of review or are you still working on it?

gparvin:
I'm re-running the tests now.  I think the failure was just an API Server issue that's not related to the changes.  @mprahl please re-review while I do some local testing.

pkg/controller/propagator/propagation.go
mprahl: @gparvin what did you decide on this comment?

pkg/controller/propagator/propagation.go
```diff
@@ -210,6 +201,89 @@ func (r *ReconcilePolicy) handleRootPolicy(instance *policiesv1.Policy) error {
 	return nil
 }
 
+// getApplicationPlacementDecisions return the placement decisions from an application
+// lifecycle placementrule
+func getApplicationPlacementDecisions(c client.Client, pb policiesv1.PlacementBinding, instance *policiesv1.Policy) ([]appsv1.PlacementDecision, *policiesv1.Placement, error) {
+	plr := &appsv1.PlacementRule{}
+	err := c.Get(context.TODO(), types.NamespacedName{Namespace: instance.GetNamespace(),
+		Name: pb.PlacementRef.Name}, plr)
+	if err != nil && !errors.IsNotFound(err) {
```
mprahl, line 210: Did you decide on how to handle the case when the error is not found? For instance, using `plr.GetName()` doesn't seem right if `plr` is just `plr := &appsv1.PlacementRule{}`.

pkg/apis/cluster/v1alpha1/placementdecision_types.go
gparvin: This will be fixed when we address the operator-sdk update and issue mentioned previously.

pkg/controller/propagator/propagation.go
gparvin: I accidentally created all updated to propagation.go in an editor opened for a different directory so the changes were not really there when I requested a new review.  The changes are there now.

mprahl (approved):
Nice!

gparvin:
@mprahl Can you review this again?  Some of the tests were failing after your last review so I had to make some updates.

mprahl (approved):

PR #9, opened 2021-08-23
adjust actions to push an edge image

gparvin:
I'm going to learn what this does.  Goal is to eventually get an edge image.

### DIFF
--- a/.github/workflows/go-postsubmit.yml
+++ b/.github/workflows/go-postsubmit.yml
@@ -1,0 +1,40 @@
+name: GoPostSubmit
+
+on:
+  push:
+    branches:
+      - main
+      - release-*
+  workflow_dispatch: {}
+
+env:
+  # Common versions
+  GO_VERSION: '1.16'
+  GO_REQUIRED_MIN_VERSION: ''
+  GOPATH: '/home/runner/work/governance-policy-propagator/governance-policy-propagator/go'
+defaults:
+  run:
+    working-directory: go/src/open-cluster-management.io/governance-policy-propagator
+
+jobs:
+  verify:
+    name: images
+    runs-on: ubuntu-latest
+    steps:
+      - name: checkout code
+        uses: actions/checkout@v2
+        with:
+          fetch-depth: 1
+          path: go/src/open-cluster-management.io/governance-policy-propagator
+      - name: install Go
+        uses: actions/setup-go@v2
+        with:
+          go-version: ${{ env.GO_VERSION }}
+      - name: install imagebuilder
+        run: go install github.com/openshift/imagebuilder/cmd/imagebuilder@v1.2.1
+      - name: images
+        run: make images
+      - name: push
+        run: |
+          echo ${{ secrets.DOCKER_PASSWORD }} | docker login quay.io --username ${{ secrets.DOCKER_USER }} --password-stdin
+          docker push quay.io/open-cluster-management/governance-policy-propagator:edge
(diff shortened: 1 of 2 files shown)


gparvin:
@dhaiducek if I want an edge tagged image -- will this get me one?  Or is there more setup to do somewhere?

dhaiducek (commented):
This will be good. I think we should also:
- Run the image build (without publish) during the PR to make sure it works
- Update the deployment to point to the new image (instead of `latest-dev`)

Also, didn't we discuss using `community` instead of `edge` to not confuse it with the concept of edge devices? Minor thing--I don't think I have a strong opinion either way.

gparvin:
> @dhaiducek if I want an edge tagged image -- will this get me one? Or is there more setup to do somewhere?

I changed the tag in another PR.  My logic I think was that since I manually pushed an edge tagged image, it can be used until we get this PR completed.  The other PR is: https://github.com/open-cluster-management-io/governance-policy-propagator/pull/10

The other setup was creating the secret which has been done now.

dhaiducek:
Thanks! Should we also add a presubmit image build for completeness?

gparvin:
@dhaiducek Help me understand how the kind tests are currently run and if the post submit will break that and if the pre-submit needs to replace that.  I really don't know how all these pieces are hooking together -- some of that we can finalize in issue https://github.com/open-cluster-management/backlog/issues/15541.

.github/workflows/go-postsubmit.yml
```diff
@@ -0,0 +1,40 @@
...
+          fetch-depth: 1
+          path: go/src/open-cluster-management.io/governance-policy-propagator
+      - name: install Go
+        uses: actions/setup-go@v2
+        with:
+          go-version: ${{ env.GO_VERSION }}
+      - name: install imagebuilder
+        run: go install github.com/openshift/imagebuilder/cmd/imagebuilder@v1.2.1
+      - name: images
+        run: make images
```
dhaiducek, line 36: For our Makefile, this should be `make build-images`. And actually we can update the Makefile to default to `TAG ?= edge` instead of `TAG ?= latest`.

https://github.com/open-cluster-management-io/governance-policy-propagator/blob/2f29465c22e5f8c0ddc74b2b556e64968090cb86/Makefile#L144-L146

.github/workflows/go-postsubmit.yml
```diff
@@ -0,0 +1,40 @@
...
+        uses: actions/checkout@v2
+        with:
+          fetch-depth: 1
+          path: go/src/open-cluster-management.io/governance-policy-propagator
+      - name: install Go
+        uses: actions/setup-go@v2
+        with:
+          go-version: ${{ env.GO_VERSION }}
+      - name: install imagebuilder
+        run: go install github.com/openshift/imagebuilder/cmd/imagebuilder@v1.2.1
```
dhaiducek, line 34: We'll have to compare what we're doing to other repos. This step doesn't look familiar to me.

Although, it does make me wonder: we're using the DockerHub golang image, so if we do this too many times we might hit the pull quota. We might need to move to a Red Hat builder image (which may be what this is).

dhaiducek:
The pre- and post- submits are independent. My thinking is if we run an image build during our pre-submit and there are build issues we might be able to catch them there, before the PR is submitted. But maybe that's not necessary?

.github/workflows/go-postsubmit.yml
gparvin: Nice!  Fixed this.

.github/workflows/go-postsubmit.yml
gparvin: i copied this from the placement repo.

PR #14, opened 2021-11-05
Merge updates from ocm-plus repo

JustinKuli:
We're working towards developing on this upstream repository, but in the meantime we need to send some of the updates from where we're currently doing active development. This brings in many of the recent changes from https://github.com/open-cluster-management/governance-policy-propagator.

I spent a while debating / dealing with either rebasing or doing a merge commit. The merge will preserve both repos' histories, so I feel like it's a good compromise (a rebase ends up rewriting commits from one of the repos).


ycao56:
/lgtm
I assume we are gonna do rebase and merge.

JustinKuli:
@ycao56 I think so... I'm not sure what the commits will really look like after that, but I think it's our best option.

JustinKuli:
Ah, I can't rebase and merge because `This branch cannot be rebased due to conflicts`. Well, one option @mprahl was talking about was a force push - at least then the history would look like I expect...

I think I have permissions to do that, unless anyone is vehemently opposed?

JustinKuli:
@mprahl, the repo had merge commits disabled, I was able to re-enable it now though. I'm not sure if that means Github will make a merge commit on top of my merge commit?

If we have to (or decide to) force-push stuff anyway, I think your top choice would be the best bet. As we've discussed, I think it's a slightly improper to end up re-writing the commits from ocm-io, although I don't know if anyone would notice or actually mind. The issue with force-pushes is that anyone who has pulled from here would be out of sync, and couldn't just fast-forward.

mprahl:
> @mprahl, the repo had merge commits disabled, I was able to re-enable it now though. I'm not sure if that means Github will make a merge commit on top of my merge commit?
> 
> If we have to (or decide to) force-push stuff anyway, I think your top choice would be the best bet. As we've discussed, I think it's a slightly improper to end up re-writing the commits from ocm-io, although I don't know if anyone would notice or actually mind. The issue with force-pushes is that anyone who has pulled from here would be out of sync, and couldn't just fast-forward.

Yes, that is the concern, however, force pushing your branch as it is with the merge commit approach should not affect other contributors to the repo because your are only force pushing new commits and not rewriting the commit history. Doing the standard merge in the GitHub UI would create an additional merge commit which makes the Git history uglier but it's probably the "safest" approach and at least maintains the original commits.

PR #15, opened 2021-11-09
Add a release workflow

JustinKuli:
The workflow will trigger on any tag pushed that looks like `v*.*.*`.
The action is based off of the one in the registration repo, but using
the build process in this repo.


dhaiducek (approved):
LGTM!

PR #16, opened 2021-11-09
Add changelog for v0.5.0

JustinKuli:
Since this is the first release for this repo, I did not fill in an
actual changelog. If we want to (in the future) we could use this file
as an initial feature list... but I don't think it's fully necessary.

Mostly this is so the auto-generated release doesn't have a broken link.


dhaiducek (approved):
Works for me! I agree though--we should populate this in the future.

PR #20, opened 2021-12-14
Reconcile with ocm-plus

dhaiducek:


dhaiducek:
Shoot--I should have noticed that. I was wondering why there were so many commits. Looks like I might have to figure out a targeted approach.

dhaiducek:
That was way smoother. I'll have to double-check my other PRs to make sure I'm not creating duplicate commits...

mprahl:
> That was way smoother. I'll have to double-check my other PRs to make sure I'm not creating duplicate commits...

Is this ready for review? It's still marked as a draft.

dhaiducek:
Yes--I'll unmark it. There's a commit in the propagator that when it was merged wasn't signed with DCO that I think was causing the issue with merging the commits:
- https://github.com/open-cluster-management/governance-policy-propagator/pull/73

If we're cherry-picking commits then things will be fine, but since we had to rebase to merge it here originally and assuming my thinking is correct, it's going to continue to make these repos pretty much impossible to merge unless it's fixed in the midstream and merged again.

mprahl (approved):

PR #21, opened 2021-12-15
Update pkg references to `open-cluster-management.io`

dhaiducek:
Our package references were pointing to midstream, so between the org change as well as the fact that these references could eventually fall behind, it seems these should be updated.

I know very little about package references, so this is somewhat guesswork of what I could observe face-value and any insights would be appreciated.

- https://github.com/open-cluster-management/backlog/issues/18572
- https://github.com/open-cluster-management-io/community/issues/71


mprahl:
@JustinKuli could you please take this code review since I recall we had a discussion about this in one of your previous sync up PRs but I don't remember the outcome?

JustinKuli:
So yes, this looks generally good: we want to update the module paths in our repos from `github.com/open-cluster-management[-io]/...` to `open-cluster-management.io/...`. Based on the example PROJECT file at https://book.kubebuilder.io/reference/project-config.html, we can/should also update those URLs to the new module path.

When (before?) we do that, we will also need to make some website updates so that `go` knows where to find our module, since the module path will no longer equal the repository path. PRs like https://github.com/open-cluster-management-io/open-cluster-management-io.github.io/pull/169. More info on how `go` does that / what these pages do: https://go.dev/ref/mod#vcs-find

@dhaiducek I assume that last part is why this PR is just a draft?

## Motivation / Background

One of the main motivations of all of that is so that forks of the community repositories don't need a pile of changes in order to point their imports to their forks. Without updating the module name, every import would need to be adjusted to the fork repo URL, ***and*** each of those forks would need to set their module name in go.mod to their fork's location.

With the updated module name, forks that want to reference their forked repos can just use a `replace` directive in their `go.mod`. See: https://go.dev/ref/mod#go-mod-file-replace for some more details on that. Important to note is "if the replacement module has a go.mod file, its module directive must match the module path it replaces." which is why we can't just use a `replace` right now.

Eg, the community propagator repo (this repo) should import the community version of the api repo. And a fork of the propagator repo (like github.com/open-cluster-management/governance-policy-propagator) should import ***its*** fork of the api repo, like github.com/open-cluster-management/api. Naively you might need to update all the imports and module names in the propagator fork and the api fork. But if you update the module name to a non-github url, you can keep the module name constant, and just have a single `replace` directive in the forked propagator like
```
replace open-cluster-management.io/api v1.2.3 => github.com/open-cluster-management/api v1.4.5
```

JustinKuli (approved):
/hold
/lgtm

Tentative approval, waiting for website changes to make the module name work.

dhaiducek:
Ah, I understand the website thing better now, thanks! (I was thinking an update needed to be made at `pkg.go.dev` or something.) Yes, I'll make the website update, double check this PR, and unhold once the website is updated. Thanks!

dhaiducek:
Here's the redirect PR we're waiting on:
- https://github.com/open-cluster-management-io/open-cluster-management-io.github.io/pull/180

I went ahead and included all of our community repositories so that we can go through and update all of them.

mprahl (approved):

mprahl (approved):
Nice!

PR #24, opened 2022-02-08
Sync upstream with midstream changes

mprahl:
Many of the cherry picked commits did not apply cleanly due to the imports being different in midstream (github.com/stolostron) vs upstream (open-cluster-management.io/). The patches had to be manually fixed to account for this.


JustinKuli (approved):
Sad about the imports. If/when stolostron propagator switches to the `open-cluster-management.io/` path, it will break all the things that import it, I think? Seems like we need a plan to carefully update and fix things one at a time without too much extended disruption, or a chance to break everything all at once.

Also, there is no upstream for the template utils? Seems odd to have this community repo depending on the stolostron package.

mprahl:
> Sad about the imports. If/when stolostron propagator switches to the `open-cluster-management.io/` path, it will break all the things that import it, I think? Seems like we need a plan to carefully update and fix things one at a time without too much extended disruption, or a chance to break everything all at once.
> 
> Also, there is no upstream for the template utils? Seems odd to have this community repo depending on the stolostron package.

@dhaiducek had checked with the community to see if we should have it also upstream, but I believe the thought process was that since it's a library, it didn't need to be upstream.

dhaiducek:
Phew! Thanks for syncing these, @mprahl!!!

PR #31, opened 2022-04-08
Fix text/template so that it doesn't confuse the action for a delimiter

mprahl:
This backports the upstream fix to the text/template library. This is
required so that fields that start with the letter `h` can be accessed
within Hub policy templates.

The second commit is to fix the CI when running against the latest Kubernetes version.

Relates:
https://github.com/stolostron/backlog/issues/21440


gparvin (approved):
Nice patch!

dhaiducek:
@mprahl Is this ready to be merged?

PR #38, opened 2022-07-12
Add an "everyEvent" Ansible Job Run Mode to the Policy addon

ChunxiAlexLuo:

https://github.com/open-cluster-management-io/enhancements/pull/60

### DIFF
--- a/api/v1beta1/policyautomation_types.go
+++ b/api/v1beta1/policyautomation_types.go
@@ -14,17 +14,27 @@ type PolicyAutomationSpec struct {
 	// +kubebuilder:validation:Required
 	PolicyRef string `json:"policyRef"`
 	// Mode decides how automation is going to be triggered
-	// +kubebuilder:validation:Enum={once,disabled}
-	// +kubebuilder:validation:Required
-	Mode string `json:"mode"`
+	Mode PolicyAutomationMode `json:"mode"`
 	// EventHook decides when automation is going to be triggered
 	// +kubebuilder:validation:Enum={noncompliant}
 	// +kubebuilder:validation:Required
 	EventHook   string `json:"eventHook,omitempty"`
 	RescanAfter string `json:"rescanAfter,omitempty"`
+	// +kubebuilder:validation:Minimum=0
+	DelayAfterRunSeconds uint `json:"delayAfterRunSeconds,omitempty"`
 	// +kubebuilder:validation:Required
 	Automation AutomationDef `json:"automationDef"`
 }
+
+// +kubebuilder:validation:Enum={once,everyEvent,disabled}
+// +kubebuilder:validation:Required
+type PolicyAutomationMode string
+
+const (
+	Once       PolicyAutomationMode = "once"
+	EveryEvent PolicyAutomationMode = "everyEvent"
+	Disabled   PolicyAutomationMode = "disabled"
+)
 
 // AutomationDef defines the automation to invoke
 type AutomationDef struct {
@@ -43,7 +53,10 @@ type AutomationDef struct {
 }
 
 // PolicyAutomationStatus defines the observed state of PolicyAutomation
-type PolicyAutomationStatus struct{}
+type PolicyAutomationStatus struct {
+	// Cluster name as the key of ClustersWithEvent
+	ClustersWithEvent map[string]ClusterEvent `json:"clustersWithEvent,omitempty"`
+}
 
 //+kubebuilder:object:root=true
 
@@ -66,6 +79,14 @@ type PolicyAutomationList struct {
 	metav1.TypeMeta `json:",inline"`
 	metav1.ListMeta `json:"metadata,omitempty"`
 	Items           []PolicyAutomation `json:"items"`
+}
+
+// PolicyAutomation events on each target cluster
+type ClusterEvent struct {
+	// Policy automation start time for everyEvent mode
+	AutomationStartTime string `json:"automationStartTime"`
+	// The last policy compliance transition event time
+	EventTime string `json:"eventTime"`
 }
 
 func init() {
--- a/controllers/automation/policyautomation_controller.go
+++ b/controllers/automation/policyautomation_controller.go
@@ -112,7 +112,7 @@ func (r *PolicyAutomationReconciler) Reconcile(
 		}
 
 		return reconcile.Result{}, nil
-	} else if policyAutomation.Spec.Mode == "disabled" {
+	} else if policyAutomation.Spec.Mode == policyv1beta1.Disabled {
 		log.Info("Automation is disabled, doing nothing")
 
 		return reconcile.Result{}, nil
@@ -175,20 +175,26 @@ func (r *PolicyAutomationReconciler) Reconcile(
 			)
 
 			return reconcile.Result{RequeueAfter: requeueAfter}, nil
-		} else if policyAutomation.Spec.Mode == "once" {
-			log := log.WithValues("mode", "once")
+		} else if policyAutomation.Spec.Mode == policyv1beta1.Once {
+			log := log.WithValues("mode", string(policyv1beta1.Once))
 			targetList := common.FindNonCompliantClustersForPolicy(policy)
 			if len(targetList) > 0 {
 				log.Info("Creating an Ansible job", "targetList", targetList)
 
-				err = common.CreateAnsibleJob(policyAutomation, r.DynamicClient, "once", targetList)
+				err = common.CreateAnsibleJob(
+					policyAutomation,
+					r.DynamicClient,
+					string(policyv1beta1.Once),
+					targetList,
+				)
+
 				if err != nil {
 					log.Error(err, "Failed to create the Ansible job")
 
 					return reconcile.Result{}, err
 				}
 
-				policyAutomation.Spec.Mode = "disabled"
+				policyAutomation.Spec.Mode = policyv1beta1.Disabled
 
 				err = r.Update(ctx, policyAutomation, &client.UpdateOptions{})
 				if err != nil {
@@ -198,6 +204,131 @@ func (r *PolicyAutomationReconciler) Reconcile(
 				}
 			} else {
 				log.Info("All clusters are compliant. Doing nothing.")
+			}
+		} else if policyAutomation.Spec.Mode == policyv1beta1.EveryEvent {
+			log := log.WithValues("mode", string(policyv1beta1.EveryEvent))
+			targetList := common.FindNonCompliantClustersForPolicy(policy)
+			// Convert slice targetList to map for search efficiency
+			targetListMap := map[string]bool{}
+			for _, target := range targetList {
+				targetListMap[target] = true
+			}
+			// The final clusters list that the new ansible job will target
+			trimmedTargetList := []string{}
+			// delayAfterRunSeconds and requeueDuration default value = zero
+			delayAfterRunSeconds := policyAutomation.Spec.DelayAfterRunSeconds
+			requeueDuration := 0
+			requeueFlag := false
+			// Automation event time grouped by the cluster name
+			eventMap := map[string]policyv1beta1.ClusterEvent{}
+			if len(policyAutomation.Status.ClustersWithEvent) > 0 {
+				eventMap = policyAutomation.Status.ClustersWithEvent
+			}
+
+			now := time.Now().UTC()
+			nowStr := now.Format(time.RFC3339)
+
+			for clusterName, clusterEvent := range eventMap {
+				originalStartTime, err := time.Parse(time.RFC3339, clusterEvent.AutomationStartTime)
+				if err != nil {
+					log.Error(err, "Failed to retrieve AutomationStartTime in ClustersWithEvent")
+					delete(eventMap, clusterName)
+				}
+
+				preEventTime, err := time.Parse(time.RFC3339, clusterEvent.EventTime)
+				if err != nil {
+					log.Error(err, "Failed to retrieve EventTime in ClustersWithEvent")
+					delete(eventMap, clusterName)
+				}
+
+				// The time that delayAfterRunSeconds setting expires
+				delayUntil := originalStartTime.Add(time.Duration(delayAfterRunSeconds) * time.Second)
+
+				// The policy is non-compliant with the target cluster
+				if targetListMap[clusterName] {
+					// Policy status changed from non-compliant to compliant
+					// then back to non-compliant during the delay period
+					if delayAfterRunSeconds > 0 && preEventTime.After(originalStartTime) {
+						if now.After(delayUntil) {
+							// The delay period passed so remove the previous event
+							delete(eventMap, clusterName)
+							// Add the cluster name to create a new ansible job
+							trimmedTargetList = append(trimmedTargetList, clusterName)
+						} else {
+							requeueFlag = true
+							// Within the delay period and use the earliest requeueDuration to requeue
+							if (requeueDuration == 0) || (requeueDuration > int(delayUntil.Sub(now)+1)) {
+								requeueDuration = int(delayUntil.Sub(now) + 1)
+							}
+							// keep the event and update eventTime
+							clusterEvent.EventTime = nowStr
+							// new event from compliant to non-compliant
+							eventMap[clusterName] = clusterEvent
+						}
+					} // Otherwise, the policy keeps non-compliant since originalStartTime, do nothing
+				} else { // The policy is compliant with the target cluster
+					if delayAfterRunSeconds > 0 && now.Before(delayUntil) {
+						// Within the delay period, keep the event and update eventTime
+						clusterEvent.EventTime = nowStr
+						// new event from non-compliant to compliant
+						eventMap[clusterName] = clusterEvent
+					} else { // No delay period or it is expired, remove the event
+						delete(eventMap, clusterName)
+					}
+				}
+			}
+
+			for _, clusterName := range targetList {
+				if _, ok := eventMap[clusterName]; !ok {
+					// Add the non-compliant clusters without previous automation event
+					trimmedTargetList = append(trimmedTargetList, clusterName)
+				}
+			}
+
+			if len(trimmedTargetList) > 0 {
+				log.Info("Creating An Ansible job", "trimmedTargetList", trimmedTargetList)
+				err = common.CreateAnsibleJob(
+					policyAutomation,
+					r.DynamicClient,
+					string(policyv1beta1.EveryEvent),
+					trimmedTargetList,
+				)
+
+				if err != nil {
+					log.Error(err, "Failed to create the Ansible job")
+
+					return reconcile.Result{}, err
+				}
+
+				automationStartTimeStr := time.Now().UTC().Format(time.RFC3339)
+
+				for _, clusterName := range trimmedTargetList {
+					eventMap[clusterName] = policyv1beta1.ClusterEvent{
+						AutomationStartTime: automationStartTimeStr,
+						EventTime:           nowStr,
+					}
+				}
+			} else {
+				log.Info("All clusters are compliant. No new Ansible job. Just update ClustersWithEvent.")
+			}
+
+			policyAutomation.Status.ClustersWithEvent = eventMap
+			// use StatusWriter to update status subresource of a Kubernetes object
+			err = r.Status().Update(ctx, policyAutomation, &client.UpdateOptions{})
+			if err != nil {
+				log.Error(err, "Failed to update ClustersWithEvent in policyAutomation status")
+
+				return reconcile.Result{}, err
+			}
+
+			if requeueFlag {
+				log.Info(
+					"Requeue for the new non-compliant event during the delay period",
+					"Delay in seconds", delayAfterRunSeconds,
+					"Requeue After", requeueDuration,
+				)
+
+				return reconcile.Result{RequeueAfter: time.Duration(requeueDuration)}, nil
 			}
 		}
 	}
--- a/deploy/crds/policy.open-cluster-management.io_policyautomations.yaml
+++ b/deploy/crds/policy.open-cluster-management.io_policyautomations.yaml
@@ -61,6 +61,9 @@ spec:
                 - name
                 - secret
                 type: object
+              delayAfterRunSeconds:
+                minimum: 0
+                type: integer
               eventHook:
                 description: EventHook decides when automation is going to be triggered
                 enum:
@@ -70,6 +73,7 @@ spec:
                 description: Mode decides how automation is going to be triggered
                 enum:
                 - once
+                - everyEvent
                 - disabled
                 type: string
               policyRef:
@@ -85,6 +89,23 @@ spec:
             type: object
           status:
             description: PolicyAutomationStatus defines the observed state of PolicyAutomation
+            properties:
+              clustersWithEvent:
+                additionalProperties:
+                  description: PolicyAutomation events on each target cluster
+                  properties:
+                    automationStartTime:
+                      description: Policy automation start time for everyEvent mode
+                      type: string
+                    eventTime:
+                      description: The last policy compliance transition event time
+                      type: string
+                  required:
+                  - automationStartTime
+                  - eventTime
+                  type: object
+                description: Cluster name as the key of ClustersWithEvent
+                type: object
             type: object
         required:
         - spec
--- a/test/e2e/case5_policy_automation_test.go
+++ b/test/e2e/case5_policy_automation_test.go
@@ -11,6 +11,7 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 
 	policiesv1 "open-cluster-management.io/governance-policy-propagator/api/v1"
+	policyv1beta1 "open-cluster-management.io/governance-policy-propagator/api/v1beta1"
 	"open-cluster-management.io/governance-policy-propagator/controllers/common"
 	"open-cluster-management.io/governance-policy-propagator/test/utils"
 )
@@ -53,7 +54,7 @@ var _ = Describe("Test policy automation", func() {
 		It("Test mode = disable", func() {
 			By("Creating an policyAutomation with mode=disable")
 			_, err := utils.KubectlWithOutput("apply",
-				"-f", "../resources/case5_policy_automation/case5-policy-automation.yaml",
+				"-f", "../resources/case5_policy_automation/case5-policy-automation-disable.yaml",
 				"-n", testNamespace)
 			Expect(err).Should(BeNil())
 			By("Should not create any ansiblejob when mode = disable")
@@ -66,13 +67,14 @@ var _ = Describe("Test policy automation", func() {
 				return len(ansiblejobList.Items)
 			}, 30, 1).Should(Equal(0))
 		})
+
 		It("Test mode = once", func() {
 			By("Patching policyAutomation with mode=once")
 			policyAutomation, err := clientHubDynamic.Resource(gvrPolicyAutomation).Namespace(testNamespace).Get(
 				context.TODO(), "create-service-now-ticket", metav1.GetOptions{},
 			)
 			Expect(err).To(BeNil())
-			policyAutomation.Object["spec"].(map[string]interface{})["mode"] = "once"
+			policyAutomation.Object["spec"].(map[string]interface{})["mode"] = string(policyv1beta1.Once)
 			_, err = clientHubDynamic.Resource(gvrPolicyAutomation).Namespace(testNamespace).Update(
 				context.TODO(), policyAutomation, metav1.UpdateOptions{},
 			)
@@ -95,7 +97,7 @@ var _ = Describe("Test policy automation", func() {
 
 				return len(ansiblejobList.Items)
 			}, 30, 1).Should(Equal(0))
-			By("Patching policy to make both cluster NonCompliant")
+			By("Patching policy to make both clusters NonCompliant")
 			opt := metav1.ListOptions{
 				LabelSelector: common.RootPolicyLabel + "=" + testNamespace + "." + case5PolicyName,
 			}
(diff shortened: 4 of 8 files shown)


api/v1beta1/policyautomation_types.go
```diff
@@ -14,14 +14,15 @@ type PolicyAutomationSpec struct {
 	// +kubebuilder:validation:Required
 	PolicyRef string `json:"policyRef"`
 	// Mode decides how automation is going to be triggered
-	// +kubebuilder:validation:Enum={once,disabled}
+	// +kubebuilder:validation:Enum={once,everyEvent,disabled}
 	// +kubebuilder:validation:Required
 	Mode string `json:"mode"`
```
mprahl, line 19: One of the things I was hoping to do as part of this feature is to make `Mode` a custom type rather than `string`. You can call the type `PolicyAutomationMode` and then define constants for each value in the enum.

This is the standard approach for defining Kubernetes APIs but was not done in the original implementation here.

api/v1beta1/policyautomation_types.go
```diff
@@ -14,14 +14,15 @@ type PolicyAutomationSpec struct {
...
 	// EventHook decides when automation is going to be triggered
 	// +kubebuilder:validation:Enum={noncompliant}
 	// +kubebuilder:validation:Required
-	EventHook   string `json:"eventHook,omitempty"`
-	RescanAfter string `json:"rescanAfter,omitempty"`
+	EventHook            string `json:"eventHook,omitempty"`
+	RescanAfter          string `json:"rescanAfter,omitempty"`
+	DelayAfterRunSeconds string `json:"delayAfterRunSeconds,omitempty"`
```
mprahl, line 25: Could you make this a `uint` or could you remove this and implement this part of the feature in a future PR?

api/v1beta1/policyautomation_types.go
```diff
@@ -68,6 +72,12 @@ type PolicyAutomationList struct {
 	Items           []PolicyAutomation `json:"items"`
 }
 
+// PolicyAutomation event time on each cluster
+type ClusterEventTime struct {
+	AutomationStartTime string `json:"automationStartTime"`
```
mprahl, line 77: Could you please add comments to these fields so that they end up documenting the fields in the generated CRD?

controllers/automation/policyautomation_controller.go
```diff
@@ -199,6 +201,121 @@ func (r *PolicyAutomationReconciler) Reconcile(
 			} else {
 				log.Info("All clusters are compliant. Doing nothing.")
 			}
+		} else if policyAutomation.Spec.Mode == "everyEvent" {
+			log := log.WithValues("mode", "everyEvent")
+			targetList := common.FindNonCompliantClustersForPolicy(policy)
+			eventTime := time.Now().UTC().Format(time.RFC3339)
+			if len(targetList) > 0 {
+				delayAfterRunSecondsStr := policyAutomation.Spec.DelayAfterRunSeconds
+				delayAfterRunSeconds, err := strconv.Atoi(delayAfterRunSecondsStr)
```
mprahl, line 210: This can be removed after changing to the type of `DelayAfterRunSeconds` to `uint`.

controllers/automation/policyautomation_controller.go
```diff
@@ -199,6 +201,121 @@ func (r *PolicyAutomationReconciler) Reconcile(
...
+					err = r.Update(ctx, policyAutomation, &client.UpdateOptions{})
+					if err != nil {
+						log.Error(err, "Failed to update clusters event time in policyAutomation status")
+
+						return reconcile.Result{}, err
+					}
+				} else {
+					eventMap := statusTemp.ClustersWithEvent
+
+					targetListStr := strings.Join(targetList, ",")
```
mprahl, line 248: Could you make this a set instead?

For example:
```suggestion
					targets = map[string]bool{}
					for _, target := range targetList {
						targets[target] = true
					}
```


Then you can just do the following to check if it's present, which is a very fast operation:
```go
if targets[clusterName] {
```

controllers/automation/policyautomation_controller.go
```diff
@@ -199,6 +201,121 @@ func (r *PolicyAutomationReconciler) Reconcile(
...
+			eventTime := time.Now().UTC().Format(time.RFC3339)
+			if len(targetList) > 0 {
+				delayAfterRunSecondsStr := policyAutomation.Spec.DelayAfterRunSeconds
+				delayAfterRunSeconds, err := strconv.Atoi(delayAfterRunSecondsStr)
+				if err == nil {
+					log.Info("Can't read delayAfterRunSeconds, set to default value zero")
+					// if delayAfterRunSeconds is unset or isn't int, set to default value 0
+					delayAfterRunSeconds = 0
+				}
+				statusTemp := policyAutomation.Status
```
mprahl, line 216: This doesn't actually create a copy since `policyAutomation.Status` is a map. I also don't think you need to create a copy because `err := r.Get(ctx, request.NamespacedName, policyAutomation)` receives a copy from the cache.

controllers/automation/policyautomation_controller.go
```diff
@@ -199,6 +201,121 @@ func (r *PolicyAutomationReconciler) Reconcile(
...
+				if len(statusTemp.ClustersWithEvent) == 0 {
```
mprahl, line 217: Could you please consolidate this code with the else branch?

I think the general flow in the `else` branch would work for this case too.

ChunxiAlexLuo:
> > Signed-off-by: Chunxi Luo [chuluo@redhat.com](mailto:chuluo@redhat.com)
> > This PR needs to be polished and combine some duplicated codes section. Also adding test cases.
> > I opened it mainly to clarify the purpose of `EventTime`. After re-reading the design doc and discussion, I don't get the case we use `EventTime` (check `delayAfterRunSecondsStr` or something else). Especially for `This is detected by looking at the eventTime in the status.clustersWithEvent entry to see if the timestamp is newer than the automationStartTime value.` If comparing `automationStartTime` + `delayAfterRunSeconds` with Now isn't enough, I must miss sth, and thanks for clarifying. @JustinKuli @mprahl
> 
> @ChunxiAlexLuo `eventTime` is set whenever the policy goes from compliant to noncompliant and `automationStartTime` is recording when the `AnsibleJob` object is created.
> 
> At first, these values will be exactly the same or almost the same depending on the implementation. This isn't useful until we have the `delayAfterRunSeconds` field.
> 
> The reason we need `eventTime` in addition to `automationStartTime` is because we need to catch the case where the policy is flaky and went from noncompliant to compliant and back to noncompliant while waiting for `delayAfterRunSeconds`. In this case, we need to trigger the `AnsibleJob` again. We don't want to trigger an Ansible job a second time if the policy stayed noncompliant after `automationStartTime + delayAfterRunSeconds < now`.
> 
> Does this make sense?

My original understanding is just the opposite here, I thought like the case @JustinKuli described https://github.com/open-cluster-management-io/enhancements/pull/60#discussion_r904034872, it will ignore the change on t=120 and t =500, then wait till t=1200 (no matter what happened during delayAfterRunSeconds). If re-run on t=610 then `eventTime` is needed.

mprahl:
> > > Signed-off-by: Chunxi Luo [chuluo@redhat.com](mailto:chuluo@redhat.com)
> > > This PR needs to be polished and combine some duplicated codes section. Also adding test cases.
> > > I opened it mainly to clarify the purpose of `EventTime`. After re-reading the design doc and discussion, I don't get the case we use `EventTime` (check `delayAfterRunSecondsStr` or something else). Especially for `This is detected by looking at the eventTime in the status.clustersWithEvent entry to see if the timestamp is newer than the automationStartTime value.` If comparing `automationStartTime` + `delayAfterRunSeconds` with Now isn't enough, I must miss sth, and thanks for clarifying. @JustinKuli @mprahl
> > 
> > 
> > @ChunxiAlexLuo `eventTime` is set whenever the policy goes from compliant to noncompliant and `automationStartTime` is recording when the `AnsibleJob` object is created.
> > At first, these values will be exactly the same or almost the same depending on the implementation. This isn't useful until we have the `delayAfterRunSeconds` field.
> > The reason we need `eventTime` in addition to `automationStartTime` is because we need to catch the case where the policy is flaky and went from noncompliant to compliant and back to noncompliant while waiting for `delayAfterRunSeconds`. In this case, we need to trigger the `AnsibleJob` again. We don't want to trigger an Ansible job a second time if the policy stayed noncompliant after `automationStartTime + delayAfterRunSeconds < now`.
> > Does this make sense?
> 
> My original understanding is just the opposite here, I thought like the case @JustinKuli described [open-cluster-management-io/enhancements#60 (comment)](https://github.com/open-cluster-management-io/enhancements/pull/60#discussion_r904034872), it will ignore the change on t=120 and t =500, then wait till t=1200 (no matter what happened during delayAfterRunSeconds). If re-run on t=610 then `eventTime` is needed.

See the last comment in that discussion:
https://github.com/open-cluster-management-io/enhancements/pull/60#discussion_r904090980

ChunxiAlexLuo:
> flaky

Ok, so we want to track at most one time for the multi-times flaky case, and only able to do this without tracking the whole hisstory.

api/v1beta1/policyautomation_types.go
```diff
@@ -14,18 +14,21 @@ type PolicyAutomationSpec struct {
...
+	EventHook            string `json:"eventHook,omitempty"`
+	RescanAfter          string `json:"rescanAfter,omitempty"`
+	DelayAfterRunSeconds uint   `json:"delayAfterRunSeconds,omitempty"`
 	// +kubebuilder:validation:Required
 	Automation AutomationDef `json:"automationDef"`
 }
 
+// +kubebuilder:validation:Enum={once,everyEvent,disabled}
+// +kubebuilder:validation:Required
+type PolicyAutomationMode string
```
mprahl, line 30: Could you please set constants of all the different automation modes?

For example, we do this for `remediationAction` in the Configuration Policy controller:
https://github.com/stolostron/config-policy-controller/blob/a8f9211ab02b134140cdfcc2b02060912556424b/api/v1/configurationpolicy_types.go#L34-L40

api/v1beta1/policyautomation_types.go
```diff
@@ -43,7 +46,10 @@ type AutomationDef struct {
 }
 
 // PolicyAutomationStatus defines the observed state of PolicyAutomation
-type PolicyAutomationStatus struct{}
+// The key of ClustersWithEvent is the cluster name
+type PolicyAutomationStatus struct {
+	ClustersWithEvent map[string]ClusterEvent `json:"clustersWithEvent"`
```
mprahl, line 51: Could you please add `omitempty` to the `json` tag?

api/v1beta1/policyautomation_types.go
```diff
@@ -14,18 +14,21 @@ type PolicyAutomationSpec struct {
...
-	Mode string `json:"mode"`
+	Mode PolicyAutomationMode `json:"mode"`
 	// EventHook decides when automation is going to be triggered
 	// +kubebuilder:validation:Enum={noncompliant}
 	// +kubebuilder:validation:Required
-	EventHook   string `json:"eventHook,omitempty"`
-	RescanAfter string `json:"rescanAfter,omitempty"`
+	EventHook            string `json:"eventHook,omitempty"`
+	RescanAfter          string `json:"rescanAfter,omitempty"`
+	DelayAfterRunSeconds uint   `json:"delayAfterRunSeconds,omitempty"`
```
mprahl, line 23: Could you please set the `// +kubebuilder:validation:Minimum` marker so that there is API validation that the value is at least 0 (i.e. no negative numbers)?

controllers/automation/policyautomation_controller.go
```diff
@@ -199,6 +199,105 @@ func (r *PolicyAutomationReconciler) Reconcile(
...
+						targetListMap[target] = true
+					}
+
+					for clusterName := range eventMap {
+						clusterEvent := eventMap[clusterName]
+						oldAutomationStartTime, err := time.Parse(time.RFC3339, clusterEvent.AutomationStartTime)
+						if err != nil {
+							log.Error(err, "Failed to convert AutomationStartTime")
+
+							return reconcile.Result{}, err
```
mprahl, line 229: This doesn't seem like an error that we can recover from. I imagine this would only happen if a user modified the status manually. In this case, I would delete the entry from `eventMap` and continue.

controllers/automation/policyautomation_controller.go
```diff
@@ -199,6 +199,105 @@ func (r *PolicyAutomationReconciler) Reconcile(
...
+						}
+
+						oldEventTime, err := time.Parse(time.RFC3339, clusterEvent.EventTime)
+						if err != nil {
+							log.Error(err, "Failed to convert EventTime")
+
+							return reconcile.Result{}, err
```
mprahl, line 236: Same as my comment for `AutomationStartTime`.

controllers/automation/policyautomation_controller.go
```diff
@@ -199,6 +199,105 @@ func (r *PolicyAutomationReconciler) Reconcile(
 			} else {
 				log.Info("All clusters are compliant. Doing nothing.")
 			}
+		} else if policyAutomation.Spec.Mode == "everyEvent" {
+			log := log.WithValues("mode", "everyEvent")
+			targetList := common.FindNonCompliantClustersForPolicy(policy)
```
mprahl, line 204: You need to filter out the clusters in `targetList` that are also in `policyAutomation.Status.ClustersWithEvent` since `FindNonCompliantClustersForPolicy` just gives you the current noncompliant clusters but the last time the `Reconcile` method ran, those clusters could have already been noncompliant.

controllers/automation/policyautomation_controller.go
```diff
@@ -199,6 +199,105 @@ func (r *PolicyAutomationReconciler) Reconcile(
 			} else {
 				log.Info("All clusters are compliant. Doing nothing.")
 			}
+		} else if policyAutomation.Spec.Mode == "everyEvent" {
```
mprahl, line 202: If there is `DelayAfterRunSeconds` set and the cluster went from non-compliant -> compliant -> noncompliant, then you'll need to requeue the request so that the `Reconcile` method is called after the `DelayAfterRunSeconds` has elapsed so that the automation is triggered again.

mprahl (changes requested):

controllers/automation/policyautomation_controller.go
ChunxiAlexLuo: I trimmed ClustersWithEvent first and then filter `targetList` as trimmedTargetList on line 275

controllers/automation/policyautomation_controller.go
```diff
@@ -199,6 +199,123 @@ func (r *PolicyAutomationReconciler) Reconcile(
...
+				}
+			}
+
+			for _, clusterName := range targetList {
+				if _, ok := eventMap[clusterName]; !ok {
+					// also add the clusters not in the existing automation event
+					trimmedTargetList = append(trimmedTargetList, clusterName)
+				}
+			}
+			log.Info("trimmedTargetList", trimmedTargetList)
```
mprahl, line 281: Could you remove this log?

controllers/automation/policyautomation_controller.go
```diff
@@ -199,6 +199,123 @@ func (r *PolicyAutomationReconciler) Reconcile(
...
+
+				for _, clusterName := range trimmedTargetList {
+					// update policyAutomation.Status.ClustersWithEvent
+					eventMap[clusterName] = policyv1beta1.ClusterEvent{
+						AutomationStartTime: automationStartTimeStr,
+						EventTime:           eventTimeStr,
+					}
+				}
+
+				log.Info("ClustersWithEvent", eventMap)
```
mprahl, line 301: Could you please remove this log message?

mprahl (changes requested):

controllers/automation/policyautomation_controller.go
```diff
@@ -199,6 +199,123 @@ func (r *PolicyAutomationReconciler) Reconcile(
...
+
+				oldEventTime, err := time.Parse(time.RFC3339, clusterEvent.EventTime)
+				if err != nil {
+					log.Error(err, "Failed to convert EventTime in ClustersWithEvent")
+
+					delete(eventMap, clusterName)
+				}
+
+				delayedTime := oldAutomationStartTime.Add(
+					time.Duration(delayAfterRunSeconds) * time.Second)
```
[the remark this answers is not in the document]
ChunxiAlexLuo: Here there is a complaint since delayAfterRunSeconds is unit and Second is time.Duration type, my original meaning is converted delayAfterRunSeconds to time.Duration type first then do multiply with Second.

api/v1beta1/policyautomation_types.go
```diff
@@ -14,18 +14,28 @@ type PolicyAutomationSpec struct {
...
 	Automation AutomationDef `json:"automationDef"`
 }
 
+// +kubebuilder:validation:Enum={once,everyevent,disabled}
+// +kubebuilder:validation:Required
+type PolicyAutomationMode string
+
+const (
+	Once       PolicyAutomationMode = "once"
+	EveryEvent PolicyAutomationMode = "everyevent"
```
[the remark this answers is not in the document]
mprahl: Why not just make it lowercase when creating the Ansible job name?

test/e2e/case5_policy_automation_test.go
```diff
@@ -135,7 +136,448 @@ var _ = Describe("Test policy automation", func() {
...
+			for _, replicatedPlc := range replicatedPlcList.Items {
+				replicatedPlc.Object["status"] = &policiesv1.PolicyStatus{
+					ComplianceState: policiesv1.Compliant,
+				}
+				_, err := clientHubDynamic.Resource(gvrPolicy).Namespace(replicatedPlc.GetNamespace()).UpdateStatus(
+					context.TODO(), &replicatedPlc, metav1.UpdateOptions{},
+				)
+				Expect(err).To(BeNil())
+			}
+			By("Waiting until the status on each cluster is updated to Compliant")
```
ChunxiAlexLuo, line 159: Based on many times of testing on OCM env, unlike `once` mode testing, we need to verify each policy status updating is completed for `everyEvent` test case. Because it will update the policy status very often and this check will increase the testing stability.

api/v1beta1/policyautomation_types.go
```diff
@@ -68,6 +81,14 @@ type PolicyAutomationList struct {
 	Items           []PolicyAutomation `json:"items"`
 }
 
+// PolicyAutomation events on each target cluster
+type ClusterEvent struct {
+	// Policy automation starts time as everyEvent mode
+	AutomationStartTime string `json:"automationStartTime"`
+	// The last time policy status from non-compliance to compliance
```
gparvin, line 88: Wouldn't the time here be the change from compliant to noncompliant?

deploy/crds/policy.open-cluster-management.io_policyautomations.yaml
```diff
@@ -85,6 +89,24 @@ spec:
...
+            properties:
+              clustersWithEvent:
+                additionalProperties:
+                  description: PolicyAutomation events on each target cluster
+                  properties:
+                    automationStartTime:
+                      description: Policy automation starts time as everyEvent mode
+                      type: string
+                    eventTime:
+                      description: The last time policy status from non-compliance
```
gparvin, line 101: Wouldn't the time here be the change from compliant to noncompliant?

gparvin (commented):
It seems like the descriptions for the API might have describe the status wrong.  Can you double check the way we describe the status fields in the API?  Other than that this looks good, thanks!

api/v1beta1/policyautomation_types.go
ChunxiAlexLuo: @gparvin  I agree here is a little weird and I will try to explain it:

In the first commit, I just update `eventTime` for the case that in the delay period, policy status change from non-compliant to compliant:
https://github.com/open-cluster-management-io/governance-policy-propagator/pull/38/commits/ce8fec06d0b5914d6670d9452fdd28e1d86ab6d2#diff-746b54415213bba767535a3153aff159b013aea7e7d74f6bcfeb0f135bd7c79cR270

Then next time when policy status changes from compliant to non-compliant again,  it will go to here:
https://github.com/open-cluster-management-io/governance-policy-propagator/pull/38/commits/ce8fec06d0b5914d6670d9452fdd28e1d86ab6d2#diff-746b54415213bba767535a3153aff159b013aea7e7d74f6bcfeb0f135bd7c79cR251

There are two ways to do this: 
A) just update `eventTime` for the event from non-compliant to compliant 
vs 
B)  updates`eventTime` for both events - non-compliant to compliant and compliant to non-compliant

If you run some test cases in mind, either A) or B) will work for the final result but B) may be easier to understand 
So I made another commit,  @gparvin @mprahl  you can comment on the second commit: https://github.com/open-cluster-management-io/governance-policy-propagator/pull/38/commits/1b2c18f6b93ec99ba17512fd60a0dd99103dd24c

@gparvin I'm not sure if your thought is the way C) just update `eventTime` for the event from compliant to non-compliant, but that way doesn't work for the multiple event case in the delay period. We can't tell the difference between the first time `compliant to non-compliant` and the second time `compliant to non-compliant` because we don't record all events details.

ChunxiAlexLuo: @gparvin @mprahl Keeping two commits for easier reviewing, I will consolidate them later.

gparvin: Perfect, thanks Alex.  That was the additional clarity I wanted to confirm so this look great.

gparvin (approved):

api/v1beta1/policyautomation_types.go
ChunxiAlexLuo: @gparvin Thanks for reviewing, I just consolidated two commits.

controllers/automation/policyautomation_controller.go
```diff
@@ -175,20 +175,26 @@ func (r *PolicyAutomationReconciler) Reconcile(
...
-		} else if policyAutomation.Spec.Mode == "once" {
-			log := log.WithValues("mode", "once")
+		} else if policyAutomation.Spec.Mode == policyv1beta1.Once {
+			log := log.WithValues("mode", string(policyv1beta1.Once))
 			targetList := common.FindNonCompliantClustersForPolicy(policy)
 			if len(targetList) > 0 {
 				log.Info("Creating an Ansible job", "targetList", targetList)
 
-				err = common.CreateAnsibleJob(policyAutomation, r.DynamicClient, "once", targetList)
+				err = common.CreateAnsibleJob(
```
mprahl, line 184: It'd be better if `CreateAnsibleJob` accepted a `PolicyAutomationMode` type instead so you don't have to convert to a string.

controllers/automation/policyautomation_controller.go
```diff
@@ -199,6 +205,128 @@ func (r *PolicyAutomationReconciler) Reconcile(
...
+							// Within the delay period and use the earliest requeueDuration to requeue
+							if (requeueDuration == 0) || (requeueDuration > int(delayUntil.Sub(now)+1)) {
+								requeueDuration = int(delayUntil.Sub(now) + 1)
+							}
+							// keep the event and update eventTime
+							clusterEvent.EventTime = nowStr
+							// new event from compliant to non-compliant
+							eventMap[clusterName] = clusterEvent
+						}
+					} // Otherwise, the policy keeps non-compliant since originalStartTime, do noting
```
mprahl, line 268:
```suggestion
					} // Otherwise, the policy keeps non-compliant since originalStartTime, do nothing
```

controllers/automation/policyautomation_controller.go
```diff
@@ -199,6 +205,128 @@ func (r *PolicyAutomationReconciler) Reconcile(
...
+				)
+
+				if err != nil {
+					log.Error(err, "Failed to create the Ansible job")
+
+					return reconcile.Result{}, err
+				}
+
+				automationStartTimeStr := time.Now().UTC().
+					Add(time.Second * time.Duration(1)).Format(time.RFC3339)
```
mprahl, line 304: Why are you adding 1 second here?

controllers/automation/policyautomation_controller.go
```diff
@@ -199,6 +205,128 @@ func (r *PolicyAutomationReconciler) Reconcile(
...
+			// use StatusWriter to update status subresource of a Kubernetes object
+			err = r.Status().Update(ctx, policyAutomation, &client.UpdateOptions{})
+			if err != nil {
+				log.Error(err, "Failed to update ClustersWithEvent in policyAutomation status")
+
+				return reconcile.Result{}, err
+			}
+
+			if requeueFlag {
+				log.Info("Requeue for the multi-events during the delay period")
```
mprahl, line 326: What do you mean by `multi-events`? Also, could you please log the delay in seconds?

api/v1beta1/policyautomation_types.go
```diff
@@ -68,6 +81,14 @@ type PolicyAutomationList struct {
 	Items           []PolicyAutomation `json:"items"`
 }
 
+// PolicyAutomation events on each target cluster
+type ClusterEvent struct {
+	// Policy automation starts time as everyEvent mode
```
mprahl, line 86:
```suggestion
	// Policy automation start time for everyEvent mode
```

api/v1beta1/policyautomation_types.go
```diff
@@ -68,6 +81,14 @@ type PolicyAutomationList struct {
...
+	AutomationStartTime string `json:"automationStartTime"`
+	// The last policy event time - the status switch between compliance and non-compliance
```
mprahl, line 88:
```suggestion
	// The last policy compliance transition event time
```

test/e2e/case5_policy_automation_test.go
```diff
@@ -135,7 +136,448 @@ var _ = Describe("Test policy automation", func() {
...
+			replicatedPlcList = utils.ListWithTimeout(clientHubDynamic, gvrPolicy, opt, 2, true, defaultTimeoutSeconds)
+			for _, replicatedPlc := range replicatedPlcList.Items {
+				Eventually(func() interface{} {
```
mprahl, line 162: This eventually will not work because `replicatedPlc` never changes inside the eventually.

test/e2e/case5_policy_automation_test.go
```diff
@@ -135,7 +136,448 @@ var _ = Describe("Test policy automation", func() {
...
+			)
+			Expect(err).To(BeNil())
+			policyAutomation.Object["spec"].(map[string]interface{})["mode"] = string(policyv1beta1.EveryEvent)
+			_, err = clientHubDynamic.Resource(gvrPolicyAutomation).Namespace(testNamespace).Update(
+				context.TODO(), policyAutomation, metav1.UpdateOptions{},
+			)
+			Expect(err).To(BeNil())
+
+			By("Should not create any new ansiblejob when Compliant, just one ansiblejob from once-mode test")
+			Consistently(func() interface{} {
```
mprahl, line 181: Can you delete the Ansible job in the once mode test so that it doesn't affect this test?

test/e2e/case5_policy_automation_test.go
```diff
@@ -135,7 +136,448 @@ var _ = Describe("Test policy automation", func() {
...
+				}
+				_, err := clientHubDynamic.Resource(gvrPolicy).Namespace(replicatedPlc.GetNamespace()).UpdateStatus(
+					context.TODO(), &replicatedPlc, metav1.UpdateOptions{},
+				)
+				Expect(err).To(BeNil())
+			}
+			By("Waiting until the status on each cluster is updated to NonCompliant")
+			replicatedPlcList = utils.ListWithTimeout(clientHubDynamic, gvrPolicy, opt, 2, true, defaultTimeoutSeconds)
+			for _, replicatedPlc := range replicatedPlcList.Items {
+				Eventually(func() interface{} {
```
mprahl, line 207: Same comment as above

test/e2e/case5_policy_automation_test.go
```diff
@@ -135,7 +136,448 @@ var _ = Describe("Test policy automation", func() {
...
+				}
+				_, err := clientHubDynamic.Resource(gvrPolicy).Namespace(replicatedPlc.GetNamespace()).UpdateStatus(
+					context.TODO(), &replicatedPlc, metav1.UpdateOptions{},
+				)
+				Expect(err).To(BeNil())
+			}
+			By("Waiting until the status on each cluster is updated to Compliant")
+			replicatedPlcList = utils.ListWithTimeout(clientHubDynamic, gvrPolicy, opt, 2, true, defaultTimeoutSeconds)
+			for _, replicatedPlc := range replicatedPlcList.Items {
+				Eventually(func() interface{} {
```
mprahl, line 253: Same as above

test/e2e/case5_policy_automation_test.go
```diff
@@ -135,7 +136,448 @@ var _ = Describe("Test policy automation", func() {
...
+				}
+				_, err := clientHubDynamic.Resource(gvrPolicy).Namespace(replicatedPlc.GetNamespace()).UpdateStatus(
+					context.TODO(), &replicatedPlc, metav1.UpdateOptions{},
+				)
+				Expect(err).To(BeNil())
+			}
+			By("Waiting until the status on each cluster is updated to NonCompliant")
+			replicatedPlcList = utils.ListWithTimeout(clientHubDynamic, gvrPolicy, opt, 2, true, defaultTimeoutSeconds)
+			for _, replicatedPlc := range replicatedPlcList.Items {
+				Eventually(func() interface{} {
```
mprahl, line 287: Same as above

test/e2e/case5_policy_automation_test.go
```diff
@@ -135,7 +136,448 @@ var _ = Describe("Test policy automation", func() {
...
+				}
+				_, err := clientHubDynamic.Resource(gvrPolicy).Namespace(replicatedPlc.GetNamespace()).UpdateStatus(
+					context.TODO(), &replicatedPlc, metav1.UpdateOptions{},
+				)
+				Expect(err).To(BeNil())
+			}
+			By("Waiting until the status on each cluster is updated to Compliant")
+			replicatedPlcList = utils.ListWithTimeout(clientHubDynamic, gvrPolicy, opt, 2, true, defaultTimeoutSeconds)
+			for _, replicatedPlc := range replicatedPlcList.Items {
+				Eventually(func() interface{} {
```
mprahl, line 333: Same as above

test/e2e/case5_policy_automation_test.go
```diff
@@ -135,7 +136,448 @@ var _ = Describe("Test policy automation", func() {
...
+					context.TODO(), metav1.ListOptions{},
+				)
+				Expect(err).To(BeNil())
+				_, err = utils.KubectlWithOutput("get", "ansiblejobs", "-n", testNamespace)
+				Expect(err).Should(BeNil())
+
+				return len(ansiblejobList.Items)
+			}, 30, 1).Should(Equal(3))
+
+			By("Patching policy to make both clusters back to Compliant")
```
mprahl, line 316: You can delete from this line to the remainder of the test because you already tested that flipping from non-compliant to compliant doesn't trigger a new job. It's unnecessary to test again.

test/e2e/case5_policy_automation_test.go
mprahl: Perhaps we can just delete this check since it seems like you didn't need it.

test/e2e/case5_policy_automation_test.go
```diff
@@ -135,7 +136,448 @@ var _ = Describe("Test policy automation", func() {
...
+			By("Should not create any new ansiblejob when Compliant, still three ansiblejobs")
+			Consistently(func() interface{} {
+				ansiblejobList, err := clientHubDynamic.Resource(gvrAnsibleJob).List(
+					context.TODO(), metav1.ListOptions{},
+				)
+				Expect(err).To(BeNil())
+
+				return len(ansiblejobList.Items)
+			}, 15, 1).Should(Equal(3))
+		})
```
mprahl, line 349: Could you please add a new line after this?

test/e2e/case5_policy_automation_test.go
```diff
@@ -135,7 +136,448 @@ var _ = Describe("Test policy automation", func() {
 				context.TODO(), "create-service-now-ticket", metav1.GetOptions{},
 			)
 			Expect(err).To(BeNil())
-			Expect(policyAutomation.Object["spec"].(map[string]interface{})["mode"]).To(Equal("disabled"))
+			Expect(
+				policyAutomation.Object["spec"].(map[string]interface{})["mode"],
+			).To(Equal(string(policyv1beta1.Disabled)))
+		})
+		// Create two events then two ansiblejobs for each event
+		It("Test mode = everyEvent without delayAfterRunSeconds", func() {
```
mprahl, line 144: Could you make each test clean up the PolicyAutomation object and the AnsibleJob objects so that each test does not depend on the other. This also allows running each test individually for debugging locally.

test/e2e/case5_policy_automation_test.go
```diff
@@ -135,7 +136,448 @@ var _ = Describe("Test policy automation", func() {
...
+				}
+				_, err := clientHubDynamic.Resource(gvrPolicy).Namespace(replicatedPlc.GetNamespace()).UpdateStatus(
+					context.TODO(), &replicatedPlc, metav1.UpdateOptions{},
+				)
+				Expect(err).To(BeNil())
+			}
+			By("Waiting until the status on each cluster is updated to NonCompliant")
+			replicatedPlcList = utils.ListWithTimeout(clientHubDynamic, gvrPolicy, opt, 2, true, defaultTimeoutSeconds)
+			for _, replicatedPlc := range replicatedPlcList.Items {
+				Eventually(func() interface{} {
```
mprahl, line 392: Same as above

test/e2e/case5_policy_automation_test.go
```diff
@@ -135,7 +136,448 @@ var _ = Describe("Test policy automation", func() {
...
+				}
+				_, err := clientHubDynamic.Resource(gvrPolicy).Namespace(replicatedPlc.GetNamespace()).UpdateStatus(
+					context.TODO(), &replicatedPlc, metav1.UpdateOptions{},
+				)
+				Expect(err).To(BeNil())
+			}
+			By("Waiting until the status on each cluster is updated to Compliant")
+			replicatedPlcList = utils.ListWithTimeout(clientHubDynamic, gvrPolicy, opt, 2, true, defaultTimeoutSeconds)
+			for _, replicatedPlc := range replicatedPlcList.Items {
+				Eventually(func() interface{} {
```
mprahl, line 438: Same as above

test/e2e/case5_policy_automation_test.go
```diff
@@ -135,7 +136,448 @@ var _ = Describe("Test policy automation", func() {
...
+				}
+				_, err := clientHubDynamic.Resource(gvrPolicy).Namespace(replicatedPlc.GetNamespace()).UpdateStatus(
+					context.TODO(), &replicatedPlc, metav1.UpdateOptions{},
+				)
+				Expect(err).To(BeNil())
+			}
+			By("Waiting until the status on each cluster is updated to NonCompliant")
+			replicatedPlcList = utils.ListWithTimeout(clientHubDynamic, gvrPolicy, opt, 2, true, defaultTimeoutSeconds)
+			for _, replicatedPlc := range replicatedPlcList.Items {
+				Eventually(func() interface{} {
```
mprahl, line 541: Same as above

test/e2e/case5_policy_automation_test.go
```diff
@@ -135,7 +136,448 @@ var _ = Describe("Test policy automation", func() {
...
+			Eventually(func() interface{} {
+				ansiblejobList, err := clientHubDynamic.Resource(gvrAnsibleJob).Namespace(testNamespace).List(
+					context.TODO(), metav1.ListOptions{},
+				)
+				Expect(err).To(BeNil())
+				_, err = utils.KubectlWithOutput("get", "ansiblejobs", "-n", testNamespace)
+				Expect(err).Should(BeNil())
+
+				return len(ansiblejobList.Items)
+			}, 240, 1).Should(Equal(5))
```
mprahl, line 570: Ideally, you would just patch the clustersWithEvent status to provide an older time for `eventTime` and `automationStartTime` before making the policies non-compliant again so that the test doesn't have to wait 240 seconds.

controllers/automation/policyautomation_controller.go
ChunxiAlexLuo: I thought about this, it seems we still have `manual` and `scan` not in PolicyAutomationMode. Do you prefer we also add them in PolicyAutomationMode or keep `CreateAnsibleJob` mode parameter as a string?

controllers/automation/policyautomation_controller.go
ChunxiAlexLuo: The original `eventTime` is the same as `AutomationStartTime` on the second level, so during testing, I add 1 second to ensure `eventTime` is ealier, since we use `preEventTime.After(originalStartTime)`, I can remove `adding 1 second `

controllers/automation/policyautomation_controller.go
mprahl: Ah, I see your point. Let's leave it as is.

controllers/automation/policyautomation_controller.go
```diff
@@ -199,6 +205,131 @@ func (r *PolicyAutomationReconciler) Reconcile(
...
+
+				if err != nil {
+					log.Error(err, "Failed to create the Ansible job")
+
+					return reconcile.Result{}, err
+				}
+
+				automationStartTimeStr := time.Now().UTC().Format(time.RFC3339)
+
+				for _, clusterName := range trimmedTargetList {
```
mprahl, line 305: What about requeuing when a new target is added to the list that wasn't in `eventMap`?

ChunxiAlexLuo: @mprahl So when a new target is added to the list that wasn't in `eventMap`, we created an ansible job immediately, any reason for `requeuing`.

ChunxiAlexLuo: > What about requeuing when a new target is added to the list that wasn't in `eventMap`?

Yes, you are correct. Need to re-queue when a new target is added to the list that wasn't in `eventMap`, since EventHook is only for `noncompliant`, not for `compliant`

mprahl: Actually, now that I'm thinking about it, maybe we don't need to since there is no action that can be taken unless it goes to compliant. That was my mistake.

test/e2e/case5_policy_automation_test.go
ChunxiAlexLuo: Meet some issues when patching the timestamp. Since it is a test time optimization, I reduced it to 150 seconds and will retry patching in another PR.

gparvin:
/hold
In case others want to check over this again.

gparvin (approved):
Thanks Alex!

mprahl (approved):
Looks good!

This is good to merge if you are planning to handle not waiting for `150` seconds in the `delayAfterRunSeconds` test in another PR.

PR #40, opened 2022-07-26
Change leader election to use the leases resource

gparvin:
Switching to using leases instead of configmapsleases for leader
election

Refs:
 - https://github.com/stolostron/backlog/issues/19110


dhaiducek (approved):
Wow. Is this all it takes? Approve? 😆

PR #47, opened 2022-09-30
Update to go version 1.19

ChunxiAlexLuo:


mprahl:
@ChunxiAlexLuo you'll need to remove these lines:
https://github.com/open-cluster-management-io/governance-policy-propagator/blob/4e7576965d782a8c7e0f37b27ed21ddd0236a36f/build/Dockerfile#L16-L18

And delete this file:
https://github.com/open-cluster-management-io/governance-policy-propagator/blob/main/build/template-delimiter.patch

This is no longer needed because Go 1.19 includes the Go stdlib fix.

mprahl (approved):

PR #52, opened 2022-10-18
Provide PolicyViolationContext to Ansible Tower

ChunxiAlexLuo:

https://github.com/open-cluster-management-io/enhancements/pull/73

### DIFF
--- a/api/v1beta1/policyautomation_types.go
+++ b/api/v1beta1/policyautomation_types.go
@@ -6,6 +6,8 @@ package v1beta1
 import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	runtime "k8s.io/apimachinery/pkg/runtime"
+
+	policyv1 "open-cluster-management.io/governance-policy-propagator/api/v1"
 )
 
 // PolicyAutomationSpec defines the desired state of PolicyAutomation
@@ -36,6 +38,8 @@ const (
 	Disabled   PolicyAutomationMode = "disabled"
 )
 
+const DefaultPolicyViolationContextLimit = 1000
+
 // AutomationDef defines the automation to invoke
 type AutomationDef struct {
 	// Type of the automation to invoke
@@ -52,6 +56,22 @@ type AutomationDef struct {
 	TowerSecret string `json:"secret"`
 	// JobTTL sets the time to live for the Kubernetes AnsibleJob object after the Ansible job run has finished.
 	JobTTL *int `json:"jobTtl,omitempty"`
+	// The maximum number of violating cluster contexts that will be provided to the Ansible job as extra variables.
+	// When policyViolationContextLimit is set to 0, it means no limit.
+	// The default value is 1000.
+	PolicyViolationContextLimit *uint `json:"policyViolationContextLimit,omitempty"`
+}
+
+// ViolationContext defines the non-compliant replicated policy information
+// that is sent to the AnsibleJob through extra_vars.
+type ViolationContext struct {
+	TargetClusters  []string `json:"targetClusters" ansibleJob:"target_clusters"`
+	PolicyName      string   `json:"policyName" ansibleJob:"policy_name"`
+	PolicyNamespace string   `json:"namespace"`
+	HubCluster      string   `json:"hubCluster" ansibleJob:"hub_cluster"`
+	PolicySets      []string `json:"policySet" ansibleJob:"policy_set"`
+	//nolint: lll
+	PolicyViolationContext map[string]ReplicatedPolicyStatus `json:"policyViolationContext" ansibleJob:"policy_violation_context"`
 }
 
 // PolicyAutomationStatus defines the observed state of PolicyAutomation
@@ -93,4 +113,23 @@ type ClusterEvent struct {
 
 func init() {
 	SchemeBuilder.Register(&PolicyAutomation{}, &PolicyAutomationList{})
+}
+
+// ReplicatedDetailsPerTemplate defines the replicated policy compliance details and history
+type ReplicatedDetailsPerTemplate struct {
+	ComplianceState policyv1.ComplianceState      `json:"compliant"`
+	History         []ReplicatedComplianceHistory `json:"history"`
+}
+
+// ReplicatedComplianceHistory defines the replicated policy compliance details history
+type ReplicatedComplianceHistory struct {
+	LastTimestamp metav1.Time `json:"lastTimestamp,omitempty" protobuf:"bytes,7,opt,name=lastTimestamp"`
+	Message       string      `json:"message,omitempty" protobuf:"bytes,4,opt,name=message"`
+}
+
+// ReplicatedPolicyStatus defines the replicated policy status
+type ReplicatedPolicyStatus struct {
+	ComplianceState  policyv1.ComplianceState       `json:"compliant"`        // used by replicated policy
+	ViolationMessage string                         `json:"violationMessage"` // used by replicated policy
+	Details          []ReplicatedDetailsPerTemplate `json:"details"`          // used by replicated policy
 }
--- a/controllers/automation/policyautomation_controller.go
+++ b/controllers/automation/policyautomation_controller.go
@@ -6,6 +6,8 @@ package automation
 import (
 	"context"
 	"fmt"
+	"net"
+	"strings"
 	"time"
 
 	"k8s.io/apimachinery/pkg/api/errors"
@@ -17,12 +19,14 @@ import (
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/builder"
 	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/client/config"
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
 	"sigs.k8s.io/controller-runtime/pkg/source"
 
 	policyv1 "open-cluster-management.io/governance-policy-propagator/api/v1"
 	policyv1beta1 "open-cluster-management.io/governance-policy-propagator/api/v1beta1"
 	"open-cluster-management.io/governance-policy-propagator/controllers/common"
+	"open-cluster-management.io/governance-policy-propagator/controllers/propagator"
 )
 
 const ControllerName string = "policy-automation"
@@ -89,6 +93,151 @@ func (r *PolicyAutomationReconciler) setOwnerReferences(
 	return nil
 }
 
+// getTargetListMap will convert slice targetList to map for search efficiency
+func getTargetListMap(targetList []string) map[string]bool {
+	targetListMap := map[string]bool{}
+	for _, target := range targetList {
+		targetListMap[target] = true
+	}
+
+	return targetListMap
+}
+
+// getClusterName will get the hub cluster name information
+func getClusterName() string {
+	hubCluster := ""
+	// Get a config to talk to the apiserver
+	cfg, err := config.GetConfig()
+	if err == nil {
+		hubCluster = cfg.Host
+	} else {
+		log.Error(err, "Unable to get the Kubernetes configuration.")
+	}
+
+	if len(hubCluster) > 0 {
+		// remove possible https head
+		hubCluster = strings.ReplaceAll(hubCluster, "https://", "")
+		// remove possible http head
+		hubCluster = strings.ReplaceAll(hubCluster, "http://", "")
+		// remove possible port number
+		hubCluster = strings.SplitN(hubCluster, ":", 2)[0]
+		// convert host name to URL if it is IP address
+		if net.ParseIP(hubCluster) != nil {
+			names, _ := net.LookupAddr(hubCluster)
+			if len(names) != 0 {
+				hubCluster = names[0]
+			}
+		}
+	}
+
+	log.V(3).Info("Hub Cluster information: %s", hubCluster)
+
+	return hubCluster
+}
+
+// getViolationContext will put the root policy information into violationContext
+// It also puts the status of the non-compliant replicated policies into violationContext
+func (r *PolicyAutomationReconciler) getViolationContext(
+	policy *policyv1.Policy,
+	targetList []string,
+	policyAutomation *policyv1beta1.PolicyAutomation,
+) (policyv1beta1.ViolationContext, error) {
+	log.V(3).Info(
+		"Get the violation context from the root policy %s/%s",
+		policy.GetNamespace(),
+		policy.GetName(),
+	)
+
+	violationContext := policyv1beta1.ViolationContext{}
+	// 1) get the target cluster list
+	violationContext.TargetClusters = targetList
+	// 2) get the root policy name
+	violationContext.PolicyName = policy.GetName()
+	// 3) get the root policy namespace
+	violationContext.PolicyNamespace = policy.GetNamespace()
+	// 4) get the root policy hub cluster name
+	violationContext.HubCluster = getClusterName()
+
+	// 5) get the policy sets of the root policy
+	plcPlacement := policy.Status.Placement
+	policySets := []string{}
+
+	for _, placement := range plcPlacement {
+		if placement.PolicySet != "" {
+			policySets = append(policySets, placement.PolicySet)
+		}
+	}
+
+	violationContext.PolicySets = policySets
+
+	replicatedPlcList := &policyv1.PolicyList{}
+
+	err := r.List(
+		context.TODO(),
+		replicatedPlcList,
+		client.MatchingLabels(propagator.LabelsForRootPolicy(policy)),
+	)
+	if err != nil {
+		log.Error(err, "Failed to list the replicated policies")
+
+		return violationContext, err
+	}
+
+	if len(replicatedPlcList.Items) == 0 {
+		log.V(2).Info("The replicated policies cannot be found.")
+
+		return violationContext, nil
+	}
+
+	policyViolationContextLimit := policyAutomation.Spec.Automation.PolicyViolationContextLimit
+	if policyViolationContextLimit == nil {
+		policyViolationContextLimit = new(uint)
+		*policyViolationContextLimit = policyv1beta1.DefaultPolicyViolationContextLimit
+	}
+
+	contextLimit := int(*policyViolationContextLimit)
+
+	targetListMap := getTargetListMap(targetList)
+	violationContext.PolicyViolationContext = make(
+		map[string]policyv1beta1.ReplicatedPolicyStatus,
+		len(replicatedPlcList.Items),
+	)
+
+	// 6) get the status of the non-compliance replicated policies
+	for _, rPlc := range replicatedPlcList.Items {
+		clusterName := rPlc.GetLabels()[common.ClusterNameLabel]
+		if !targetListMap[clusterName] {
+			continue // skip the compliance replicated policies
+		}
+
+		rPlcStatus := policyv1beta1.ReplicatedPolicyStatus{}
+		// Convert PolicyStatus to ReplicatedPolicyStatus and skip the unnecessary items
+		err := common.TypeConverter(rPlc.Status, &rPlcStatus)
+		if err != nil { // still assign the empty rPlcStatus to PolicyViolationContext later
+			log.Error(err, "The PolicyStatus cannot be converted to the type ReplicatedPolicyStatus.")
+		}
+
+		// get the latest violation message from the replicated policy
+		statusDetails := rPlc.Status.Details
+		if len(statusDetails) > 0 && len(statusDetails[0].History) > 0 {
+			rPlcStatus.ViolationMessage = statusDetails[0].History[0].Message
+		}
+
+		violationContext.PolicyViolationContext[clusterName] = rPlcStatus
+		if contextLimit > 0 && len(violationContext.PolicyViolationContext) == contextLimit {
+			log.V(2).Info(
+				"PolicyViolationContextLimit is %s so skipping %s remaining replicated policies violations.",
+				fmt.Sprint(contextLimit),
+				fmt.Sprint(len(replicatedPlcList.Items)-contextLimit),
+			)
+
+			break
+		}
+	}
+
+	return violationContext, nil
+}
+
 // Reconcile reads that state of the cluster for a Policy object and makes changes based on the state read
 // and what is in the Policy.Spec
 // Note:
@@ -150,7 +299,12 @@ func (r *PolicyAutomationReconciler) Reconcile(
 	if policyAutomation.Annotations["policy.open-cluster-management.io/rerun"] == "true" {
 		log.Info("Creating an Ansible job", "mode", "manual")
 
-		err = common.CreateAnsibleJob(policyAutomation, r.DynamicClient, "manual", nil)
+		err = common.CreateAnsibleJob(
+			policyAutomation,
+			r.DynamicClient,
+			"manual",
+			policyv1beta1.ViolationContext{},
+		)
 		if err != nil {
 			log.Error(err, "Failed to create the Ansible job", "mode", "manual")
 
@@ -194,8 +348,8 @@ func (r *PolicyAutomationReconciler) Reconcile(
 			targetList := common.FindNonCompliantClustersForPolicy(policy)
 			if len(targetList) > 0 {
 				log.Info("Creating An Ansible job", "targetList", targetList)
-
-				err = common.CreateAnsibleJob(policyAutomation, r.DynamicClient, "scan", targetList)
+				violationContext, _ := r.getViolationContext(policy, targetList, policyAutomation)
+				err = common.CreateAnsibleJob(policyAutomation, r.DynamicClient, "scan", violationContext)
 				if err != nil {
 					return reconcile.Result{RequeueAfter: requeueAfter}, err
 				}
@@ -216,11 +370,12 @@ func (r *PolicyAutomationReconciler) Reconcile(
 			if len(targetList) > 0 {
 				log.Info("Creating an Ansible job", "targetList", targetList)
 
+				violationContext, _ := r.getViolationContext(policy, targetList, policyAutomation)
 				err = common.CreateAnsibleJob(
 					policyAutomation,
 					r.DynamicClient,
 					string(policyv1beta1.Once),
-					targetList,
+					violationContext,
 				)
 
 				if err != nil {
@@ -243,11 +398,7 @@ func (r *PolicyAutomationReconciler) Reconcile(
 		} else if policyAutomation.Spec.Mode == policyv1beta1.EveryEvent {
 			log := log.WithValues("mode", string(policyv1beta1.EveryEvent))
 			targetList := common.FindNonCompliantClustersForPolicy(policy)
-			// Convert slice targetList to map for search efficiency
-			targetListMap := map[string]bool{}
-			for _, target := range targetList {
-				targetListMap[target] = true
-			}
+			targetListMap := getTargetListMap(targetList)
 			// The clusters map that the new ansible job will target
 			trimmedTargetMap := map[string]bool{}
 			// delayAfterRunSeconds and requeueDuration default value = zero
@@ -330,11 +481,12 @@ func (r *PolicyAutomationReconciler) Reconcile(
 					trimmedTargetList = append(trimmedTargetList, clusterName)
 				}
 				log.Info("Creating An Ansible job", "trimmedTargetList", trimmedTargetList)
+				violationContext, _ := r.getViolationContext(policy, trimmedTargetList, policyAutomation)
 				err = common.CreateAnsibleJob(
 					policyAutomation,
 					r.DynamicClient,
 					string(policyv1beta1.EveryEvent),
-					trimmedTargetList,
+					violationContext,
 				)
 
 				if err != nil {
--- a/controllers/common/ansible.go
+++ b/controllers/common/ansible.go
@@ -5,6 +5,7 @@ package common
 import (
 	"context"
 	"encoding/json"
+	"reflect"
 	"strings"
 
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
@@ -17,7 +18,7 @@ import (
 
 // CreateAnsibleJob creates ansiblejob with given PolicyAutomation
 func CreateAnsibleJob(policyAutomation *policyv1beta1.PolicyAutomation,
-	dynamicClient dynamic.Interface, mode string, targetClusters []string,
+	dynamicClient dynamic.Interface, mode string, violationContext policyv1beta1.ViolationContext,
 ) error {
 	ansibleJob := &unstructured.Unstructured{
 		Object: map[string]interface{}{
@@ -32,23 +33,36 @@ func CreateAnsibleJob(policyAutomation *policyv1beta1.PolicyAutomation,
 		},
 	}
 
+	mapExtraVars := map[string]interface{}{}
 	if policyAutomation.Spec.Automation.ExtraVars != nil {
 		// This is to translate the runtime.RawExtension to a map[string]interface{}
-		mapExtraVars := map[string]interface{}{}
-
 		err := json.Unmarshal(policyAutomation.Spec.Automation.ExtraVars.Raw, &mapExtraVars)
 		if err != nil {
 			return err
 		}
-
-		ansibleJob.Object["spec"].(map[string]interface{})["extra_vars"] = mapExtraVars
 	}
 
-	if targetClusters != nil {
-		// nolint: forcetypeassert
-		extravars := ansibleJob.Object["spec"].(map[string]interface{})["extra_vars"].(map[string]interface{})
-		extravars["target_clusters"] = targetClusters
+	values := reflect.ValueOf(violationContext)
+	typesOf := values.Type()
+	// add every violationContext fields into mapExtraVars as well as the empty values,
+	// or when the whole violationContext is empty
+	for i := 0; i < values.NumField(); i++ {
+		tag := typesOf.Field(i).Tag
+		value := values.Field(i).Interface()
+
+		var fieldName string
+		if tag.Get("ansibleJob") != "" {
+			fieldName = tag.Get("ansibleJob")
+		} else if tag.Get("json") != "" {
+			fieldName = strings.SplitN(tag.Get("json"), ",", 2)[0]
+		} else {
+			fieldName = typesOf.Field(i).Name
+		}
+
+		mapExtraVars[fieldName] = value
 	}
+
+	ansibleJob.Object["spec"].(map[string]interface{})["extra_vars"] = mapExtraVars
 
 	if policyAutomation.Spec.Automation.JobTTL != nil {
 		ansibleJob.Object["spec"].(map[string]interface{})["job_ttl"] = *policyAutomation.Spec.Automation.JobTTL
--- a/controllers/common/common.go
+++ b/controllers/common/common.go
@@ -5,6 +5,7 @@ package common
 
 import (
 	"context"
+	"encoding/json"
 	"errors"
 	"fmt"
 	"strings"
@@ -196,4 +197,14 @@ func ParseRootPolicyLabel(rootPlc string) (name, namespace string, err error) {
 	}
 
 	return rootSplit[1], rootSplit[0], nil
+}
+
+// TypeConverter is a helper function to converter type struct a to b
+func TypeConverter(a, b interface{}) error {
+	js, err := json.Marshal(a)
+	if err != nil {
+		return err
+	}
+
+	return json.Unmarshal(js, b)
 }
--- a/deploy/crds/policy.open-cluster-management.io_policyautomations.yaml
+++ b/deploy/crds/policy.open-cluster-management.io_policyautomations.yaml
@@ -55,6 +55,12 @@ spec:
                       job
                     minLength: 1
                     type: string
+                  policyViolationContextLimit:
+                    description: The maximum number of violating cluster contexts
+                      that will be provided to the Ansible job as extra variables.
+                      When policyViolationContextLimit is set to 0, it means no limit.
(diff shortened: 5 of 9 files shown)


test/e2e/case5_policy_automation_test.go
```diff
@@ -137,9 +142,29 @@ var _ = Describe("Test policy automation", func() {
...
+			extraVars := spec.(map[string]interface{})["extra_vars"].(map[string]interface{})
+			Expect(extraVars["policy_name"]).To(Equal("case5-test-policy"))
+			Expect(extraVars["namespace"]).To(Equal("policy-propagator-test"))
+			Expect(len(extraVars["target_clusters"].([]interface{}))).To(Equal(1))
+			Expect(extraVars["target_clusters"].([]interface{})[0]).To(Equal("managed1"))
+			Expect(len(extraVars["policy_set"].([]interface{}))).To(Equal(1))
+			Expect(extraVars["policy_set"].([]interface{})[0]).To(Equal("case5-test-policyset"))
+			managed1 := extraVars["policy_violation_context"].(map[string]interface{})["managed1"]
+			compliant := managed1.(map[string]interface{})["compliant"]
+			Expect(compliant).To(Equal(string(policiesv1.NonCompliant)))
```
gparvin, line 166: Is the hub cluster and the message from the violation context getting checked too?

gparvin (commented):
Double check all the fields are checked with the test.  Thanks!

test/e2e/case5_policy_automation_test.go
ChunxiAlexLuo: Good question. 

1) I found `message` is from `PolicyStatus.Details` which is used by the replicated policy so it will always empty when checking the root policy, so maybe we should change the design and move the `message` into `PolicyViolationContext` for each replicated policies?

2) I use policy.GetClusterName() to get the root policy hub cluster name and I think it doesn't work since there isn't a cluster name in the metadata. I'm looking for a better way to get the hub cluster name.

api/v1/policy_types.go
```diff
@@ -108,13 +108,26 @@ type DetailsPerTemplate struct {
 	History         []ComplianceHistory `json:"history,omitempty"`
 }
 
+// ReplicatedDetailsPerTemplate defines the replicated policy compliance details and history
+type ReplicatedDetailsPerTemplate struct {
```
mprahl, line 112: Could you please move this to `api/v1beta1/policyautomation_types.go` since this replicated policies actually do use `DetailsPerTemplate`?

api/v1/policy_types.go
```diff
@@ -108,13 +108,26 @@ type DetailsPerTemplate struct {
...
+
 // ComplianceHistory defines compliance details history
 type ComplianceHistory struct {
 	LastTimestamp metav1.Time `json:"lastTimestamp,omitempty" protobuf:"bytes,7,opt,name=lastTimestamp"`
 	Message       string      `json:"message,omitempty" protobuf:"bytes,4,opt,name=message"`
 	EventName     string      `json:"eventName,omitempty"`
 }
 
+// ReplicatedComplianceHistory defines the replicated policy compliance details history
+type ReplicatedComplianceHistory struct {
```
mprahl, line 126: Could you please move this to api/v1beta1/policyautomation_types.go since this replicated policies actually do use `ComplianceHistory`?

api/v1/policy_types.go
```diff
@@ -125,6 +138,13 @@ type PolicyStatus struct {
 	Details         []*DetailsPerTemplate `json:"details,omitempty"`   // used by replicated policy
 }
 
+// ReplicatedPolicyStatus defines the replicated policy status
+type ReplicatedPolicyStatus struct {
```
mprahl, line 142: Could you please move this to `api/v1beta1/policyautomation_types.go` since replicated policies actually use `PolicyStatus`?

api/v1beta1/policyautomation_types.go
```diff
@@ -52,6 +54,25 @@ type AutomationDef struct {
...
+// ViolationContext defines the non-compliance replicated policy information
+// that were sent to the Anisbile job through extra_var.
+type ViolationContext struct {
+	// +kubebuilder:validation:Required
+	// +kubebuilder:validation:MinLength=1
+	TargetClusters []string `json:"targetClusters,omitempty"`
+	// +kubebuilder:validation:Required
+	// +kubebuilder:validation:MinLength=1
+	PolicyName             string                                     `json:"policyName"`
+	Namespace              string                                     `json:"namespace,omitempty"`
```
mprahl, line 71: I think I'd rather not have `omitempty` on any of these. That way the Ansible user does not need to check if the key exists before checking the value.

api/v1beta1/policyautomation_types.go
```diff
@@ -52,6 +54,25 @@ type AutomationDef struct {
...
 	JobTTL *int `json:"jobTtl,omitempty"`
+	// The maximum number of violation contexts will be provided to the Ansible Tower as extra variable.
+	// +kubebuilder:validation:Minimum=0
+	PolicyViolationContextLimit uint `json:"policyViolationContextLimit,omitempty"`
+}
+
+// ViolationContext defines the non-compliance replicated policy information
+// that were sent to the Anisbile job through extra_var.
+type ViolationContext struct {
+	// +kubebuilder:validation:Required
```
mprahl, line 65: I would remove these `kubebuilder` markers since this struct won't be generated into a CRD.

api/v1beta1/policyautomation_types.go
```diff
@@ -52,6 +54,25 @@ type AutomationDef struct {
 	TowerSecret string `json:"secret"`
 	// JobTTL sets the time to live for the Kubernetes AnsibleJob object after the Ansible job run has finished.
 	JobTTL *int `json:"jobTtl,omitempty"`
+	// The maximum number of violation contexts will be provided to the Ansible Tower as extra variable.
+	// +kubebuilder:validation:Minimum=0
+	PolicyViolationContextLimit uint `json:"policyViolationContextLimit,omitempty"`
+}
+
+// ViolationContext defines the non-compliance replicated policy information
```
mprahl, line 62:
```suggestion
// ViolationContext defines the non-compliant replicated policy information
```

api/v1beta1/policyautomation_types.go
```diff
@@ -52,6 +54,25 @@ type AutomationDef struct {
...
+// that were sent to the Anisbile job through extra_var.
```
mprahl, line 63:
```suggestion
// that is sent to the AnsibleJob through extra_vars.
```

api/v1beta1/policyautomation_types.go
```diff
@@ -52,6 +54,25 @@ type AutomationDef struct {
 	TowerSecret string `json:"secret"`
 	// JobTTL sets the time to live for the Kubernetes AnsibleJob object after the Ansible job run has finished.
 	JobTTL *int `json:"jobTtl,omitempty"`
+	// The maximum number of violation contexts will be provided to the Ansible Tower as extra variable.
```
mprahl, line 57:
```suggestion
	// The maximum number of violating cluster contexts that will be provided to the Ansible job as extra variables.
```

mprahl: Can you document the default?

controllers/automation/policyautomation_controller.go
```diff
@@ -89,6 +89,100 @@ func (r *PolicyAutomationReconciler) setOwnerReferences(
...
+}
+
+// getViolationContext will put the root policy information into violationContext
+// It also put the status of the non-compliance replicated policies into violationContext
+func (r *PolicyAutomationReconciler) getViolationContext(
+	policy *policyv1.Policy,
+	targetList []string,
+	policyAutomation *policyv1beta1.PolicyAutomation,
+) (policyv1beta1.ViolationContext, error) {
+	log.V(3).Info("Get the violation context from the root policy %s", policy.ObjectMeta.Name)
```
mprahl, line 109: Could you please also include the namespace of the root policy?

controllers/automation/policyautomation_controller.go
```diff
@@ -89,6 +89,100 @@ func (r *PolicyAutomationReconciler) setOwnerReferences(
...
+
+	violationContext := policyv1beta1.ViolationContext{}
+	// 1) get the target cluster list
+	violationContext.TargetClusters = targetList
+	// 2) get the root policy name
+	violationContext.PolicyName = policy.ObjectMeta.Name
```
mprahl, line 115: Does this have `GetName()` so we are consistent with how you get the namespace?

controllers/automation/policyautomation_controller.go
```diff
@@ -89,6 +89,100 @@ func (r *PolicyAutomationReconciler) setOwnerReferences(
...
+	// 3) get the root policy namespace
+	violationContext.Namespace = policy.GetNamespace()
+	// 4) get the root policy hub cluster name
+	violationContext.HubCluster = policy.GetClusterName()
```
mprahl, line 119: Where is this defined? I'm having a hard time finding this in the diff.

controllers/automation/policyautomation_controller.go
```diff
@@ -89,6 +89,100 @@ func (r *PolicyAutomationReconciler) setOwnerReferences(
...
+
+	// 5) get the latest violation message from the root policy
+	statusDetails := policy.Status.Details
+	if len(statusDetails) > 0 && len(statusDetails[0].History) > 0 {
+		violationContext.ViolationMessage = statusDetails[0].History[0].Message
+	}
```
mprahl, line 125: I don't think the history is available on the root policy, so we can't include an overall policy message. Sorry about misstating that in the original GitHub issue.

controllers/automation/policyautomation_controller.go
```diff
@@ -89,6 +89,100 @@ func (r *PolicyAutomationReconciler) setOwnerReferences(
...
+
+	// 6) get the policy set of the root policy
```
mprahl, line 127:
```suggestion
	// 6) get the policy sets of the root policy
```

controllers/automation/policyautomation_controller.go
```diff
@@ -89,6 +89,100 @@ func (r *PolicyAutomationReconciler) setOwnerReferences(
...
+	plcPlacement := policy.Status.Placement
+	policySet := []string{}
```
mprahl, line 129:
```suggestion
	policySets := []string{}
```

api/v1beta1/policyautomation_types.go
mprahl, line 71: Should this be `PolicyNamespace`?

api/v1beta1/policyautomation_types.go
```diff
@@ -52,6 +54,25 @@ type AutomationDef struct {
...
+type ViolationContext struct {
+	// +kubebuilder:validation:Required
+	// +kubebuilder:validation:MinLength=1
+	TargetClusters []string `json:"targetClusters,omitempty"`
+	// +kubebuilder:validation:Required
+	// +kubebuilder:validation:MinLength=1
+	PolicyName             string                                     `json:"policyName"`
+	Namespace              string                                     `json:"namespace,omitempty"`
+	HubCluster             string                                     `json:"hubCluster,omitempty"`
+	PolicySet              []string                                   `json:"policySet,omitempty"`
```
mprahl, line 73:
```suggestion
	PolicySets              []string                                   `json:"policySet,omitempty"`
```

controllers/automation/policyautomation_controller.go
```diff
@@ -89,6 +89,100 @@ func (r *PolicyAutomationReconciler) setOwnerReferences(
...
+
+	for _, placement := range plcPlacement {
+		if placement.PolicySet != "" {
+			policySet = append(policySet, placement.PolicySet)
+		}
+	}
+
+	if len(policySet) > 0 {
+		violationContext.PolicySet = policySet
+	}
```
mprahl, line 139: The if statement isn't needed in my opinion.
```suggestion
	violationContext.PolicySet = policySet
```

controllers/automation/policyautomation_controller.go
```diff
@@ -89,6 +89,100 @@ func (r *PolicyAutomationReconciler) setOwnerReferences(
...
+		client.MatchingLabels(common.LabelsForRootPolicy(policy)),
+	)
+	if err != nil {
+		log.Error(err, "Failed to list the replicated policies")
+
+		return violationContext, err
+	}
+
+	if len(replicatedPlcList.Items) == 0 {
+		log.V(2).Info("The replicated policies can not be found.")
```
mprahl, line 155: Optional: This is the far more common spelling.
```suggestion
		log.V(2).Info("The replicated policies cannot be found.")
```

mprahl, line 155: Could you include a return after this so you can get rid of the `else` below? It makes it easier to read without additional indentation.

controllers/automation/policyautomation_controller.go
```diff
@@ -89,6 +89,100 @@ func (r *PolicyAutomationReconciler) setOwnerReferences(
...
+	} else {
+		policyViolationContextLimit := policyAutomation.Spec.Automation.PolicyViolationContextLimit
+		if policyViolationContextLimit == 0 {
+			policyViolationContextLimit = 1000
+		}
+		targetListMap := getTargetListMap(targetList)
+		violationContext.PolicyViolationContext = make(map[string]policyv1.ReplicatedPolicyStatus)
```
mprahl, line 162: Optional: A small performance optimization.
```suggestion
		violationContext.PolicyViolationContext = make(map[string]policyv1.ReplicatedPolicyStatus, 0, len(replicatedPlcList.Items))
```

api/v1beta1/policyautomation_types.go
mprahl: It'd be even better if it was a null value rather than an empty string:
https://stackoverflow.com/a/31048860

controllers/automation/policyautomation_controller.go
```diff
@@ -89,6 +89,100 @@ func (r *PolicyAutomationReconciler) setOwnerReferences(
...
+		for _, rPlc := range replicatedPlcList.Items {
+			clusterName := rPlc.GetLabels()[common.ClusterNameLabel]
+			// 7) get the status of the non-compliance replicated policies
+			if targetListMap[clusterName] {
```
mprahl, line 166: Optional: You can remove another indentation by doing this:

```go
if !targetListMap[clusterName] {
	continue
}
...
```

controllers/common/common.go
```diff
@@ -201,3 +204,25 @@ func GetNumWorkers(listLength int, concurrencyPerPolicy int) int {
 
 	return numWorkers
 }
+
+// TypeConverter is a helper function to converter type struct a to b
+func TypeConverter(a, b interface{}) error {
```
mprahl, line 209: Cool!

controllers/automation/policyautomation_controller.go
```diff
@@ -89,6 +89,100 @@ func (r *PolicyAutomationReconciler) setOwnerReferences(
...
+				// Convert PolicyStatus to ReplicatedPolicyStatus and skip the unnecessary items
+				err := common.TypeConverter(rPlc.Status, &rPlcStatus)
+				if err != nil {
+					log.Error(err, "The PolicyStatus can not be convert to type of ReplicatedPolicyStatus.")
+
+					continue
+				}
+				violationContext.PolicyViolationContext[clusterName] = rPlcStatus
+				if len(violationContext.PolicyViolationContext) == int(policyViolationContextLimit) {
+					break
```
mprahl, line 177: It'd be good to log that you are skipping X remaining policy violations.

controllers/automation/policyautomation_controller.go
```diff
@@ -89,6 +89,100 @@ func (r *PolicyAutomationReconciler) setOwnerReferences(
...
+		violationContext.PolicyViolationContext = make(map[string]policyv1.ReplicatedPolicyStatus)
+		for _, rPlc := range replicatedPlcList.Items {
+			clusterName := rPlc.GetLabels()[common.ClusterNameLabel]
+			// 7) get the status of the non-compliance replicated policies
+			if targetListMap[clusterName] {
+				rPlcStatus := policyv1.ReplicatedPolicyStatus{}
+				// Convert PolicyStatus to ReplicatedPolicyStatus and skip the unnecessary items
+				err := common.TypeConverter(rPlc.Status, &rPlcStatus)
+				if err != nil {
+					log.Error(err, "The PolicyStatus can not be convert to type of ReplicatedPolicyStatus.")
```
mprahl, line 171:
```suggestion
					log.Error(err, "The PolicyStatus cannot be converted to the type ReplicatedPolicyStatus.")
```

controllers/common/ansible.go
```diff
@@ -44,10 +45,19 @@ func CreateAnsibleJob(policyAutomation *policyv1beta1.PolicyAutomation,
...
-		// nolint: forcetypeassert
+	if violationContext.TargetClusters != nil {
 		extravars := ansibleJob.Object["spec"].(map[string]interface{})["extra_vars"].(map[string]interface{})
-		extravars["target_clusters"] = targetClusters
+		values := reflect.ValueOf(violationContext)
+		typesOf := values.Type()
+		// add every violationContext fields into extravars
+		// here use the snake key to match the design and
+		// don't need to set target_clusters specifically
+		for i := 0; i < values.NumField(); i++ {
```
mprahl, line 55: Optional: Could you perhaps define a new tag (e.g. `ansiblejob`) on each struct field and use that rather than try to automatically convert? Just an idea if we want to be explicit.

controllers/common/common.go
```diff
@@ -201,3 +204,25 @@ func GetNumWorkers(listLength int, concurrencyPerPolicy int) int {
...
+	return json.Unmarshal(js, b)
+}
+
+var (
+	matchFirstCap = regexp.MustCompile("(.)([A-Z][a-z]+)")
+	matchAllCap   = regexp.MustCompile("([a-z0-9])([A-Z])")
+)
+
+func ToSnakeCase(str string) string {
+	snake := matchFirstCap.ReplaceAllString(str, "${1}_${2}")
```
mprahl, line 224: Could you please explain in this comment how this works? I'm not quite following.

controllers/common/common.go
```diff
@@ -201,3 +204,25 @@ func GetNumWorkers(listLength int, concurrencyPerPolicy int) int {
...
+
+	return json.Unmarshal(js, b)
+}
+
+var (
+	matchFirstCap = regexp.MustCompile("(.)([A-Z][a-z]+)")
+	matchAllCap   = regexp.MustCompile("([a-z0-9])([A-Z])")
+)
+
+func ToSnakeCase(str string) string {
```
mprahl, line 223: Could you please add unit tests for this function?

api/v1beta1/policyautomation_types.go
```diff
@@ -52,6 +54,25 @@ type AutomationDef struct {
 	TowerSecret string `json:"secret"`
 	// JobTTL sets the time to live for the Kubernetes AnsibleJob object after the Ansible job run has finished.
 	JobTTL *int `json:"jobTtl,omitempty"`
+	// The maximum number of violation contexts will be provided to the Ansible Tower as extra variable.
+	// +kubebuilder:validation:Minimum=0
```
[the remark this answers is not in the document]
ChunxiAlexLuo: @mprahl  Both `the user set PolicyViolationContextLimit to 0` and `the use doesn't set PolicyViolationContextLimit` cases will use the default value 1000 since `unset is also zero here`. We can use other default values than 1000 or remove the limitation.  But I'm not sure if there is a need or a good way to treat `PolicyViolationContextLimit = null` and `PolicyViolationContextLimit = 0` as different cases.

api/v1beta1/policyautomation_types.go
ChunxiAlexLuo: Will do this after we decide on how to deal with the default case.

api/v1beta1/policyautomation_types.go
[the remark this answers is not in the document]
mprahl: Could we use a default here with the `+kubebuilder:default` marker? Then we could consider having 0 be considered no limit.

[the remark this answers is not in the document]
ChunxiAlexLuo: @mprahl So there are two solutions:

1）When the user set `PolicyViolationContextLimit = 0` or doesn‘t set `PolicyViolationContextLimit` (the default value is `zero`),  there isn't limitation.

2) When the user set `PolicyViolationContextLimit = 0` or doesn‘t set `PolicyViolationContextLimit` (the default value is `zero`), then use the default limitation = 1000.

controllers/automation/policyautomation_controller.go
ChunxiAlexLuo: This is the function we discuss the logic yesterday.

controllers/automation/policyautomation_controller.go
ChunxiAlexLuo: Also discussed this with @gparvin, so either we remove the `ViolationMessage`, or we get the latest history message for each replicated policy and put them under `PolicyViolationContext` per `clusterName`

api/v1beta1/policyautomation_types.go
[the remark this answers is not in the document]
mprahl: I think the behavior that we want is the user explicitly specifying 0 causes there to be no limit, otherwise, it'll default to `1000` or perhaps a smaller number.

controllers/automation/policyautomation_controller.go
mprahl: I like the second option.

deploy/crds/policy.open-cluster-management.io_policyautomations.yaml
```diff
@@ -55,6 +55,13 @@ spec:
                       job
                     minLength: 1
                     type: string
+                  policyViolationContextLimit:
+                    description: The maximum number of violating cluster contexts
+                      that will be provided to the Ansible job as extra variables.
+                      PolicyViolationContextLimit = 0 means no limitation and the
+                      default value is DefaultPolicyViolationContextLimit.
```
gparvin, line 62: If a customer looks at the CRD will they see what `DefaultPolicyViolationContextLimit` is set to?  I think you have to show that value.

mprahl:
@ChunxiAlexLuo could you please update the upstream design document with the API changes around the compliance message? Then we can tag QE to be aware of it for their test development.

api/v1beta1/policyautomation_types.go
```diff
@@ -94,3 +114,24 @@ type ClusterEvent struct {
...
+
+// ReplicatedComplianceHistory defines the replicated policy compliance details history
+type ReplicatedComplianceHistory struct {
+	LastTimestamp metav1.Time `json:"lastTimestamp,omitempty" protobuf:"bytes,7,opt,name=lastTimestamp"`
+	Message       string      `json:"message,omitempty" protobuf:"bytes,4,opt,name=message"`
+}
+
+// ReplicatedPolicyStatus defines the replicated policy status
+type ReplicatedPolicyStatus struct {
+	// +kubebuilder:validation:Enum=Compliant;NonCompliant
```
mprahl, line 134: Could you please remove this kube builder marker since this struct isn't generated into a CRD?

controllers/automation/policyautomation_controller.go
```diff
@@ -89,6 +90,136 @@ func (r *PolicyAutomationReconciler) setOwnerReferences(
...
+// getClusterName will get the hub cluster name information
+func getClusterName() string {
+	// Get a config to talk to the apiserver
+	cfg := config.GetConfigOrDie()
+
+	var hubCluster string
+	if len(cfg.Host) > 0 {
+		hubCluster = cfg.Host
+	} else {
+		hubCluster = cfg.APIPath
```
mprahl, line 112: I think we would just stick to an empty string in this case.

api/v1beta1/policyautomation_types.go
```diff
@@ -52,6 +56,23 @@ type AutomationDef struct {
 	TowerSecret string `json:"secret"`
 	// JobTTL sets the time to live for the Kubernetes AnsibleJob object after the Ansible job run has finished.
 	JobTTL *int `json:"jobTtl,omitempty"`
+	// The maximum number of violating cluster contexts that will be provided to the Ansible job as extra variables.
+	// PolicyViolationContextLimit = 0 means no limitation.
+	// The default value DefaultPolicyViolationContextLimit = 1000.
```
mprahl, line 61:
```suggestion
	// The maximum number of violating cluster contexts that will be provided to the Ansible job as extra variables.
	// When policyViolationContextLimit is set to 0, it means no limit.
	// The default value is 1000.
```

api/v1beta1/policyautomation_types.go
```diff
@@ -94,3 +114,24 @@ type ClusterEvent struct {
...
+
+// ReplicatedDetailsPerTemplate defines the replicated policy compliance details and history
+type ReplicatedDetailsPerTemplate struct {
+	// +kubebuilder:pruning:PreserveUnknownFields
+	ComplianceState policyv1.ComplianceState      `json:"compliant,omitempty"`
+	History         []ReplicatedComplianceHistory `json:"history,omitempty"`
+}
+
+// ReplicatedComplianceHistory defines the replicated policy compliance details history
+type ReplicatedComplianceHistory struct {
```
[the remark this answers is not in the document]
ChunxiAlexLuo: An example after removing `omitempty`  as the suggestion

```
spec:
  extra_vars:
    hub_cluster: kubernetes.default.svc.cluster.local.
    namespace: policy-propagator-test
    policy_name: case5-test-policy
    policy_set:
    - case5-test-policyset
    policy_violation_context:
      managed1:
        compliant: NonCompliant
        details: null
        violationMessage: ""
    sn_priority: 1
    sn_severity: 1
    target_clusters:
    - managed1
  job_template_name: Demo Job Template
  job_ttl: 86400
  tower_auth_secret: toweraccess
```

ChunxiAlexLuo:
> @ChunxiAlexLuo could you please update the upstream design document with the API changes around the compliance message? Then we can tag QE to be aware of it for their test development.

@mprahl opened a upstream design document update PR for this https://github.com/open-cluster-management-io/enhancements/pull/79

api/v1beta1/policyautomation_types.go
```diff
@@ -94,3 +114,22 @@ type ClusterEvent struct {
...
+	LastTimestamp metav1.Time `json:"lastTimestamp,omitempty" protobuf:"bytes,7,opt,name=lastTimestamp"`
+	Message       string      `json:"message,omitempty" protobuf:"bytes,4,opt,name=message"`
+}
+
+// ReplicatedPolicyStatus defines the replicated policy status
+type ReplicatedPolicyStatus struct {
+	ComplianceState  policyv1.ComplianceState        `json:"compliant"`        // used by replicated policy
+	ViolationMessage string                          `json:"violationMessage"` // used by replicated policy
+	Details          []*ReplicatedDetailsPerTemplate `json:"details"`          // used by replicated policy
```
mprahl, line 134: We want an empty list in this case when converted to an extra var. That way I don't have check the type in my Ansible playbook. I can just loop over it.
```suggestion
	Details          []ReplicatedDetailsPerTemplate `json:"details"`          // used by replicated policy
```

api/v1beta1/policyautomation_types.go
```diff
@@ -52,6 +56,22 @@ type AutomationDef struct {
...
+	// The maximum number of violating cluster contexts that will be provided to the Ansible job as extra variables.
+	// When policyViolationContextLimit is set to 0, it means no limit.
+	// The default value is 1000.
+	PolicyViolationContextLimit *uint `json:"policyViolationContextLimit,omitempty"`
+}
+
+// ViolationContext defines the non-compliant replicated policy information
+// that is sent to the AnsibleJob through extra_vars.
+type ViolationContext struct {
+	TargetClusters  []string `json:"targetClusters,omitempty" ansibleJob:"target_clusters"`
```
mprahl, line 68: I know this is never empty, but it'd be good to be consistent.
```suggestion
	TargetClusters  []string `json:"targetClusters" ansibleJob:"target_clusters"`
```

controllers/automation/policyautomation_controller.go
```diff
@@ -89,6 +92,150 @@ func (r *PolicyAutomationReconciler) setOwnerReferences(
...
+		targetListMap[target] = true
+	}
+
+	return targetListMap
+}
+
+// getClusterName will get the hub cluster name information
+func getClusterName() string {
+	// Get a config to talk to the apiserver
+	cfg := config.GetConfigOrDie()
```
mprahl, line 108: In theory this should never fail, but I don't think a panic is necessary here. I think returning an empty string if the config can't be retrieved would be preferable.

controllers/automation/policyautomation_controller.go
```diff
@@ -89,6 +92,150 @@ func (r *PolicyAutomationReconciler) setOwnerReferences(
...
+	hostStr := cfg.Host
```
mprahl, line 109: I don't think you need this separate variable. I think you could just always use `hubCluster`.

controllers/automation/policyautomation_controller.go
```diff
@@ -89,6 +92,150 @@ func (r *PolicyAutomationReconciler) setOwnerReferences(
...
+		hostStr = strings.ReplaceAll(hostStr, "http://", "")
+		// remove possible port number
+		hostStr = strings.SplitN(hostStr, ":", 2)[0]
+		hubCluster = hostStr
+		// convert host name to URL if it is IP address
+		if net.ParseIP(hostStr) != nil {
+			hostAddr, err := net.LookupAddr(hostStr)
+			if err == nil {
+				hubCluster = hostAddr[0]
+			}
```
mprahl, line 125: After looking at the docs, it's unclear when an error is returned but it does seem that an error can be returned even when some results are returned, so I think this is safer:
```suggestion
			names, _ := net.LookupAddr(hostStr)
			if len(names) > 1 {
				hubCluster = names[0]
			}
```

controllers/automation/policyautomation_controller.go
```diff
@@ -89,6 +92,150 @@ func (r *PolicyAutomationReconciler) setOwnerReferences(
...
+		}
+	}
+
+	log.V(3).Info("Hub Cluster information: %s", hubCluster)
+
+	return hubCluster
+}
+
+// getViolationContext will put the root policy information into violationContext
+// It also put the status of the non-compliance replicated policies into violationContext
```
mprahl, line 135:
```suggestion
// It also puts the status of the non-compliant replicated policies into violationContext
```

controllers/automation/policyautomation_controller.go
```diff
@@ -89,6 +92,150 @@ func (r *PolicyAutomationReconciler) setOwnerReferences(
...
+			continue // skip the compliance replicated policies
+		}
+
+		rPlcStatus := policyv1beta1.ReplicatedPolicyStatus{}
+		// Convert PolicyStatus to ReplicatedPolicyStatus and skip the unnecessary items
+		err := common.TypeConverter(rPlc.Status, &rPlcStatus)
+		if err != nil {
+			log.Error(err, "The PolicyStatus cannot be converted to the type ReplicatedPolicyStatus.")
+
+			continue
```
mprahl, line 215: I'm actually thinking an empty `rPlcStatus` would be better than skipping it entirely.

controllers/common/ansible.go
```diff
@@ -44,10 +45,28 @@ func CreateAnsibleJob(policyAutomation *policyv1beta1.PolicyAutomation,
 		ansibleJob.Object["spec"].(map[string]interface{})["extra_vars"] = mapExtraVars
 	}
 
-	if targetClusters != nil {
-		// nolint: forcetypeassert
+	if violationContext.TargetClusters != nil {
```
mprahl, line 48: If it's nil, could you please use an empty map instead? This makes the extra vars "API" more consistent.

test/e2e/case5_policy_automation_test.go
```diff
@@ -137,9 +142,33 @@ var _ = Describe("Test policy automation", func() {
 				Expect(err).To(BeNil())
 				_, err = utils.KubectlWithOutput("get", "ansiblejobs", "-n", testNamespace)
 				Expect(err).Should(BeNil())
+				spec := ansiblejobList.Items[0].Object["spec"]
+				extraVars := spec.(map[string]interface{})["extra_vars"].(map[string]interface{})
+
+				return len(extraVars)
+			}, 30, 1).Should(Equal(8))
+
+			By("Check each violation context field in extraVars")
```
mprahl, line 151:
```suggestion
			By("Check each violation context field in extra_vars")
```

test/e2e/case5_policy_automation_test.go
```diff
@@ -137,9 +142,33 @@ var _ = Describe("Test policy automation", func() {
...
+			ansiblejobList, err := clientHubDynamic.Resource(gvrAnsibleJob).Namespace(testNamespace).List(
+				context.TODO(), metav1.ListOptions{},
+			)
+			Expect(err).To(BeNil())
+			spec := ansiblejobList.Items[0].Object["spec"]
+			extraVars := spec.(map[string]interface{})["extra_vars"].(map[string]interface{})
+			Expect(extraVars["policy_name"]).To(Equal("case5-test-policy"))
+			Expect(extraVars["namespace"]).To(Equal("policy-propagator-test"))
```
mprahl, line 159: I think `policy-propagator-test` is stored as a global variable. Could you please use the variable if this is the case?

controllers/common/ansible.go
ChunxiAlexLuo:
```
  extra_vars:
    hub_cluster: ""
    namespace: ""
    policy_name: ""
    policy_set: null
    policy_violation_context: null
    sn_priority: 1
    sn_severity: 1
    target_clusters: null
```

controllers/propagator/replication_utils.go
```diff
@@ -16,7 +16,7 @@ import (
 )
 
 // labelsForRootPolicy returns the labels for given policy
-func labelsForRootPolicy(plc *policiesv1.Policy) map[string]string {
+func LabelsForRootPolicy(plc *policiesv1.Policy) map[string]string {
```
ChunxiAlexLuo, line 19: https://github.com/open-cluster-management-io/governance-policy-propagator/pull/58

@JustinKuli @mprahl  I know neither this way nor making some similar code in `policyautomation_controller.go` looks nice, let me know your preference.

test/e2e/case5_policy_automation_test.go
```diff
@@ -137,9 +142,33 @@ var _ = Describe("Test policy automation", func() {
...
+			// rather than verifying the hub_cluster name
+			By("Check hub_cluster : " + extraVars["hub_cluster"].(string))
+			Expect(len(extraVars["hub_cluster"].(string)) > 0).To(BeTrue())
+			Expect(len(extraVars["target_clusters"].([]interface{}))).To(Equal(1))
+			Expect(extraVars["target_clusters"].([]interface{})[0]).To(Equal("managed1"))
+			Expect(len(extraVars["policy_set"].([]interface{}))).To(Equal(1))
+			Expect(extraVars["policy_set"].([]interface{})[0]).To(Equal("case5-test-policyset"))
+			managed1 := extraVars["policy_violation_context"].(map[string]interface{})["managed1"]
+			compliant := managed1.(map[string]interface{})["compliant"]
+			Expect(compliant).To(Equal(string(policiesv1.NonCompliant)))
```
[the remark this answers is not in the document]
ChunxiAlexLuo: @mprahl Do you mean ViolationMessage or Details in ReplicatedPolicyStatus? They are not guaranteed so I don't check them.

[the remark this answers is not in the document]
mprahl: Yes, I do. Could you please set the history in the policy's status in the `UpdateStatus` call above? Then it will be guaranteed.

controllers/propagator/replication_utils.go
mprahl: This is fine from my perspective.

JustinKuli: Yeah, seems fine to me as well

controllers/automation/policyautomation_controller.go
```diff
@@ -89,6 +93,152 @@ func (r *PolicyAutomationReconciler) setOwnerReferences(
...
+	return targetListMap
+}
+
+// getClusterName will get the hub cluster name information
+func getClusterName() string {
+	hubCluster := ""
+	// Get a config to talk to the apiserver
+	cfg, err := config.GetConfig()
+	if err == nil {
+		hubCluster = cfg.Host
```
mprahl, line 112: Could you please log the error?

controllers/common/ansible.go
mprahl: @ChunxiAlexLuo is this actually the case? I was expecting `policy_violation_context` to be `{}` and `target_clusters` to be `[]`.

ChunxiAlexLuo: @mprahl Yes, this is the actual data for `manual` mode. When `targetList` is empty, getViolationContext() will not be called and CreateAnsibleJob() only be called under `manual` mode. Since we need to keep the `mapExtraVars` value from `policyAutomation.Spec.Automation.ExtraVars`, I insert the value from `violationContext` into `mapExtraVars`.

```
		log.Info("Creating an Ansible job", "mode", "manual")

		err = common.CreateAnsibleJob(
			policyAutomation,
			r.DynamicClient,
			"manual",
			policyv1beta1.ViolationContext{},
		)
```

mprahl: Okay, so the `mapExtraVars[fieldName] = value` line must just be assigning `nil` instead of `[]`/`{}` if the slice/map is empty. Is there a way to guard against this?

ChunxiAlexLuo: @mprahl Tested some ways but don't find an easy way,  except the way that set these fields to `[]/{}` later when checking them are `nil`, which looks ugly and inflexible when adding new fields.

mprahl (approved):
Nice!

PR #57, opened 2022-10-25
Initialize klog explicitly

mprahl:
Updating go-template-utils causes the klog flags to not be initialized implicitly.

Signed-off-by: mprahl <mprahl@users.noreply.github.com>

### DIFF
--- a/main.go
+++ b/main.go
@@ -75,6 +75,8 @@ func init() {
 }
 
 func main() {
+	klog.InitFlags(nil)
+
 	zflags := zaputil.FlagConfig{
 		LevelName:   "log-level",
 		EncoderName: "log-encoder",


ChunxiAlexLuo (approved):
/lgtm

JustinKuli (commented):
What about https://github.com/open-cluster-management-io/governance-policy-propagator/blob/main/main.go#L120 ?

mprahl:
> What about https://github.com/open-cluster-management-io/governance-policy-propagator/blob/main/main.go#L120 ?

That is called after `flag.Parse()` and it's also not binding it to `flag.CommandLine`, so that doesn't resolve this issue.

The values from `flag.CommandLine` do make their way into the `klogFlags` flag set.

PR #58, opened 2022-10-26
Policyset dependencies

JustinKuli:
Some refactoring to (I think) get things in better shape for the changes, and then this is the main commit message:

Since whether dependencies are satisfied will be calculated on the managed cluster, and since PolicySets are not propagated to managed clusters, in order to work "as expected" by a user, any PolicySet dependencies should be replaced with the Policies that are in that set.

To keep those lists up to date, the root policies need to reconcile whenever the dependent PolicySets are updated, accomplished here with the new dynamic watch library. So now Policies watch items referenced in their templates, as well as relevant PolicySets.

Note: when a PolicySet does not exist, the dependency is not changed - it is expected that the status message from the managed cluster (indicating that the set does not exist) will be sufficient for users to understand the issue.

Refs:
 - https://github.com/stolostron/backlog/issues/26181

### DIFF
--- a/controllers/propagator/replication_utils.go
+++ b/controllers/propagator/replication_utils.go
@@ -1,0 +1,60 @@
+package propagator
+
+import (
+	"k8s.io/apimachinery/pkg/api/equality"
+	policiesv1 "open-cluster-management.io/governance-policy-propagator/api/v1"
+	"open-cluster-management.io/governance-policy-propagator/controllers/common"
+	appsv1 "open-cluster-management.io/multicloud-operators-subscription/pkg/apis/apps/placementrule/v1"
+)
+
+// labelsForRootPolicy returns the labels for given policy
+func labelsForRootPolicy(plc *policiesv1.Policy) map[string]string {
+	return map[string]string{common.RootPolicyLabel: fullNameForPolicy(plc)}
+}
+
+// fullNameForPolicy returns the fully qualified name for given policy
+// full qualified name: ${namespace}.${name}
+func fullNameForPolicy(plc *policiesv1.Policy) string {
+	return plc.GetNamespace() + "." + plc.GetName()
+}
+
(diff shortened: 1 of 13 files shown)


JustinKuli:
```
Pushing image to KinD cluster
kind load docker-image localhost:5000/governance-policy-propagator:latest --name test-hub
Image: "localhost:5000/governance-policy-propagator:latest" with ID "sha256:0687892eab0a3f0e9c50076a7cfa82b0fc755a742eb93a1f28f94985f0550552" not yet present on node "test-hub-control-plane", loading...
ERROR: failed to load image: command "docker exec --privileged -i test-hub-control-plane ctr --namespace=k8s.io images import --digests --snapshotter=overlayfs -" failed with error: exit status 1

Command Output: ctr: failed to resolve layers: failure checking for compressed blobs: failed to detect media type of layer: failed to read content store to detect layer media type: content digest sha256:8cedbb6c20a1b8c838c2c7e8bc3c9ef01a35ddcae4c7673bc7cda880922facae: not found
make: *** [Makefile:230: kind-deploy-controller-dev] Error 1
```
Fun 😆

controllers/propagator/replication_utils.go
```diff
@@ -0,0 +1,184 @@
...
+			template.ExtraDependencies, root.Namespace)
+		if err != nil {
+			return replicated, err
+		}
+	}
+
+	return replicated, nil
+}
+
+func (r *PolicyReconciler) canonicalizeDependencies(
```
willkutler, line 87: I think a comment describing what this function does would be useful, since the formatted dependency array is not really used in this repo so it might be a bit confusing why we are changing it

willkutler (approved):

willkutler:
This looks good to me! Just added a nit about adding some comments to clarify what these dependency formatting functions do.

PR #61, opened 2022-10-27
Add `hub_template_active_watches` metric

dhaiducek:
Adds a Gauge Vector that records the number of active watches for each root policy.

Addresses:
- https://github.com/stolostron/backlog/issues/25866

### DIFF
--- a/controllers/propagator/policy_controller.go
+++ b/controllers/propagator/policy_controller.go
@@ -98,6 +98,14 @@ type PolicyReconciler struct {
 func (r *PolicyReconciler) Reconcile(ctx context.Context, request ctrl.Request) (ctrl.Result, error) {
 	log := log.WithValues("Request.Namespace", request.Namespace, "Request.Name", request.Name)
 
+	// Set the hub template watch metric after reconcile
+	defer func() {
+		hubTempWatches := r.DynamicWatcher.GetWatchCount()
+		log.V(3).Info("Setting hub template watch metric", "value", hubTempWatches)
+
+		hubTemplateActiveWatchesMetric.Set(float64(hubTempWatches))
+	}()
+
 	log.Info("Reconciling the policy")
 
 	// Fetch the Policy instance
--- a/controllers/propagator/propagation.go
+++ b/controllers/propagator/propagation.go
@@ -187,7 +187,8 @@ func (r *PolicyReconciler) cleanUpPolicy(instance *policiesv1.Policy) error {
 		return nil
 	}
 
-	log.V(2).Info("Deleting %d replicated policies because root policy was deleted", len(replicatedPlcList.Items))
+	log.V(2).Info(
+		"Deleting replicated policies because root policy was deleted", "count", len(replicatedPlcList.Items))
 
 	policiesChan := make(chan policiesv1.Policy, len(replicatedPlcList.Items))
 	deletionResultsChan := make(chan deletionResult, len(replicatedPlcList.Items))
@@ -532,6 +533,7 @@ func (r *PolicyReconciler) recordWarning(instance *policiesv1.Policy, msgPrefix 
 // method because it makes the retries more targeted and prevents race conditions, such as a
 // placement binding getting updated, from causing inconsistencies.
 func (r *PolicyReconciler) handleRootPolicy(instance *policiesv1.Policy) error {
+	// Generate a metric for elapsed handling time for each policy
 	entryTS := time.Now()
 	defer func() {
 		now := time.Now()
--- a/test/e2e/case14_templates_metrics_test.go
+++ b/test/e2e/case14_templates_metrics_test.go
@@ -1,0 +1,92 @@
+// Copyright (c) 2021 Red Hat, Inc.
+// Copyright Contributors to the Open Cluster Management project
+
+package e2e
+
+import (
+	"context"
+
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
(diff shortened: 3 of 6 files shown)


controllers/propagator/policy_controller.go
```diff
@@ -124,6 +125,19 @@ func (r *PolicyReconciler) Reconcile(ctx context.Context, request ctrl.Request)
 				return reconcile.Result{}, err
 			}
 
+			// Clean up hub template watch metric
+			log.V(3).Info("Deleting hub template watch metric")
+
+			metricDeleted := hubTemplateActiveWatchesMetric.Delete(prometheus.Labels{
```
mprahl, line 131: Do you think this would be better in `cleanUpPolicy`?

controllers/propagator/propagation.go
```diff
@@ -187,7 +187,8 @@ func (r *PolicyReconciler) cleanUpPolicy(instance *policiesv1.Policy) error {
 		return nil
 	}
 
-	log.V(2).Info("Deleting %d replicated policies because root policy was deleted", len(replicatedPlcList.Items))
+	log.V(2).Info(
+		fmt.Sprintf("Deleting %d replicated policies because root policy was deleted", len(replicatedPlcList.Items)))
```
mprahl, line 191: Good catch, though I think it should be key-value pairs like:
```suggestion
		"Deleting replicated policies because root policy was deleted", "count", len(replicatedPlcList.Items))
```

controllers/propagator/policy_controller.go
```diff
@@ -160,6 +174,20 @@ func (r *PolicyReconciler) Reconcile(ctx context.Context, request ctrl.Request)
 			return reconcile.Result{RequeueAfter: duration}, nil
 		}
 
+		// Collect the hub template watch metric; If there's an error, just log it
+		hubTempWatches := r.DynamicWatcher.GetWatchCount()
```
mprahl, line 178: Do you think we could put this code in a deferred function instead?

controllers/propagator/policy_controller.go
```diff
@@ -160,6 +174,20 @@ func (r *PolicyReconciler) Reconcile(ctx context.Context, request ctrl.Request)
...
+		log.V(3).Info("Setting hub template watch metric", "value", hubTempWatches)
+
+		gaugeMetric, err := hubTemplateActiveWatchesMetric.GetMetricWith(prometheus.Labels{
```
mprahl, line 181: `GetWatchCount` is for *all* watches and not just the watches for that particular policy.

I do like your idea of being able to distinguish which policy has a lot of watches. Perhaps we could add a method to the dynamic watcher library to accept an `ObjectIdentifier` to get the watch count?

dhaiducek: The latest changes I have locally use Gauge instead of GaugeVec. I think I need to focus time on my other issues at this point, especially since the cleanup of GaugeVec becomes a little sticky. 🙂

controllers/propagator/policy_controller.go
dhaiducek: I put it here so that it would be more visible alongside the other metric setting. I figured that way if we wanted to update it in the future, it'd all be in a single file and a single function.

controllers/propagator/policy_controller.go
dhaiducek: Ah, that's a good idea. I'll update it.

controllers/propagator/policy_controller.go
dhaiducek: ...but using `defer` covers the cleanup case also, right?

test/e2e/case14_templates_metrics_test.go
```diff
@@ -0,0 +1,88 @@
...
+	"open-cluster-management.io/governance-policy-propagator/test/utils"
+)
+
+var _ = Describe("Test policy template metrics", func() {
+	const (
+		policyName = "case9-test-policy"
+		policyYaml = "../resources/case9_templates/case9-test-policy.yaml"
+	)
+
+	Describe("Create policy, placement and referenced resource in ns:"+testNamespace, func() {
```
mprahl, line 22: Could you please add `Ordered` to the arguments?

controllers/propagator/policy_controller.go
mprahl: Yup since we would just want it to be `0` when there are no more watches and not delete the metric. If we had a watch count per policy, then we would delete the metric when the policy gets deleted.

test/e2e/case14_templates_metrics_test.go
mprahl: It's just a good practice in my opinion and it does allow for `BeforeAll` and similar functions.

test/e2e/case14_templates_metrics_test.go
```diff
@@ -0,0 +1,88 @@
...
+		})
+
+		It("should correctly report root policy hub template watches when propagated", func() {
+			By("Checking metric endpoint for root policy hub template watches")
+			Eventually(func() interface{} {
+				return utils.GetMetrics("hub_templates_active_watches", "\"[0-9]\"")
+			}, defaultTimeoutSeconds, 1).Should(Equal([]string{"2"}))
+		})
+
+		It("should clean up", func() {
```
mprahl, line 73: Could you please also add this to an `AfterAll` so in case the test fails, it doesn't leave test data on the cluster?

Sorry I missed this before.

dhaiducek: In this case, I think I'd prefer to leave this as part of the test since there's a check to make sure the metric went back down to zero, but I'll add an additional AfterAll() to make sure things are cleaned up on failure.

mprahl (approved):

PR #62, opened 2022-10-28
Fix some issues in replication code

JustinKuli:
The new-ish functions are now tested (and fixed for problems the tests and reviewers found), and better documented. Hopefully some of the "fake" functions in the new tests can find some more use in the future.


dhaiducek (approved):
LGTM!

Seems like we need a unit test utils file for those shiny new fake functions to live in, but that's probably a musing for another time.

PR #64, opened 2022-10-31
Aggregate Pending status from replicated policies

JustinKuli:
When a replicated policy is in the Pending state, and all others for that root policy are Compliant, then the root policy should be Pending.

This change extracts some other aggregation logic in a way that should be more testable, and adds unit tests for some previous existing behavior.

Refs:
 - https://github.com/stolostron/backlog/issues/26182

### DIFF
--- a/controllers/propagator/aggregation.go
+++ b/controllers/propagator/aggregation.go
@@ -1,0 +1,126 @@
+package propagator
+
+import (
+	"context"
+	"sort"
+
+	retry "github.com/avast/retry-go/v3"
+	appsv1 "open-cluster-management.io/multicloud-operators-subscription/pkg/apis/apps/placementrule/v1"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	policiesv1 "open-cluster-management.io/governance-policy-propagator/api/v1"
+	"open-cluster-management.io/governance-policy-propagator/controllers/common"
+)
+
+// calculatePerClusterStatus lists up all policies replicated from the input policy, and stores
+// their compliance states in the result list. Additionally, clusters in the failedClusters input
+// will be marked as NonCompliant in the result. The result is sorted by cluster name. An error
+// will be returned if lookup of the replicated policies fails, and the retries also fail.
+func (r *PolicyReconciler) calculatePerClusterStatus(
+	instance *policiesv1.Policy, failedClusters decisionSet,
+) ([]*policiesv1.CompliancePerClusterStatus, error) {
+	if instance.Spec.Disabled {
+		return nil, nil
+	}
+
+	replicatedPlcList := &policiesv1.PolicyList{}
+
+	log.V(1).Info("Getting the replicated policies")
+
+	err := retry.Do(
+		func() error {
+			return r.List(
+				context.TODO(),
+				replicatedPlcList,
+				client.MatchingLabels(LabelsForRootPolicy(instance)),
+			)
+		},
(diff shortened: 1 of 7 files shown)


controllers/propagator/aggregation.go
```diff
@@ -0,0 +1,126 @@
...
+	"sort"
+
+	retry "github.com/avast/retry-go/v3"
+	appsv1 "open-cluster-management.io/multicloud-operators-subscription/pkg/apis/apps/placementrule/v1"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	policiesv1 "open-cluster-management.io/governance-policy-propagator/api/v1"
+	"open-cluster-management.io/governance-policy-propagator/controllers/common"
+)
+
```
JustinKuli, line 14: I extracted `calculatePerClusterStatus` as a separate function with the intention of unit testing it, but I couldn't get the fakeclient to work with the `List` call that this makes. Using `envtest` would probably work, but would take more setup than I wanted to do here.

controllers/propagator/aggregation.go
```diff
@@ -0,0 +1,126 @@
...
+		)
+
+		status = append(status, &policiesv1.CompliancePerClusterStatus{
+			ComplianceState:  policiesv1.NonCompliant,
+			ClusterName:      clusterDecision.ClusterName,
+			ClusterNamespace: clusterDecision.ClusterNamespace,
+		})
+	}
+
+	sort.Slice(status, func(i, j int) bool {
```
ChunxiAlexLuo, line 84: Is there any need to do sorting here? I know it would be looks nice but the overhead seems unnecessary for the usage.

controllers/propagator/aggregation.go
```diff
@@ -0,0 +1,126 @@
...
+		return status[i].ClusterName < status[j].ClusterName
+	})
+
+	return status, nil
+}
+
+// calculateRootCompliance uses the input per-cluster statuses to determine what a root policy's
+// ComplianceState should be. General precedence is: NonCompliant > Pending > Unknown > Compliant.
```
ChunxiAlexLuo, line 92: calculateRootCompliance looks good to me.

test/e2e/case2_aggregation_test.go
```diff
@@ -15,13 +15,13 @@ import (
 	"open-cluster-management.io/governance-policy-propagator/test/utils"
 )
 
-const (
-	case2PolicyName string = "case2-test-policy"
-	case2PolicyYaml string = "../resources/case2_aggregation/case2-test-policy.yaml"
-)
-
 var _ = Describe("Test policy status aggregation", func() {
```
ChunxiAlexLuo, line 18: Have you tested the unknown case here

controllers/propagator/aggregation.go
JustinKuli: I think it's necessary because the list is copied into the status after this. If the order was inconsistent, the status might be updated when there aren't really any changes.

test/e2e/case2_aggregation_test.go
JustinKuli: Not here, but it is tested in the unit-style test `TestCalculateRootCompliance`

controllers/propagator/aggregation.go
ChunxiAlexLuo: > I think it's necessary because the list is copied into the status after this. If the order was inconsistent, the status might be updated when there aren't really any changes.

Got it.

ChunxiAlexLuo (approved):
/lgtm

PR #67, opened 2022-11-08
update violationMessage field name

ChunxiAlexLuo:


ChunxiAlexLuo:
Have a concern about using the nested loop and `ansibleJob` tag so this might be a better approach.

mprahl (approved):

PR #70, opened 2022-11-18
Only pass violation context when non compliant cluster exists

ChunxiAlexLuo:

Still pass policy info when all clusters are compliant, just skip violation context

### DIFF
--- a/controllers/automation/policyautomation_controller.go
+++ b/controllers/automation/policyautomation_controller.go
@@ -7,6 +7,7 @@ import (
 	"context"
 	"fmt"
 	"net"
+	"strconv"
 	"strings"
 	"time"
 
@@ -170,6 +171,11 @@ func (r *PolicyAutomationReconciler) getViolationContext(
 
 	violationContext.PolicySets = policySets
 
+	// skip policy_violation_context if all clusters are compliant
+	if len(targetList) == 0 {
+		return violationContext, nil
+	}
+
 	replicatedPlcList := &policyv1.PolicyList{}
 
 	err := r.List(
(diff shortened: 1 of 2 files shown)


mprahl:
@ChunxiAlexLuo sorry for the misunderstanding. What I would like is to always set the policy violation context but you would modify `getViolationContext` to return here when `targetList` has a length of 0:
https://github.com/open-cluster-management-io/governance-policy-propagator/blob/5456b9f3e954c24b32af4fe391e095400bc48600/controllers/automation/policyautomation_controller.go#L173

ChunxiAlexLuo:
> @ChunxiAlexLuo sorry for the misunderstanding. What I would like is to always set the policy violation context but you would modify `getViolationContext` to return here when `targetList` has a length of 0:
> 
> https://github.com/open-cluster-management-io/governance-policy-propagator/blob/5456b9f3e954c24b32af4fe391e095400bc48600/controllers/automation/policyautomation_controller.go#L173

Updated

controllers/automation/policyautomation_controller.go
```diff
@@ -170,68 +170,70 @@ func (r *PolicyAutomationReconciler) getViolationContext(
 
 	violationContext.PolicySets = policySets
 
-	replicatedPlcList := &policyv1.PolicyList{}
+	if len(targetList) > 0 {
```
mprahl, line 173: Instead of this if statement, you could do something like this before it:

```go
if len(targetList) == 0 {
	return violationContext, nil
}
```

controllers/automation/policyautomation_controller.go
```diff
@@ -297,13 +299,9 @@ func (r *PolicyAutomationReconciler) Reconcile(
 	}
 
 	if policyAutomation.Annotations["policy.open-cluster-management.io/rerun"] == "true" {
-		violationContext := policyv1beta1.ViolationContext{}
-
 		targetList := common.FindNonCompliantClustersForPolicy(policy)
 		if len(targetList) > 0 {
```
mprahl, line 303: Could you please remove this if/else and just do this?
```go
log.Info("Creating an Ansible job", "mode", "manual", "targetClusters", strconv.Itoa(len(targetList)))
```

mprahl (dismissed):

ChunxiAlexLuo:
The GitHub Action flow hasn't been triggered previously.

gparvin (approved):
Thanks for this fix!

PR #71, opened 2022-11-21
enable pending status for policy set

willkutler:

Allows the status of a `PolicySet` to be `pending`. The compliance state for the policy set follows the same rules as parent policies - if a policy set contains pending policies, the status will be pending unless there are noncompliant policies as well, in which case the policy set will be noncompliant.

The PR also changes the wording of the status message that shows up in a policy set when some policies in the set do not have a status. Previously it referred to those policies as pending, which could be easily confused with policies with the pending compliance state. In order to avoid this confusion, this PR changes the wording from pending policies to policies that are awaiting status.

Needed for https://issues.redhat.com/browse/ACM-2101 and https://issues.redhat.com/browse/ACM-2165

### DIFF
--- a/controllers/policyset/policyset_controller.go
+++ b/controllers/policyset/policyset_controller.go
@@ -105,9 +105,10 @@ func (r *PolicySetReconciler) processPolicySet(ctx context.Context, plcSet *poli
 	// compile results and compliance state from policy statuses
 	compliancesFound := []string{}
 	deletedPolicies := []string{}
-	pendingPolicies := []string{}
+	unknownPolicies := []string{}
 	disabledPolicies := []string{}
-	aggregatedCompliance := "Compliant"
+	pendingPolicies := []string{}
+	aggregatedCompliance := policyv1.Compliant
 	placementsByBinding := map[string]policyv1beta1.PolicySetStatusPlacement{}
 
 	// if there are no policies in the policyset, status should be empty
@@ -203,11 +204,18 @@ func (r *PolicySetReconciler) processPolicySet(ctx context.Context, plcSet *poli
 			// aggregate compliance state
 			plcComplianceState := complianceInRelevantClusters(childPlc.Status.Status, clusters)
 			if plcComplianceState == "" {
-				pendingPolicies = append(pendingPolicies, string(childPlcName))
+				unknownPolicies = append(unknownPolicies, string(childPlcName))
 			} else {
-				compliancesFound = append(compliancesFound, string(childPlcName))
-				if plcComplianceState == "NonCompliant" {
-					aggregatedCompliance = "NonCompliant"
+				if plcComplianceState == policyv1.Pending {
+					pendingPolicies = append(pendingPolicies, string(childPlcName))
+					if aggregatedCompliance != policyv1.NonCompliant {
+						aggregatedCompliance = policyv1.Pending
+					}
+				} else {
+					compliancesFound = append(compliancesFound, string(childPlcName))
+					if plcComplianceState == policyv1.NonCompliant {
+						aggregatedCompliance = policyv1.NonCompliant
+					}
 				}
 			}
 		}
@@ -220,10 +228,10 @@ func (r *PolicySetReconciler) processPolicySet(ctx context.Context, plcSet *poli
 
 	builtStatus := policyv1beta1.PolicySetStatus{
 		Placement:     generatedPlacements,
-		StatusMessage: getStatusMessage(disabledPolicies, pendingPolicies, deletedPolicies),
+		StatusMessage: getStatusMessage(disabledPolicies, unknownPolicies, deletedPolicies, pendingPolicies),
 	}
-	if showCompliance(compliancesFound, pendingPolicies) {
-		builtStatus.Compliant = aggregatedCompliance
+	if showCompliance(compliancesFound, unknownPolicies, pendingPolicies) {
+		builtStatus.Compliant = string(aggregatedCompliance)
 	}
 
 	if !equality.Semantic.DeepEqual(plcSet.Status, builtStatus) {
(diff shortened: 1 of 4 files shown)


controllers/policyset/policyset_controller.go
```diff
@@ -337,6 +339,11 @@ func complianceInRelevantClusters(status []*policyv1.CompliancePerClusterStatus,
 			if string(status[i].ComplianceState) == "NonCompliant" {
 				compliance = "NonCompliant"
 				complianceFound = true
+			} else if string(status[i].ComplianceState) == "Pending" {
```
mprahl, line 342: Optional: Rather than convert to a string, I think you could use the `Pending` constant from the local `api/v1` package.

test/resources/case11_policyset_controller/case11-statuscheck-8.yaml
```diff
@@ -0,0 +1,13 @@
...
+  name: case11-test-policyset
+spec:
+  policies:
+  - case11-test-policy
+status:
+  compliant: Pending
+  placement:
+  - placementBinding: case11-test-policyset-pb
+    placementRule: case11-test-policyset-plr
+  statusMessage: All policies are reporting status
```
mprahl, line 13: Optional: This might be a bit misleading since there are policies that are pending.

mprahl (approved):
/hold in case you want to address some optional comments

test/resources/case11_policyset_controller/case11-statuscheck-8.yaml
willkutler: I wasn't sure whether to change this since pending is technically reporting status. Do you think it would be better to list the pending policies in the status message like we do for disabled policies? @mprahl @JustinKuli

JustinKuli (approved):

controllers/policyset/policyset_controller.go
```diff
@@ -234,44 +241,56 @@ func (r *PolicySetReconciler) processPolicySet(ctx context.Context, plcSet *poli
...
+	if len(pendingPolicies) > 0 {
+		allReporting = false
+		statusMessage = fmt.Sprintf(statusMessage+separator+
+			"Policies awaiting pending dependencies: %s", strings.Join(pendingPolicies, ", "))
+		separator = "; "
+	}
 
 	if len(disabledPolicies) > 0 {
+		allReporting = false
 		statusMessage = fmt.Sprintf("Disabled policies: %s", strings.Join(disabledPolicies, ", "))
```
mprahl, line 264: Did you mean to append this to `statusMessage`?

controllers/policyset/policyset_controller.go
```diff
@@ -234,44 +241,56 @@ func (r *PolicySetReconciler) processPolicySet(ctx context.Context, plcSet *poli
...
+	pendingPolicies []string,
+) string {
+	statusMessage := ""
 	separator := ""
+	allReporting := true
+
+	if len(pendingPolicies) > 0 {
+		allReporting = false
+		statusMessage = fmt.Sprintf(statusMessage+separator+
+			"Policies awaiting pending dependencies: %s", strings.Join(pendingPolicies, ", "))
```
mprahl, line 258: This is easier to read in my opinion
```suggestion
		statusMessage += seperator + fmt.Sprintf("Policies awaiting pending dependencies: %s", strings.Join(pendingPolicies, ", "))
```

controllers/policyset/policyset_controller.go
```diff
@@ -234,44 +241,56 @@ func (r *PolicySetReconciler) processPolicySet(ctx context.Context, plcSet *poli
...
 	}
 
-	if len(pendingPolicies) > 0 {
-		if separator == "" {
-			statusMessage = ""
-		}
-
+	if len(unknownPolicies) > 0 {
+		allReporting = false
 		statusMessage = fmt.Sprintf(statusMessage+separator+
```
mprahl, line 270: Same readability comment as for `pendingPolicies`.

controllers/policyset/policyset_controller.go
```diff
@@ -234,44 +241,56 @@ func (r *PolicySetReconciler) processPolicySet(ctx context.Context, plcSet *poli
...
 		separator = "; "
 	}
 
 	if len(deletedPolicies) > 0 {
-		if separator == "" {
-			statusMessage = ""
-		}
-
+		allReporting = false
 		statusMessage = fmt.Sprintf(statusMessage+separator+"Deleted policies: %s", strings.Join(deletedPolicies, ", "))
```
mprahl, line 277: Same readability comment as for `pendingPolicies`.

controllers/policyset/policyset_controller.go
```diff
@@ -330,21 +349,26 @@ func (r *PolicySetReconciler) SetupWithManager(mgr ctrl.Manager) error {
...
+	compliance := policyv1.Compliant
 
 	for i := range status {
 		if clusterInList(relevantClusters, status[i].ClusterName) {
-			if string(status[i].ComplianceState) == "NonCompliant" {
-				compliance = "NonCompliant"
+			if string(status[i].ComplianceState) == string(policyv1.NonCompliant) {
+				compliance = policyv1.NonCompliant
 				complianceFound = true
+			} else if string(status[i].ComplianceState) == string(policyv1.Pending) {
```
mprahl, line 359: Same as above

controllers/policyset/policyset_controller.go
```diff
@@ -330,21 +349,26 @@ func (r *PolicySetReconciler) SetupWithManager(mgr ctrl.Manager) error {
 // Helper function to filter out compliance statuses that are not in scope
 func complianceInRelevantClusters(status []*policyv1.CompliancePerClusterStatus, relevantClusters []string) string {
```
mprahl, line 350: Optional: Ideally, this would return `ComplianceState` instead of a string. In my opinion, we should keep it typed until we really need to convert it to a string.

mprahl:
@willkutler could you please also include the issue URL in the commit message? This allows someone to go directly to the issue when looking at the commit history rather than figure out what PR it came from.

Additionally, could you please add some of the description in your PR to the commit message as well? This makes it easier to know why a change was made when looking at the commit history.

Lastly, could you please capitalize the first letter in your commit message to conform to our contributing guidelines?

controllers/policyset/policyset_controller.go
willkutler: no, this is just a bool to determine whether to use the "All policies are reporting status" message or not

willkutler:
@mprahl added some changes based on your comments. let me know if it needs anything else!

controllers/policyset/policyset_controller.go
```diff
@@ -203,11 +204,17 @@ func (r *PolicySetReconciler) processPolicySet(ctx context.Context, plcSet *poli
...
 			if plcComplianceState == "" {
-				pendingPolicies = append(pendingPolicies, string(childPlcName))
+				unknownPolicies = append(unknownPolicies, string(childPlcName))
 			} else {
-				compliancesFound = append(compliancesFound, string(childPlcName))
-				if plcComplianceState == "NonCompliant" {
-					aggregatedCompliance = "NonCompliant"
+				if plcComplianceState == policyv1.Pending &&
+					aggregatedCompliance != policyv1.NonCompliant {
+					pendingPolicies = append(pendingPolicies, string(childPlcName))
```
mprahl, line 211: What is the reason for not appending to `pendingPolicies` when `aggregatedCompliance == policyv1.NonCompliant`? I think we only want to not set `aggregatedCompliance` in this case.

controllers/policyset/policyset_controller.go
```diff
@@ -203,11 +204,17 @@ func (r *PolicySetReconciler) processPolicySet(ctx context.Context, plcSet *poli
...
+					aggregatedCompliance = policyv1.Pending
+				} else {
+					compliancesFound = append(compliancesFound, string(childPlcName))
+					if plcComplianceState == policyv1.NonCompliant {
+						aggregatedCompliance = policyv1.NonCompliant
+					}
```
mprahl, line 217: I think you could simplify this by having this outside of the if/else and change it to:
```go
if aggregatedCompliance != policyv1.NonCompliant && aggregatedCompliance != policyv1.Pending {
aggregatedCompliance = plcComplianceState
}
```

controllers/policyset/policyset_controller.go
```diff
@@ -234,44 +241,56 @@ func (r *PolicySetReconciler) processPolicySet(ctx context.Context, plcSet *poli
...
+) string {
+	statusMessage := ""
 	separator := ""
+	allReporting := true
 
-	if len(disabledPolicies) > 0 {
-		statusMessage = fmt.Sprintf("Disabled policies: %s", strings.Join(disabledPolicies, ", "))
+	if len(pendingPolicies) > 0 {
+		allReporting = false
+		statusMessage += fmt.Sprintf(separator+"Policies awaiting pending dependencies: %s",
```
mprahl, line 257: Optional: It's not needed here since it's the first encountered if statement.
```suggestion
		statusMessage += fmt.Sprintf("Policies awaiting pending dependencies: %s",
```

willkutler:
@mprahl good catch on the compliance logic, thanks! Let me know what you think of the commit I added to fix it - I can do it like https://github.com/open-cluster-management-io/governance-policy-propagator/pull/71#discussion_r1030589991 instead if you'd prefer that, but I think it might be a little clearer to read like this.

mprahl:
> @mprahl good catch on the compliance logic, thanks! Let me know what you think of the commit I added to fix it - I can do it like [#71 (comment)](https://github.com/open-cluster-management-io/governance-policy-propagator/pull/71#discussion_r1030589991) instead if you'd prefer that, but I think it might be a little clearer to read like this.

That looks good to me! Once you squash your commit, I'll go ahead and approve it.

willkutler:
@mprahl is this ok to merge?

willkutler:
realized I forgot to remove the separator from the first status message string - just fixed and squashed

mprahl (approved):

PR #73, opened 2022-11-29
Add CRD validation for namespaces in dependencies

JustinKuli:
It is invalid to set a namespace on dependencies from our known types (ConfigurationPolicy, IamPolicy, CertificatePolicy), because those will be in the clusternamespace on the managed cluster.


dhaiducek (approved):
LGTM!

That's some fancy validation! 🙂

PR #81, opened 2023-01-09
Provide the correct Hub DNS name in the Ansible extra_vars

mprahl:
Using the Kubernetes configuration leads to a local DNS name when running in the cluster. This commit changes it to use the OpenShift dnses custom resource to get the value. Other Kubernetes distributions could have their own implementations added later.

Relates:
https://issues.redhat.com/browse/ACM-2410


gparvin (approved):
Thanks for the fix here!

PR #85, opened 2023-01-13
add test for ocm_handle_root_policy_duration metric

willkutler:
ref:
- https://issues.redhat.com/browse/ACM-2513

### DIFF
--- a/test/e2e/case14_root_policy_metrics_test.go
+++ b/test/e2e/case14_root_policy_metrics_test.go
@@ -5,6 +5,8 @@ package e2e
 
 import (
 	"context"
+	"fmt"
+	"strconv"
 
 	. "github.com/onsi/ginkgo/v2"
 	. "github.com/onsi/gomega"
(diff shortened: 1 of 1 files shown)


dhaiducek:
@willkutler Would you please add the Jira issue URL to the commit message so that it can be tracked in `stolostron`?

willkutler:
> @willkutler Would you please add the Jira issue URL to the commit message so that it can be tracked in `stolostron`?

Done!

dhaiducek (changes requested):
I think this looks good, but since it's being combined with another test, would you consider:

- Renaming the file: `case14_templates_metrics_test.go` --> `case14_root_policy_metrics_test.go`
- Rewording the root `Describe`: `"Test policy template metrics"` --> `"Test root policy metrics"`

I think that'd be clearer now that there are two metrics types in the same test. What do you think?

willkutler:
@dhaiducek looks good to me! I just made those rewording changes and squashed

dhaiducek (approved):
LGTM! Thanks, @willkutler!

PR #86, opened 2023-01-17
Refactor root policy status updates

mprahl:
In the event that a policy status constantly changes on hundreds of managed clusters, the CPU and memory utilization would significantly increase. This was due to status updates in replicated policies always triggering the expensive code of generating all replicated policies and updating them if necessary as well as calculate the root policy status by looking at all the replicated policy statuses.

This new approach adds a new controller that handles just status updates from replicated policies and then updates root policy status based on the current state. The main policy propagator controller still gets called for all other events.

As part of this, some refactoring was done to improve resource utilization and performance. One example is instead of listing all managed clusters to see if there is a managed cluster with the same name as the policy namespace, a single get request is made. In the event of 3000 managed clusters, this significantly reduces memory utilization as a list operation would have created 3000 copies from the controller-runtime cache.

The retries logic around individual API calls was removed because almost all of the failures can't be resolved with retries since they are gets and lists from the controller-runtime cache. Update operations that fail are likely due to the underlying object changing since it was retrieved from the cache, so a retry on the whole reconcile is a better approach. This is beneficial because a failed reconcile will get retried right away now rather than wait for 5 minutes.

These changes in a local scale test reduced CPU and memory utilization significantly. Memory and CPU utilization no longer grow linearly and instead stabilize at low resource utilization.

Relates:
https://issues.redhat.com/browse/ACM-2543
https://issues.redhat.com/browse/ACM-2774

Signed-off-by: mprahl <mprahl@users.noreply.github.com>


controllers/policystatus/policy_status_controller.go
```diff
@@ -0,0 +1,140 @@
...
+	Scheme          *runtime.Scheme
+}
+
+// Reconcile handles a replicated policy status update and then updates the root policy status. It does not handle
+// creations or deletions of replicated policies. Those are handled by the main policy controller. This Reconcile
+// may have a lot retries if several clusters are updating the same policy at the same time due to a race condition
+// where the root policy status update hasn't been updated in the cache.
+func (r *PolicyStatusReconciler) Reconcile(ctx context.Context, request ctrl.Request) (ctrl.Result, error) {
+	log := log.WithValues("Request.Namespace", request.Namespace, "Request.Name", request.Name)
+	log.Info("Reconciling the root policy status")
```
JustinKuli, line 60: I'm thinking this should be `V(1)` - since it comes from every replicated policy, it could be very noisy in a busy cluster. Right?

controllers/propagator/aggregation.go
```diff
@@ -4,65 +4,46 @@ import (
...
-		key := appsv1.PlacementDecision{
-			ClusterName:      rPlc.GetLabels()[common.ClusterNameLabel],
-			ClusterNamespace: rPlc.GetLabels()[common.ClusterNamespaceLabel],
-		}
+	status := make([]*policiesv1.CompliancePerClusterStatus, 0, len(allDecisions))
 
-		if failed := failedClusters[key]; failed {
+	// Update the status based on the processed decisions
+	for decision := range allDecisions {
+		if failed := failedClusters[decision]; failed {
```
JustinKuli, line 27: Nit
```suggestion
		if failedClusters[decision] {
```

controllers/propagator/aggregation.go
```diff
@@ -4,65 +4,46 @@ import (
...
 			// for now. This will be handled later.
 			continue
 		}
 
+		rPlc := &policiesv1.Policy{}
+		key := types.NamespacedName{
+			Namespace: decision.ClusterNamespace, Name: instance.Namespace + "." + instance.Name,
+		}
+
+		err := r.Get(context.TODO(), key, rPlc)
```
JustinKuli, line 38: So the `List` was replaced by multiple `Gets` - one for each successfully replicated policy? It's not clear to me that this will perform better. If there is an overhead to each "request" to the cache, then it feels like this would not be better.

controllers/propagator/policyPredicate.go
```diff
@@ -0,0 +1,29 @@
...
+		UpdateFunc: func(e event.UpdateEvent) bool {
+			// nolint: forcetypeassert
+			oldPolicy := e.ObjectOld.(*policiesv1.Policy)
+			// nolint: forcetypeassert
+			updatedPolicy := e.ObjectNew.(*policiesv1.Policy)
+
+			// Ignore pure status updates since those are handled by a separate controller
+			return !equality.Semantic.DeepEqual(oldPolicy.ObjectMeta.Labels, updatedPolicy.ObjectMeta.Labels) ||
+				!equality.Semantic.DeepEqual(oldPolicy.ObjectMeta.Annotations, updatedPolicy.ObjectMeta.Annotations) ||
+				!equality.Semantic.DeepEqual(oldPolicy.Spec, updatedPolicy.Spec)
```
JustinKuli, line 26: Can we check generation instead? This DeepEqual could be expensive.

test/e2e/case3_mutation_recovery_test.go
```diff
@@ -142,27 +142,6 @@ var _ = Describe("Test unexpected policy mutation", func() {
...
-		By("Getting root policy again")
-		yamlPlc := utils.ParseYaml("../resources/case3_mutation_recovery/managed-both-status-compliant.yaml")
-		Eventually(func() interface{} {
-			rootPlc = utils.GetWithTimeout(
-				clientHubDynamic, gvrPolicy, case3PolicyName, testNamespace, true, defaultTimeoutSeconds,
-			)
-
-			return rootPlc.Object["status"]
-		}, defaultTimeoutSeconds, 1).Should(utils.SemanticEqual(yamlPlc.Object["status"]))
-	})
```
JustinKuli, line 165: Is it possible to come up with a way that the root policy status would be re-calculated if it was deleted like this? I don't really like the idea of removing this behavior, but I do understand the trade-off

controllers/common/lock.go
```diff
@@ -0,0 +1,44 @@
...
+
+import (
+	"sync"
+
+	"k8s.io/apimachinery/pkg/types"
+)
+
+// PoliciesLock allows locking at the root policy level so that multiple controllers/goroutines don't race with each
+// other and cause unnecessary reconcile retries.
+type PoliciesLock struct {
```
JustinKuli, line 13: I want to think more about this lock in general. I might need some time to stew with it, or I might need to ask questions directly (so be prepared ;) )

JustinKuli (changes requested):

controllers/policymetrics/policymetrics_controller.go
```diff
@@ -52,18 +51,16 @@ func (r *MetricReconciler) Reconcile(ctx context.Context, request ctrl.Request)
...
-		log.Error(err, "Failed to list clusters, going to requeue the request")
+		log.Error(err, "Failed to determine if the policy is a replicated policy")
 
 		return reconcile.Result{}, err
 	}
 
 	var promLabels map[string]string
 
-	if common.IsInClusterNamespace(request.Namespace, clusterList.Items) {
+	if inClusterNs {
```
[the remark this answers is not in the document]
mprahl: Good point. At least it'll be "deleted" as part of a restart.

I think we need a separate issue to refactor this metrics controller in general.

controllers/common/lock.go
```diff
@@ -0,0 +1,44 @@
...
+
+	if _, ok := p.policies[policy]; !ok {
+		p.policies[policy] = &sync.Mutex{}
+	}
+
+	lock := p.policies[policy]
+
+	p.internalLock.Unlock()
+
+	lock.Lock()
```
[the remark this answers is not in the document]
mprahl: `lock.Lock()` is blocking if the lock is already held for that root policy, so the internal lock is released after reading from the map is done to avoid having an artificial lock on all root policies.

controllers/policystatus/policy_status_controller.go
```diff
@@ -0,0 +1,140 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/controller"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+	"sigs.k8s.io/controller-runtime/pkg/source"
+
+	policiesv1 "open-cluster-management.io/governance-policy-propagator/api/v1"
+	"open-cluster-management.io/governance-policy-propagator/controllers/common"
+	"open-cluster-management.io/governance-policy-propagator/controllers/propagator"
+)
+
+const ControllerName string = "policy-status"
```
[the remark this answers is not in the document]
mprahl: I'll change it to `root-policy-status` unless you have another idea.  :smile:

controllers/policystatus/policy_status_controller.go
mprahl: Yeah, you're right. I'll change it.

controllers/propagator/aggregation.go
mprahl: I replaced the `List` call not for performance reasons but due to race conditions. I found that using `allDecisions` as the authoritative list of replicated policies was more reliable since the tests would flake with missing replicated policies in the status. I believe this was due to the cache not being updated with all the new replicated policies in the time of the `List` call. This didn't show up previously because this method was called again as part of the Reconcile of the root policy status update and then it would get corrected. I since relaxed the predicates, so perhaps this is no longer an issue.

It does use less memory to get one policy at a time from the cache as long as the garbage collection happens when the single replicated policy goes out of scope.

Do you want me to try changing it back?

controllers/propagator/policyPredicate.go
mprahl: Great idea! Thanks.

test/e2e/case3_mutation_recovery_test.go
mprahl: I could change the predicate to reconcile when the status is empty but that only partially fixes the problem and could be wasteful.

What do you think?

controllers/common/lock.go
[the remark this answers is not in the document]
JustinKuli: I think that makes sense. But now I'm wondering if https://pkg.go.dev/sync#Map can be used here?

test/e2e/case3_mutation_recovery_test.go
JustinKuli: How wasteful would it be? Your suggestion seems like a quick and easy way to get this reconcile to happen. What's only partially fixed?

JustinKuli:
Let's say I make a new policy that will be replicated to 1000 managed clusters. As the compliance on the managed clusters is calculated and trickles in, the status on the root policy needs to be updated. (I expect this statement will always be true, correct me if I'm wrong.)

Previously: the first replicated policy with a status update would trigger a reconcile on the root policy. Any replicated policy updates after that would be effectively batched into that one reconcile. After that reconcile starts, any "new" replicated policy updates would be batched into **just 1 more root policy reconcile**. If all 1000 clusters reported their status within 60s, and the reconcile takes 1 second each time, there would be a maximum of something like 61 reconciles (and 61 associated update requests to the k8s apiserver).

With this change (as I understand it): **every** replicated policy status update results in a policystatus reconcile. Each of those reconciles is admittedly smaller than the previous propagator reconcile, but each one has a separate update request to the k8s apiserver, so we'd need 1000 requests to the k8s apiserver. And each one uses the root policy's lock, with potentially many goroutines queueing up to obtain the lock.

I like the idea of a smaller reconcile to deal with replicated policy status updates. But I don't like the increased number of apiserver requests here. IMO, the best would be if we could have a smaller reconcile triggered by replicated policy updates, which handles *all* the replicated policies at once, and has some limit to how often it can run (eg max every 15s while updates are trickling in).

mprahl:
> Let's say I make a new policy that will be replicated to 1000 managed clusters. As the compliance on the managed clusters is calculated and trickles in, the status on the root policy needs to be updated. (I expect this statement will always be true, correct me if I'm wrong.)
> 
> Previously: the first replicated policy with a status update would trigger a reconcile on the root policy. Any replicated policy updates after that would be effectively batched into that one reconcile. After that reconcile starts, any "new" replicated policy updates would be batched into **just 1 more root policy reconcile**. If all 1000 clusters reported their status within 60s, and the reconcile takes 1 second each time, there would be a maximum of something like 61 reconciles (and 61 associated update requests to the k8s apiserver).
> 
> With this change (as I understand it): **every** replicated policy status update results in a policystatus reconcile. Each of those reconciles is admittedly smaller than the previous propagator reconcile, but each one has a separate update request to the k8s apiserver, so we'd need 1000 requests to the k8s apiserver. And each one uses the root policy's lock, with potentially many goroutines queueing up to obtain the lock.
> 
> I like the idea of a smaller reconcile to deal with replicated policy status updates. But I don't like the increased number of apiserver requests here. IMO, the best would be if we could have a smaller reconcile triggered by replicated policy updates that doesn't generate all the replicated policies, but does handle _all_ the replicated policies at once, and has some limit to how often it can run (eg max every 15s while updates are trickling in).



> Let's say I make a new policy that will be replicated to 1000 managed clusters. As the compliance on the managed clusters is calculated and trickles in, the status on the root policy needs to be updated. (I expect this statement will always be true, correct me if I'm wrong.)
> 
> Previously: the first replicated policy with a status update would trigger a reconcile on the root policy. Any replicated policy updates after that would be effectively batched into that one reconcile. After that reconcile starts, any "new" replicated policy updates would be batched into **just 1 more root policy reconcile**. If all 1000 clusters reported their status within 60s, and the reconcile takes 1 second each time, there would be a maximum of something like 61 reconciles (and 61 associated update requests to the k8s apiserver).
> 
> With this change (as I understand it): **every** replicated policy status update results in a policystatus reconcile. Each of those reconciles is admittedly smaller than the previous propagator reconcile, but each one has a separate update request to the k8s apiserver, so we'd need 1000 requests to the k8s apiserver. And each one uses the root policy's lock, with potentially many goroutines queueing up to obtain the lock.
> 
> I like the idea of a smaller reconcile to deal with replicated policy status updates. But I don't like the increased number of apiserver requests here. IMO, the best would be if we could have a smaller reconcile triggered by replicated policy updates that doesn't generate all the replicated policies, but does handle _all_ the replicated policies at once, and has some limit to how often it can run (eg max every 15s while updates are trickling in).

If we want to optimize the API server impact, the new controller could reconcile on the root policy instead and be responsible for comparing *all* replicated policies. It's less efficient in the controller by a lot, but it does help the load on the Kubernetes API server since it inherently batches update requests.

What do you think?

JustinKuli:
I have some preliminary information from the performance testing I'm working on. The raw data can be found in [my repo's artifacts on the action runs](https://github.com/JustinKuli/governance-policy-propagator-1/actions/workflows/kind-performance.yml) - you'll need to use `go tool pprof` to read the profiles, but there is also some metric collection which is human-readable and useful. The meat of the tests are [here](https://github.com/JustinKuli/governance-policy-propagator-1/blob/performance-baseline/test/performance/scale_test.go) but the tl;dr is that they put policies on 300, 1000, and 2000 fake managed clusters with some threads changing some of their compliances, and collect profiling information.

It's probably not a very robust performance platform, but it does demonstrate a substantial cpu savings with the changes in this PR. The main branch and smaller changes I'd quickly been playing with kept CPU usage around `105%` (the machine had 2 cores, so I'm assuming the max would be 200%), while the tests based on this PR hovered around `15%`.

But there are some specific metrics with this implementation that worry me. Here's what I thought were the interesting bits, gathered 2+ minutes after a policy was replicated to 2000 clusters (with some threads sending them updates in the meantime):
```
# HELP controller_runtime_reconcile_time_seconds Length of time per reconciliation per controller
# TYPE controller_runtime_reconcile_time_seconds histogram
controller_runtime_reconcile_time_seconds_bucket{controller="policy-metrics",le="0.05"} 28978
controller_runtime_reconcile_time_seconds_bucket{controller="policy-metrics",le="+Inf"} 28979
controller_runtime_reconcile_time_seconds_sum{controller="policy-metrics"} 3.545132122000024

controller_runtime_reconcile_time_seconds_bucket{controller="policy-propagator",le="0.3"} 10
controller_runtime_reconcile_time_seconds_bucket{controller="policy-propagator",le="+Inf"} 13
controller_runtime_reconcile_time_seconds_sum{controller="policy-propagator"} 19.775843079

controller_runtime_reconcile_time_seconds_bucket{controller="policy-status",le="0.05"} 1814
controller_runtime_reconcile_time_seconds_bucket{controller="policy-status",le="0.15"} 3611
controller_runtime_reconcile_time_seconds_bucket{controller="policy-status",le="0.25"} 5307
controller_runtime_reconcile_time_seconds_bucket{controller="policy-status",le="0.5"} 7156
controller_runtime_reconcile_time_seconds_bucket{controller="policy-status",le="1.75"} 7623
controller_runtime_reconcile_time_seconds_bucket{controller="policy-status",le="+Inf"} 7624
controller_runtime_reconcile_time_seconds_sum{controller="policy-status"} 1499.6511942349987

# HELP controller_runtime_reconcile_total Total number of reconciliations per controller
# TYPE controller_runtime_reconcile_total counter
controller_runtime_reconcile_total{controller="policy-status",result="error"} 2648
controller_runtime_reconcile_total{controller="policy-status",result="success"} 4976

# HELP workqueue_adds_total Total number of adds handled by workqueue
# TYPE workqueue_adds_total counter
workqueue_adds_total{name="policy-metrics"} 28979
workqueue_adds_total{name="policy-propagator"} 13
workqueue_adds_total{name="policy-status"} 9424

# HELP workqueue_depth Current depth of workqueue
# TYPE workqueue_depth gauge
workqueue_depth{name="policy-metrics"} 0
workqueue_depth{name="policy-propagator"} 0
workqueue_depth{name="policy-status"} 1795

# HELP workqueue_queue_duration_seconds How long in seconds an item stays in workqueue before being requested
# TYPE workqueue_queue_duration_seconds histogram
workqueue_queue_duration_seconds_bucket{name="policy-propagator",le="0.001"} 4
workqueue_queue_duration_seconds_bucket{name="policy-propagator",le="1"} 11
workqueue_queue_duration_seconds_bucket{name="policy-propagator",le="+Inf"} 13
workqueue_queue_duration_seconds_sum{name="policy-propagator"} 19.609075393

workqueue_queue_duration_seconds_bucket{name="policy-status",le="0.001"} 1563
workqueue_queue_duration_seconds_bucket{name="policy-status",le="1"} 1623
workqueue_queue_duration_seconds_bucket{name="policy-status",le="10"} 2179
workqueue_queue_duration_seconds_bucket{name="policy-status",le="+Inf"} 7629
workqueue_queue_duration_seconds_sum{name="policy-status"} 169400.92749583925
```

JustinKuli (approved):
Looking good! An updated performance test showed that main used around `300%` of 4 cores when dealing with 3000 fake managed clusters with a policy that was being updated, but the same test on this PR used only around `46%` of 4 cores. I don't think we need a more robust performance test or statistical analysis to know that this is much better!

The other metrics I checked out mostly looked good - although I should say my tests last night might have been causing more replicated policy updates than I meant. The one thing we should follow up on:
```
controller_runtime_reconcile_errors_total{controller="root-policy-status"} 2589
```

Unfortunately I clobbered the logs from the test run, so I'm not sure what the error was.

/hold

I'll run an updated performance test of the previous commit so we can compare to that more directly (if you're not curious about that, we can un-hold the PR)

JustinKuli:
For the curious: an updated test based on 2b76c76 showed only `26%` cpu usage for dealing with 3000 clusters (compare to `46%` in the commit we merged), but even with a more reasonable rate of replicated policy updates, it seems like it wasn't keeping up with the queue:
```
# HELP workqueue_depth Current depth of workqueue
workqueue_depth{name="policy-status"} 2788

# HELP workqueue_queue_duration_seconds How long in seconds an item stays in workqueue before being requested
workqueue_queue_duration_seconds_bucket{name="policy-status",le="0.1"} 2142
workqueue_queue_duration_seconds_bucket{name="policy-status",le="1"} 2211
workqueue_queue_duration_seconds_bucket{name="policy-status",le="10"} 2818
workqueue_queue_duration_seconds_bucket{name="policy-status",le="+Inf"} 7402
workqueue_queue_duration_seconds_sum{name="policy-status"} 216054.53274635164
```

Compared to the metrics for the commit we merged:
```
# HELP workqueue_depth Current depth of workqueue
workqueue_depth{name="root-policy-status"} 1

# HELP workqueue_queue_duration_seconds How long in seconds an item stays in workqueue before being requested
workqueue_queue_duration_seconds_bucket{name="root-policy-status",le="0.01"} 2799
workqueue_queue_duration_seconds_bucket{name="root-policy-status",le="0.1"} 7626
workqueue_queue_duration_seconds_bucket{name="root-policy-status",le="1"} 8077
workqueue_queue_duration_seconds_bucket{name="root-policy-status",le="10"} 8079
workqueue_queue_duration_seconds_bucket{name="root-policy-status",le="+Inf"} 8079
workqueue_queue_duration_seconds_sum{name="root-policy-status"} 270.45633124500023
```

tl;dr I'm happy with what we merged

mprahl:
> For the curious: an updated test based on [2b76c76](https://github.com/open-cluster-management-io/governance-policy-propagator/commit/2b76c765e3486ef18e5f43cf64f054dfebe2bb2f) showed only `26%` cpu usage for dealing with 3000 clusters (compare to `46%` in the commit we merged), but even with a more reasonable rate of replicated policy updates, it seems like it wasn't keeping up with the queue:
> 
> ```
> # HELP workqueue_depth Current depth of workqueue
> workqueue_depth{name="policy-status"} 2788
> 
> # HELP workqueue_queue_duration_seconds How long in seconds an item stays in workqueue before being requested
> workqueue_queue_duration_seconds_bucket{name="policy-status",le="0.1"} 2142
> workqueue_queue_duration_seconds_bucket{name="policy-status",le="1"} 2211
> workqueue_queue_duration_seconds_bucket{name="policy-status",le="10"} 2818
> workqueue_queue_duration_seconds_bucket{name="policy-status",le="+Inf"} 7402
> workqueue_queue_duration_seconds_sum{name="policy-status"} 216054.53274635164
> ```
> 
> Compared to the metrics for the commit we merged:
> 
> ```
> # HELP workqueue_depth Current depth of workqueue
> workqueue_depth{name="root-policy-status"} 1
> 
> # HELP workqueue_queue_duration_seconds How long in seconds an item stays in workqueue before being requested
> workqueue_queue_duration_seconds_bucket{name="root-policy-status",le="0.01"} 2799
> workqueue_queue_duration_seconds_bucket{name="root-policy-status",le="0.1"} 7626
> workqueue_queue_duration_seconds_bucket{name="root-policy-status",le="1"} 8077
> workqueue_queue_duration_seconds_bucket{name="root-policy-status",le="10"} 8079
> workqueue_queue_duration_seconds_bucket{name="root-policy-status",le="+Inf"} 8079
> workqueue_queue_duration_seconds_sum{name="root-policy-status"} 270.45633124500023
> ```
> 
> tl;dr I'm happy with what we merged

Thanks for the great analysis!

PR #89, opened 2023-01-20
Add concurrency to the policy-metrics controller

mprahl:
In scale tests, the policy-metrics controller has bad latency since there is a reconcile per root or replicated policy status update. This was improved in PR #86, but it needs concurrency to be able to keep up.

Relates:
https://issues.redhat.com/browse/ACM-2543


dhaiducek (approved):
LGTM!

PR #91, opened 2023-01-27
Add the spec.copyPolicyMetadata field

mprahl:
This new field allows a replicated policy to not have labels and annotations that are not managed by the policy framework. This is useful in the case where policies are deployed with ArgoCD and you don't want them to show up in the ArgoCD UI.

Additionally, the argocd.argoproj.io/compare-options annotation is now always set to IgnoreExtraneous on the replicated policies to avoid ArgoCD trying to manage the replicated policies.

Relates:
https://issues.redhat.com/browse/ACM-1690

### DIFF
--- a/api/v1/policy_types.go
+++ b/api/v1/policy_types.go
@@ -69,6 +69,12 @@ type PolicySpec struct {
 	// This provides the ability to enable and disable your policies.
 	Disabled bool `json:"disabled"`
 
+	// If set to true (default), all the policy's labels and annotations will be copied to the replicated policy.
+	// If set to false, only the policy framework specific policy labels and annotations will be copied to the
+	// replicated policy.
+	// +kubebuilder:validation:Optional
+	CopyPolicyMetadata *bool `json:"copyPolicyMetadata,omitempty"`
+
 	// This value (Enforce or Inform) will override the remediationAction on each template
 	RemediationAction RemediationAction `json:"remediationAction,omitempty"`
 
--- a/controllers/propagator/replication.go
+++ b/controllers/propagator/replication.go
@@ -16,6 +16,8 @@ import (
 	"open-cluster-management.io/governance-policy-propagator/controllers/common"
 )
 
+const argoCDCompareOptionsAnnotation = "argocd.argoproj.io/compare-options"
+
 // equivalentReplicatedPolicies compares replicated policies. Returns true if they match.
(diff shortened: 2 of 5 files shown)


api/v1/policy_types.go
```diff
@@ -69,6 +69,12 @@ type PolicySpec struct {
 	// This provides the ability to enable and disable your policies.
 	Disabled bool `json:"disabled"`
 
+	// If set to true (default), all the policy's labels and annotations will be copied to the replicated policy.
+	// If set to false, only the policy framework specific policy labels and annotations will be copied to the
+	// replicated policy.
+	// +kubebuilder:validation:Optional
+	CopyPolicyMetadata *bool `json:"copyPolicyMetadata,omitempty"`
```
mprahl, line 76: Note that this is a pointer to avoid setting a default in the CRD to always cause this field to be shown when unset.

controllers/propagator/replication.go
```diff
@@ -16,6 +16,8 @@ import (
 	"open-cluster-management.io/governance-policy-propagator/controllers/common"
 )
 
+const argoCDCompareOptionsAnnotation = "argocd.argoproj.io/compare-options"
+
 // equivalentReplicatedPolicies compares replicated policies. Returns true if they match.
 func equivalentReplicatedPolicies(plc1 *policiesv1.Policy, plc2 *policiesv1.Policy) bool {
```
dhaiducek, line 22: Does `equivalentReplicatedPolicies()` need to be updated also with this logic so that it doesn't try to keep replicating the policy?

mprahl: This is covered because `buildReplicatedPolicy` is called to get the "desired" replicated policy.

dhaiducek (dismissed):
/hold

LGTM! Not sure whether other reviews are needed--unhold if not! 🙂

mprahl:
> /hold
> 
> LGTM! Not sure whether other reviews are needed--unhold if not! slightly_smiling_face

Thanks for the review!

I'll wait for a second +1 since it is a CRD change.

JustinKuli (dismissed):
LGTM!

mprahl:
@dhaiducek and @JustinKuli , I realized I had a typo in the commit subject, so I had to force push. Could you please approve it again?

dhaiducek (approved):
...and now I have two e-mail threads for this PR 😉

PR #94, opened 2023-02-20
Move `EquivalentReplicatedPolicies` to `common`

dhaiducek:
`governance-policy-framework-addon` was relying on it to make comparisons


dhaiducek:
If preferred, I could migrate this function directly to `governance-policy-framework-addon`. I wasn't 100% sure whether this function was important enough to have a common behavior between the controllers, especially since it's pretty straightforward...

PR #99, opened 2023-03-29
Add support for the copy template functions to the propagator

gparvin:
Adding the `copyConfigMapData` and the `copySecretData` template functions to the policy propagator.

Refs:
 - https://issues.redhat.com/browse/ACM-2611

### DIFF
--- a/test/e2e/case9_templates_test.go
+++ b/test/e2e/case9_templates_test.go
@@ -29,6 +29,9 @@ const (
 	case9EncryptionSecretName         = "policy-encryption-key"
 	case9SecretName                   = "case9-secret"
 	IVAnnotation                      = "policy.open-cluster-management.io/encryption-iv"
+	case9PolicyNameCopy               = "case9-test-policy-copy"
+	case9PolicyYamlCopy               = "../resources/case9_templates/case9-test-policy_copy.yaml"
+	case9PolicyYamlCopiedRepl         = "../resources/case9_templates/case9-test-replpolicy_copied-"
 )
 
 var _ = Describe("Test policy templates", func() {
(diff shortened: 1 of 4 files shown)


test/e2e/case9_templates_test.go
```diff
@@ -282,3 +285,150 @@ var _ = Describe("Test encrypted policy templates", func() {
...
+					clientHubDynamic, gvrSecret, case9EncryptionSecretName, managedCluster,
+					false, defaultTimeoutSeconds,
+				)
+			})
+
+			It("should clean up", func() {
+				utils.Kubectl("delete", "-f", case9PolicyYamlCopy, "-n", testNamespace)
+				opt := metav1.ListOptions{}
+				utils.ListWithTimeout(clientHubDynamic, gvrPolicy, opt, 0, false, defaultTimeoutSeconds)
+			})
```
JustinKuli, line 431: The way you have the test organized (which is otherwise ok), I don't think these can *just* be in an AfterAll, but it would be nice to have an extra `AfterAll` version of them (ignoring not-found errors) to make sure things are cleaned up even on failures.

test/resources/case9_templates/case9-test-policy_copy.yaml
```diff
@@ -0,0 +1,90 @@
...
+    - objectDefinition:
+        apiVersion: policy.open-cluster-management.io/v1
+        kind: ConfigurationPolicy
+        metadata:
+          name: case9-test-configpolicy
+        spec:
+          remediationAction: inform
+          namespaceSelector:
+            exclude: ["kube-*"]
+            include: ["default"]
```
JustinKuli, line 18: Nit: using a namespace selector when the resource has a namespace is unnecessary. And in this case, the namespace in the resource does not satisfy the selector. I'm surprised that works, and I think it's probably ok that it works, but it is that little bit extra confusing.

JustinKuli (changes requested):

test/e2e/case9_templates_test.go
gparvin: I added an AfterAll which I think handles this request.  Thanks!

test/resources/case9_templates/case9-test-policy_copy.yaml
gparvin: Yes, confusing is correct so I have removed them.

yiraeChristineKim (approved):
Justin mentioned already put cleanup into afterall would be better, sometimes it cleans up when spec is running so.. looks good

PR #100, opened 2023-04-06
Requeue propagation errors

JustinKuli:
The main change is to return an error when some clusters fail to be propagated to. The test verifies this works by using a cluster namespace that is Terminating during the propagation.

Also cleans up some random bits of code I noticed while working on this.

Refs:
 - https://issues.redhat.com/browse/ACM-4606


dhaiducek (approved):
LGTM! I went looking for how the `requeueErrorDelay` variables were used originally...and discovered it was part of the cleanup! 😂

PR #102, opened 2023-04-13
Bug: Only one policy violation status was passed into ansiblejob 

yiraeChristineKim:
Create a policy with disabled set to true. Then, create a policyAutomation with everyevent mode and "policyViolationLimits" set to 0.

Ref: https://issues.redhat.com/browse/ACM-4777

Signed-off-by: Yi Rae Kim <yikim@redhat.com>

### DIFF
--- a/controllers/automation/policyPredicate.go
+++ b/controllers/automation/policyPredicate.go
@@ -3,6 +3,7 @@
 package automation
 
 import (
+	"github.com/google/go-cmp/cmp"
 	"sigs.k8s.io/controller-runtime/pkg/event"
 	"sigs.k8s.io/controller-runtime/pkg/predicate"
 
@@ -21,7 +22,7 @@ var policyPredicateFuncs = predicate.Funcs{
 		//nolint:forcetypeassert
 		plcObjOld := e.ObjectOld.(*policiesv1.Policy)
 
-		return plcObjNew.Status.ComplianceState != plcObjOld.Status.ComplianceState
+		return !cmp.Equal(plcObjNew.Status.Status, plcObjOld.Status.Status)
 	},
 	CreateFunc: func(e event.CreateEvent) bool {
 		return false
--- a/controllers/automation/policyautomation_controller.go
+++ b/controllers/automation/policyautomation_controller.go
@@ -360,7 +360,8 @@ func (r *PolicyAutomationReconciler) Reconcile(
 			if len(targetList) > 0 {
 				log.Info("Creating An Ansible job", "targetList", targetList)
 				violationContext, _ := r.getViolationContext(ctx, policy, targetList, policyAutomation)
-				err = common.CreateAnsibleJob(policyAutomation, r.DynamicClient, "scan", violationContext)
+				err = common.CreateAnsibleJob(policyAutomation, r.DynamicClient, "scan",
+					violationContext)
 				if err != nil {
 					return reconcile.Result{RequeueAfter: requeueAfter}, err
 				}
@@ -381,6 +382,16 @@ func (r *PolicyAutomationReconciler) Reconcile(
 			if len(targetList) > 0 {
 				log.Info("Creating an Ansible job", "targetList", targetList)
 
+				AjExist, err := common.MatchPAGeneration(policyAutomation,
+					r.DynamicClient, policyAutomation.GetGeneration())
+				if err != nil {
+					log.Error(err, "Failed to get Ansible job's generation")
+
+					return reconcile.Result{}, err
+				}
+				if AjExist {
+					return reconcile.Result{}, nil
+				}
 				violationContext, _ := r.getViolationContext(ctx, policy, targetList, policyAutomation)
 				err = common.CreateAnsibleJob(
 					policyAutomation,
--- a/controllers/common/ansible.go
+++ b/controllers/common/ansible.go
@@ -5,17 +5,67 @@ package common
 import (
 	"context"
 	"encoding/json"
+	"fmt"
 	"reflect"
 	"regexp"
+	"strconv"
 	"strings"
 
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/client-go/dynamic"
+	ctrl "sigs.k8s.io/controller-runtime"
 
 	policyv1beta1 "open-cluster-management.io/governance-policy-propagator/api/v1beta1"
 )
+
+const (
+	ControllerName             string = "policy-automation"
+	PolicyAutomationLabel      string = "policy.open-cluster-management.io/policyautomation-name"
+	PolicyAutomationGeneration string = "policy.open-cluster-management.io/policyautomation-generation"
+)
+
+var log = ctrl.Log.WithName(ControllerName)
+
+var ansibleJobRes = schema.GroupVersionResource{
+	Group: "tower.ansible.com", Version: "v1alpha1",
+	Resource: "ansiblejobs",
+}
+
+// Check any ansiblejob is made by input genteration number
+// Returning "true" means the policy automation already created ansiblejob with the generation
+func MatchPAGeneration(policyAutomation *policyv1beta1.PolicyAutomation,
+	dynamicClient dynamic.Interface, generation int64,
+) (bool, error) {
+	ansiblejobList, err := dynamicClient.Resource(ansibleJobRes).Namespace(policyAutomation.GetNamespace()).List(
+		context.TODO(), metav1.ListOptions{
+			LabelSelector: fmt.Sprintf("%s=%s", PolicyAutomationLabel, policyAutomation.GetName()),
+		},
+	)
+	if err != nil {
+		log.Error(err, "Failed to get ansiblejob list")
+
+		return false, err
+	}
+
+	ansiblejobLen := len(ansiblejobList.Items)
+	// Check whether new PolicyAutomation
+	if ansiblejobLen == 0 {
+		return false, nil
+	}
+
+	s := strconv.FormatInt(generation, 10)
+
+	for _, aj := range ansiblejobList.Items {
+		annotations := aj.GetAnnotations()
+		if annotations[PolicyAutomationGeneration] == s {
+			return true, nil
+		}
+	}
+
+	return false, nil
+}
 
 // CreateAnsibleJob creates ansiblejob with given PolicyAutomation
 func CreateAnsibleJob(policyAutomation *policyv1beta1.PolicyAutomation,
@@ -25,6 +75,12 @@ func CreateAnsibleJob(policyAutomation *policyv1beta1.PolicyAutomation,
 		Object: map[string]interface{}{
 			"apiVersion": "tower.ansible.com/v1alpha1",
 			"kind":       "AnsibleJob",
+			"metadata": map[string]interface{}{
+				"annotations": map[string]interface{}{
+					PolicyAutomationGeneration: strconv.
+						FormatInt(policyAutomation.GetGeneration(), 10),
+				},
+			},
 			"spec": map[string]interface{}{
 				"job_template_name": policyAutomation.Spec.Automation.Name,
 				"tower_auth_secret": policyAutomation.Spec.Automation.TowerSecret,
@@ -62,6 +118,11 @@ func CreateAnsibleJob(policyAutomation *policyv1beta1.PolicyAutomation,
 
 		mapExtraVars[fieldName] = value
 	}
+
+	label := map[string]string{
+		PolicyAutomationLabel: policyAutomation.GetName(),
+	}
+	ansibleJob.SetLabels(label)
 
 	ansibleJob.Object["spec"].(map[string]interface{})["extra_vars"] = mapExtraVars
 
--- a/test/e2e/case5_policy_automation_test.go
+++ b/test/e2e/case5_policy_automation_test.go
@@ -5,6 +5,7 @@ package e2e
 
 import (
 	"context"
+	"sort"
 	"time"
 
 	. "github.com/onsi/ginkgo/v2"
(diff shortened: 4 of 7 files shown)


test/e2e/case5_policy_automation_test.go
```diff
@@ -423,50 +484,7 @@ var _ = Describe("Test policy automation", func() {
 				return len(ansiblejobList.Items)
 			}, 30, 1).Should(Equal(2))
 
-			By("Removing config map")
```
yiraeChristineKim, line 426: This is repeatable so I put into cleanup function

yiraeChristineKim, line 426: This is repeatable. so I put into cleanupPolicyAutomation func

controllers/automation/policyPredicate.go
```diff
@@ -21,7 +22,7 @@ var policyPredicateFuncs = predicate.Funcs{
 		// nolint: forcetypeassert
 		plcObjOld := e.ObjectOld.(*policiesv1.Policy)
 
-		return plcObjNew.Status.ComplianceState != plcObjOld.Status.ComplianceState
+		return !cmp.Equal(plcObjNew.Status.Status, plcObjOld.Status.Status)
 	},
```
yiraeChristineKim, line 26: this was main reason of the bug

mprahl: Could you please explain this a bit more? I would think we only want to reconcile when the policy status has changed compliance since it seems that this predicate is only looking at replicated policies.

yiraeChristineKim: so when 3 clusters noncompliant in a row, at the first cluster, policy's status changes to noncompliant so reconcile only pickup first cluster status.

mprahl: Oh, I see. I misread the code. The predicate is only allowing root policies and not replicated policies. Then you are checking if the compliance per cluster has changed. This makes sense.

yiraeChristineKim: you right

controllers/common/ansible.go
```diff
@@ -62,6 +90,27 @@ func CreateAnsibleJob(policyAutomation *policyv1beta1.PolicyAutomation,
...
+	// when createnewAnsiblejob is ture and ansiblejob does not exist, skip this part and creates new ansiblejob
+	// when createNewAnsibleJob is false but ansiblejob does not exist, it creates new ansiblejob
+	if !createNewAnsibleJob {
+		existingAnsibleJob := CheckExistAnsibleJob(policyAutomation, dynamicClient)
+
+		if existingAnsibleJob != nil {
+			log.V(2).Info("Modifiy existing ansiblejob")
+
+			// Update with new policyAutomation's policy_violation
+			existingAnsibleJob.Object["spec"].(map[string]interface{})["extra_vars"] = mapExtraVars
```
mprahl, line 102: Creating a new AnsibleJob when more clusters have been violated is the right thing to do because an AnsibleJob might have already triggered a playbook run on Ansible Tower. Updating the `extra_vars` would lead to confusion.

I'm thinking that only your predicate change is needed and not any of the other changes.

yiraeChristineKim: that means if we have 100 clusters that means we create maybe 90 ansiblejobs  depend on  status changing time it could be 100

yiraeChristineKim: oh also we need to change previous e2e test too

mprahl: @yiraeChristineKim yes, it's possible. Those AnsibleJob objects do periodically get cleaned up by the Ansible operator though.

The way the controller-runtime reconcile request queue works is that if there are multiple reconcile requests for the same Policy, it consolidates them to 1 request, so in theory, it should be less than 100. For example, if you have a reconcile request for the root policy and while you're creating the AnsibleJob there have been 5 updates to the root policy, you'd only end up with the 1 additional reconcile request for the root policy.

yiraeChristineKim: oh yeah i guess you right cuz I saw only 2 request when it is 3 cluster changes

mprahl: @yiraeChristineKim let me know when you're ready for another review

test/e2e/case5_policy_automation_test.go
```diff
@@ -361,21 +440,21 @@ var _ = Describe("Test policy automation", func() {
...
+			By("Should create any new ansiblejob or not when policies become Compliant")
 			Consistently(func() interface{} {
 				ansiblejobList, err := clientHubDynamic.Resource(gvrAnsibleJob).List(
 					context.TODO(), metav1.ListOptions{},
 				)
 				Expect(err).To(BeNil())
 
 				return len(ansiblejobList.Items)
-			}, 15, 1).Should(Equal(1))
+			}, 15, 1).Should(BeNumerically(">=", ansiblelistlen))
```
yiraeChristineKim, line 451: during change to compliant status, it can create ansiblejob

controllers/automation/policyautomation_controller.go
```diff
@@ -195,6 +196,29 @@ func (r *PolicyAutomationReconciler) getViolationContext(
 		return violationContext, nil
 	}
 
+	// targetList is noncompliant list, this step compares compliant
+	// so that check any changes happened during reconcile
+	for _, clusterName := range targetList {
```
mprahl, line 201: I don't think this block of code is required. If there is a mismatch in the status in the root policy and replicated policy, it means the policy on the cluster has become compliant and the root policy status will soon be updated or already has but the cached value in memory is stale.

Regardless, the cluster became noncompliant so that should still trigger an AnsibleJob. For example, if a policy goes non-compliant and then quickly to compliant and your Ansible job creates a ServiceNow ticket to notify a systems administrator, you'd still want the ServiceNow ticket created.

controllers/automation/policyautomation_controller.go
```diff
@@ -378,10 +409,24 @@ func (r *PolicyAutomationReconciler) Reconcile(
 		} else if policyAutomation.Spec.Mode == policyv1beta1.Once {
 			log := log.WithValues("mode", string(policyv1beta1.Once))
 			targetList := common.FindNonCompliantClustersForPolicy(policy)
+
 			if len(targetList) > 0 {
+				//nolint:all
+				exist, _ := common.CheckAnsiblejobExist(policyAutomation, r.DynamicClient)
```
mprahl, line 415: You don't need to check for the existing AnsibleJob because the way the "Once" mode works, it creates an AnsibleJob and then the mode is set to disabled. If it's put back to "Once" mode, it means the user explicitly did this and wants the automation to be rerun on the next triggering event.

controllers/automation/policyautomation_controller.go
```diff
@@ -464,7 +507,12 @@ func (r *PolicyAutomationReconciler) Reconcile(
 							// new event from compliant to non-compliant
 							eventMap[clusterName] = clusterEvent
 						}
-					} // Otherwise, the policy keeps non-compliant since originalStartTime, do nothing
+					} else {
+						// Otherwise, the policy keeps non-compliant since originalStartTime, do nothing
+						if !trimmedTargetMap[clusterName] {
```
mprahl, line 512: Could you please explain this change a bit more?

controllers/automation/policyautomation_controller.go
JustinKuli: I think suggested this, because there had been an issue in the tests where the reconcile was triggered twice, and the second time wasn't "seeing" that the mode had been set to disabled. It would also be possible in general for the AnsibleJob to be created, but the update to the automation (setting it to disabled) might not succeed for some other reason, meaning two AnsibleJobs would be created for one `Once` event.

yiraeChristineKim: This is the part that Justin help me with. I guess updating mode status takes a while. Therefore, when it falls to second event it is still mode=once so it creates second ansiblejob

controllers/automation/policyautomation_controller.go
yiraeChristineKim: This is not critical but I assumed maybe this is unclear(confusing) for the customers, cuz policyautomation says noncompliant but ansiblejob is compliant. also, I would like to avoid creating useless ansiblejob as much as possible. can delete it if u want

controllers/automation/policyautomation_controller.go
yiraeChristineKim: without this, it creates one by one, in other words, it creates 2-3 ansible jobs when we have 3 policies(clusters) changed. also our ticket 
This is their expectation so 
Expected results:
All violations status of all clusters should be passed

controllers/automation/policyautomation_controller.go
mprahl: In that case, I'd prefer to "patch" instead of "update" the PolicyAutomation object when setting this to disabled to significantly reduce the change of an update failing. Otherwise, you'd need to be able to distinguish between a duplicate AnsibleJob trigger versus a user resetting the mode back to "Once".

controllers/common/common.go
```diff
@@ -30,6 +30,10 @@ const (
 
 var ErrInvalidLabelValue = errors.New("unexpected format of label value")
 
+func IsClusterNameMatch(clusterName string, policy policiesv1.Policy) bool {
```
mprahl, line 33: This function is unused?

controllers/automation/policyautomation_controller.go
```diff
@@ -309,13 +309,13 @@ func (r *PolicyAutomationReconciler) Reconcile(
 			"clusterCount", strconv.Itoa(len(targetList)))
 
 		violationContext, _ := r.getViolationContext(ctx, policy, targetList, policyAutomation)
-
```
mprahl, line 312: Could you please not change the whitespace in unrelated parts of the code? If you want to commit this, then please add it in a separate commit.

controllers/automation/policyautomation_controller.go
```diff
@@ -378,10 +380,20 @@ func (r *PolicyAutomationReconciler) Reconcile(
...
+
+				if exist {
+					return reconcile.Result{}, nil
+				}
+
 				log.Info("Creating an Ansible job", "targetList", targetList)
 
 				violationContext, _ := r.getViolationContext(ctx, policy, targetList, policyAutomation)
+
+				//nolint:all
```
mprahl, line 396: Why is this needed?

controllers/automation/policyautomation_controller.go
mprahl: This comment also applies to elsewhere in the diff.

controllers/automation/policyautomation_controller.go
```diff
@@ -491,7 +500,7 @@ func (r *PolicyAutomationReconciler) Reconcile(
 				for clusterName := range trimmedTargetMap {
 					trimmedTargetList = append(trimmedTargetList, clusterName)
 				}
-				log.Info("Creating An Ansible job", "trimmedTargetList", trimmedTargetList)
```
mprahl, line 494: Is there a reason to remove this log message?

controllers/common/ansible.go
```diff
@@ -73,8 +128,9 @@ func CreateAnsibleJob(policyAutomation *policyv1beta1.PolicyAutomation,
 		Resource: "ansiblejobs",
 	}
 
-	// Ansible tower API requires the lower case naming
-	ansibleJob.SetGenerateName(strings.ToLower(policyAutomation.GetName() + "-" + mode + "-"))
+	order := GetAnsibleJobOrder(policyAutomation, dynamicClient)
+	ansibleJob.SetGenerateName(strings.ToLower(policyAutomation.GetName() +
+		"-" + mode + "-" + strconv.Itoa(order) + "-"))
```
mprahl, line 133: Why is this change necessary?

controllers/automation/policyautomation_controller.go
yiraeChristineKim: I tried patch same result

mprahl: @yiraeChristineKim what was the result you saw? The reconcile ran again?

yiraeChristineKim: so update disable doesn't work for the second event. the second event is supposed not to fall in once. In the second event update failed(error message changed something..).  As a result, randomly it creates 1 ansiblejob or 2 ansiblejobs

yiraeChristineKim: I added sleep for 100 milliseconds.. this works well

mprahl: @yiraeChristineKim what if you updated `policyAuomtationPredicateFuncs` so that when the mode is disabled, it doesn't trigger a reconcile?

controllers/common/ansible.go
yiraeChristineKim: to sort easier, since we move checansiblejob it is ok to remove

controllers/automation/policyautomation_controller.go
yiraeChristineKim: same result, first try e2e test, it creates one ansiblejob but second try, it creates 2 ansiblejobs

mprahl: After thinking about this, it's not the PolicyAutomation causing the reconcile in the case you're fixing but an additional replicated policy causing the reconcile.

yiraeChristineKim: oh wait what I read it  **This is the equivalent of calling Watches** https://github.com/kubernetes-sigs/controller-runtime/blob/v0.14.6/pkg/builder/controller.go#L83 i thought watch for same

controllers/common/ansible.go
yiraeChristineKim: I tried to get the latest ansiblejob but tried to sort with time, but time in ansiblejob, the second is the smallest one. so hard to sort, ` creationTimestamp: "2023-04-27T18:56:04Z"` in  yaml file

yiraeChristineKim: is there best way to sort by createtime

test/e2e/case5_policy_automation_test.go
```diff
@@ -205,11 +285,36 @@ var _ = Describe("Test policy automation", func() {
 				policyAutomation.Object["spec"].(map[string]interface{})["mode"],
 			).To(Equal(string(policyv1beta1.Disabled)))
 
-			By("Patching policy to make both clusters back to Compliant")
+			By("Change mode to once again, should create one more ansiblejob")
```
yiraeChristineKim, line 288: important part of e2e test for this bug fix

controllers/common/ansible.go
```diff
@@ -5,25 +5,109 @@ package common
...
+	stringGeneration := lastAJAnnotation[PolicyAutomationGeneration]
+
+	return stringGeneration, nil
+}
+
 // CreateAnsibleJob creates ansiblejob with given PolicyAutomation
 func CreateAnsibleJob(policyAutomation *policyv1beta1.PolicyAutomation,
 	dynamicClient dynamic.Interface, mode string, violationContext policyv1beta1.ViolationContext,
 ) error {
+	order := GetAnsibleJobOrder(policyAutomation, dynamicClient)
```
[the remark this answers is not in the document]
JustinKuli: It seems like the console might already sort it based on start and finish time, so that might be good enough? https://github.com/stolostron/console/blob/main/frontend/src/routes/Governance/components/AutomationDetailsSidebar.tsx#L102 (note, this is just the first place I found, I didn't check that it's the table you might be thinking of)

test/e2e/case5_policy_automation_test.go
```diff
@@ -170,12 +267,13 @@ var _ = Describe("Test policy automation", func() {
...
 			Expect(err).To(BeNil())
-			spec := ansiblejobList.Items[0].Object["spec"]
+			index := len(ansiblejobList.Items) - 1
+			spec := ansiblejobList.Items[index].Object["spec"]
 			extraVars := spec.(map[string]interface{})["extra_vars"].(map[string]interface{})
 			Expect(extraVars["policy_name"]).To(Equal("case5-test-policy"))
 			Expect(extraVars["policy_namespace"]).To(Equal(testNamespace))
 			Expect(extraVars["hub_cluster"]).To(Equal("millienium-falcon.tatooine.local"))
-			Expect(len(extraVars["target_clusters"].([]interface{}))).To(Equal(1))
+			Expect(len(extraVars["target_clusters"].([]interface{}))).Should(BeNumerically(">=", 1))
```
yiraeChristineKim, line 276: last ansiblejob include managed2 managed3 usually

controllers/common/ansible.go
```diff
@@ -5,26 +5,111 @@ package common
...
+
+	if err != nil {
+		log.Error(err, "Faile to get ansiblejob list")
+
+		return "0", err
+	}
+
+	if ansiblejobLen == 0 {
+		return "0", nil
+	}
```
dhaiducek, line 75: Mainly a spelling error ("Faile"), but I think reordering these makes sense also.
```suggestion
	if err != nil {
		log.Error(err, "Failed to get ansiblejob list")

		return "0", err
	}

	ansiblejobLen := len(ansiblejobList.Items)
	if ansiblejobLen == 0 {
		return "0", nil
	}
```

controllers/common/ansible.go
```diff
@@ -5,26 +5,111 @@ package common
...
 	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/client-go/dynamic"
+	ctrl "sigs.k8s.io/controller-runtime"
 
 	policyv1beta1 "open-cluster-management.io/governance-policy-propagator/api/v1beta1"
 )
 
+const (
+	ControllerName             string = "policy-automation"
+	PolicyAutomationLabel      string = "ansiblejob.policy.open-cluster-management.io/policyautomation-name"
```
dhaiducek, line 26: I'd think this would be PolicyAutomation instead of AnsibleJob?
```suggestion
	PolicyAutomationLabel      string = "policyautomation.policy.open-cluster-management.io/policyautomation-name"
```

dhaiducek (commented):
Some minor comments to consider, but overall looks great to me! Nicely done!

controllers/common/ansible.go
yiraeChristineKim: This label attaches to ansiblejob not policyautomation. In other words, add "policyautomation name" on ansiblejob

yiraeChristineKim: good good let me double check

dhaiducek: (Similar to the other keys here)

controllers/common/ansible.go
```diff
@@ -5,26 +5,109 @@ package common
...
 	"k8s.io/client-go/dynamic"
+	ctrl "sigs.k8s.io/controller-runtime"
 
 	policyv1beta1 "open-cluster-management.io/governance-policy-propagator/api/v1beta1"
 )
 
+const (
+	ControllerName             string = "policy-automation"
+	PolicyAutomationLabel      string = "policyautomation.policy.open-cluster-management.io/policyautomation-name"
+	PolicyAutomationGeneration string = "policyautomation.policy.open-cluster-management.io/policyautomation-generation"
```
dhaiducek, line 27: Sorry to go back and forth on this, but I think leaving this as just the API group might make the most sense and follow previous keys.

```suggestion
	PolicyAutomationLabel      string = "policy.open-cluster-management.io/policyautomation-name"
	PolicyAutomationGeneration string = "policy.open-cluster-management.io/policyautomation-generation"
```

controllers/common/ansible.go
```diff
@@ -5,26 +5,109 @@ package common
...
+	)
+	if err != nil {
+		return 0
+	}
+
+	return len(ansiblejobList.Items) + 1
+}
+
+// Sort and get policy automation's generation of the last ansiblejob
+func GetLastAjGeneration(policyAutomation *policyv1beta1.PolicyAutomation,
```
mprahl, line 56: Could this return an integer instead of a string?

controllers/common/ansible.go
```diff
@@ -5,26 +5,109 @@ package common
...
+		Namespace(policyAutomation.GetNamespace()).List(
+		context.TODO(), metav1.ListOptions{
+			LabelSelector: fmt.Sprintf("%s=%s", PolicyAutomationLabel, policyAutomation.GetName()),
+		},
+	)
+	if err != nil {
+		return 0
+	}
+
+	return len(ansiblejobList.Items) + 1
```
mprahl, line 52: This is not stable since AnsibleJob objects automatically get pruned over time.

mprahl (changes requested):

controllers/common/ansible.go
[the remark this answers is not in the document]
yiraeChristineKim: I should get the latest ansiblejob.  I used timestamp for sort, but the smallest unit is second so hard to sort.

controllers/common/ansible.go
yiraeChristineKim: I changed to "policiy" :)

controllers/automation/policyautomation_controller.go
```diff
@@ -381,6 +382,16 @@ func (r *PolicyAutomationReconciler) Reconcile(
 			if len(targetList) > 0 {
 				log.Info("Creating an Ansible job", "targetList", targetList)
 
+				lastAjGeneration, err := common.GetLastAjGeneration(policyAutomation, r.DynamicClient)
```
mprahl, line 385: To write down what I recommended earlier. Since the goal is to see if there is an AnsibleJob for this PolicyAutomation of this generation, you don't need to know the latest one, just that there is one that matches the criteria.

controllers/common/ansible.go
```diff
@@ -5,26 +5,109 @@ package common
...
+	}
+
+	ansiblejobLen := len(ansiblejobList.Items)
+	if ansiblejobLen == 0 {
+		return "0", nil
+	}
+
+	// sort by AnsiblejobIndex annotation
+	sort.Slice(ansiblejobList.Items, func(i, j int) bool {
+		p1, err := strconv.Atoi(ansiblejobList.Items[i].GetAnnotations()[AnsiblejobIndex])
```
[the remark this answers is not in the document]
yiraeChristineKim: good

test/e2e/case5_policy_automation_test.go
```diff
@@ -538,33 +629,34 @@ var _ = Describe("Test policy automation", func() {
 				return allUpdated
 			}, 30, 1).Should(BeTrue())
 
-			By("Should only create one ansiblejob for the first event during delayAfterRunSeconds period")
+			By("The last ansiblejob Should has managed3 in target_clsuter for" +
```
mprahl, line 632:
```suggestion
			By("checking the last AnsibleJob has managed3 in target_clsuter for" +
```

mprahl:
@yiraeChristineKim could you please troubleshoot the CI failure? Otherwise, this looks good to me.

mprahl (approved):

PR #107, opened 2023-05-02
Sanitize AnsibleJob name

JustinKuli:
Refs:
 - https://issues.redhat.com/browse/ACM-5093


dhaiducek (approved):
LGTM! I'm not at all sure on this, but just wondering if there's any logic in the UI that might also need to be updated.

PR #106, opened 2023-04-19
Upgrade `go-template-utils` to v3.2.0

dhaiducek:
ref: https://issues.redhat.com/browse/ACM-5032

PR #109, opened 2023-05-05
More package upgrades

dhaiducek:
Through my package upgrade work, I found that the K8s API package was actually blocking upgrade paths. Replacing it let more upgrades through.

Followup to:
- #106


go.mod
```diff
@@ -94,5 +94,6 @@ require (
 replace (
 	golang.org/x/crypto => golang.org/x/crypto v0.0.0-20220321153916-2c7772ba3064 // CVE-2021-43565
 	golang.org/x/text => golang.org/x/text v0.3.8 // CVE-2022-32149
-	k8s.io/client-go => k8s.io/client-go v0.26.1
+	k8s.io/api => k8s.io/api v0.26.4
```
JustinKuli, line 97: Why replace it here, as opposed to just using `v0.26.4` in the top require section? My understanding was that we have client-go down here because we need to declare it as `v12.0.0+incompatible` for some other dependency...

dhaiducek: I suppose it'd work either way, but for some reason `go get -u` won't work unless this replace is here:
```
$ go get -u ./...
open-cluster-management.io/governance-policy-propagator imports
        k8s.io/client-go/kubernetes/scheme imports
        k8s.io/api/resource/v1alpha1: cannot find module providing package k8s.io/api/resource/v1alpha1
```
So I think having the replace here will create less work in the future running the upgrade command. I'll add a note if that makes sense to you.

dhaiducek: (The `v1alpha1` resource API was bumped to `v1alpha2` in `v0.27.1`.)

JustinKuli: I guess... if it works, it works 🤷

JustinKuli (approved):

go.mod
dhaiducek: Yeah--I'm starting to get a handle on Go packages, but it's definitely still largely a black box to me.

dhaiducek:
Re-running CI, but I created an issue for that flaky templates test: https://issues.redhat.com/browse/ACM-5349

PR #110, opened 2023-05-09
Support selective policy enforcement

Missxiaoguo:
Two new fields are added in placement binding API:
- remediationActionOverride.remediationAction: provide the ability to override the policy's remediationAction to enforce.
- subFilter: provide the ability to select a subset of bound clusters where the policy should be enforced to.

The propagator is now processing all the placement bindings first and get a list of all the clusters decisions to be handled for the policy, then concurrently processing all decisions.

Each determined cluster decision contains the cluster where the policy should be replicated and the policy remediationActionOverride which has value of either empty or enforce.

When replicating the policy on cluster namespace, update the remediationAction to enforce if its root policy is not enforce and it's selected to be enforced.

Refs:
- https://issues.redhat.com/browse/ACM-1397
- https://issues.redhat.com/browse/CNF-7853


Missxiaoguo:
@mprahl @JustinKuli @dhaiducek 
Finally get this posted. Please help to review! Thank you all!

api/v1/placementbinding_types.go
```diff
@@ -40,6 +40,13 @@ type PlacementSubject struct {
 // PlacementBindingStatus defines the observed state of PlacementBinding
 type PlacementBindingStatus struct{}
 
+// RemediationActionOverride defines the overrides of the policy RemediationAction
+type RemediationActionOverride struct {
+	// +kubebuilder:validation:Optional
+	// +kubebuilder:validation:Enum=Enforce;enforce
```
jc-rh, line 46: I started wondering if the opposite case exists. Could someone want to keep the default action as enforce for a set of clusters and override to inform for another set or a subset of the parent placement?

dhaiducek: I was wondering the same thing. Could the `RemediationAction` field from the Policy struct be referenced here instead of redefining it here?
https://github.com/open-cluster-management-io/governance-policy-propagator/blob/0370f385cb946b4ebf01ad9b320b13ab5d1eff5c/api/v1/policy_types.go#L13-L14

api/v1/placementbinding_types.go
```diff
@@ -49,6 +56,11 @@ type PlacementBindingStatus struct{}
 type PlacementBinding struct {
 	metav1.TypeMeta   `json:",inline"`
 	metav1.ObjectMeta `json:"metadata,omitempty"`
+	// +kubebuilder:validation:Optional
+	RemediationActionOverride RemediationActionOverride `json:"remediationActionOverride,omitempty"`
```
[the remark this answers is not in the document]
JustinKuli: I like the `overrides` idea. Thinking out loud, maybe the type would be called `BindingOverrides`, then using it from inside the propagator would be like `policyv1.BindingOverrides.RemediationAction`

api/v1/placementbinding_types.go
JustinKuli: I think the logic was simplified by only being able to override in one direction. If we allow an override to `inform`, then what happens if I have two Bindings, one overriding to `enforce`, one overriding to `inform`, and they *both* target the same cluster?

It's not in general a bad idea, but it complicates things.

api/v1/placementbinding_types.go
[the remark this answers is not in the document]
Missxiaoguo: Yeah, If we anticipate the need to support multiple types of overrides in the futher, I prefer better to have a separate struct to encapsulate all possible overrides. It's eaiser to maintain as the number of overrides increases

controllers/propagator/propagation.go
```diff
@@ -279,6 +292,147 @@ func (set decisionSet) namespaces() []string {
...
+// If a match is found, it returns the index of the matched decision; otherwise, it returns -1.
+func containDecision(clusterDecisions []clusterDecision, decision appsv1.PlacementDecision) int {
+	for id, clusterDecision := range clusterDecisions {
+		if clusterDecision.Cluster == decision {
+			return id
+		}
+	}
+
+	return -1
+}
```
JustinKuli, line 343: I wonder if this kind of `clusterDecisions` thing could have been cleaner as a map of some kind. But it is not necessary to change at this point.

JustinKuli (commented):
Thanks for including tests! This is looking pretty good to me.

api/v1/placementbinding_types.go
jc-rh: Yes, definitely. We can avoid this for now and expand the enum later which should be a backward compatible API change, when we have a concrete use case for the opposite override. However, we will have to deal with this kind of conflict as soon as we extend the overrides to other non binary fields, such as evaluation interval or dependency list. Each override may need different conflict resolution. A common one could be reporting as an error if we decide to be super strict on multiple bindings with overlapping overrides.

controllers/propagator/propagation.go
Missxiaoguo: yeah, agree. In our case, we don't care the order of the decisions. A map is more proper here for fast lookup and enforce uniqueness.
 As we are changing the API to` policiesv1.BindingOverrides`, we could use this map
`map[appsv1.PlacementDecision]policiesv1.BindingOverrides{}` to collect all information to be handled.

Missxiaoguo: Updated to use a map. Hope it looks cleaner now.

JustinKuli (approved):
/hold

I think using a map here makes it even more clean, thanks! Just holding if other reviewers have any more comments.

imiller0 (commented):
Looks good. A minor suggestion on using enum in preference of boolean. 
Also there is an added testcase with subfiltering, but is there test coverage for override with subfiltering disabled (ie adding a new cluster into the decision using the override)?

Missxiaoguo:
> Looks good. A minor suggestion on using enum in preference of boolean. Also there is an added testcase with subfiltering, but is there test coverage for override with subfiltering disabled (ie adding a new cluster into the decision using the override)?

A unit test covers the overriding with disabled subFilter. I think it's good enough.

api/v1/placementbinding_types.go
```diff
@@ -49,6 +57,11 @@ type PlacementBindingStatus struct{}
 type PlacementBinding struct {
 	metav1.TypeMeta   `json:",inline"`
 	metav1.ObjectMeta `json:"metadata,omitempty"`
+	// +kubebuilder:validation:Optional
+	BindingOverrides BindingOverrides `json:"bindingOverrides,omitempty"`
+	// This field provides the ability to select a subset of bound clusters. It's default to false.
+	// +kubebuilder:validation:Optional
+	SubFilter bool `json:"subFilter,omitempty"`
```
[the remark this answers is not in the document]
jc-rh: To really make it truly usable in the future, we probably need give the field a more generic name (bindingMode?) and has the value as none|subfilter.

imiller0 (commented):
Looks good to me.

JustinKuli (approved):
Looks great, thanks again!

dhaiducek (approved):
/unhold

This looks great--thanks so much for the PR!

PR #117, opened 2023-06-16
Webhook to check name + namespace exceeds 63 characters

yiraeChristineKim:
Description: No validation if name + namespace exceeds 63 characters and Policy applied via CLI or Gitops now

After fix: `Error from server (Forbidden): error when creating "policy.yaml": admission webhook "policy.open-cluster-management.io.webhook" denied the request: the combined length of namespace and policy name (namespaceName.policyName) should not exceed 63 characters`

This error message shows on terminal when policy name + namespace exceeds 63 characters

Ref: https://issues.redhat.com/browse/ACM-4394

### DIFF
--- a/.github/workflows/kind.yml
+++ b/.github/workflows/kind.yml
@@ -81,8 +81,13 @@ jobs:
 
     - name: Verify Deployment Configuration
       run: |
+        make webhook
         make build-images
         make kind-deploy-controller-dev
+    
+    - name: E2E Tests for Webhook
+      run: |
+        make e2e-test-webhook
 
     - name: Debug
       if: ${{ failure() }}
@@ -93,3 +98,4 @@ jobs:
       if: ${{ always() }}
       run: |
         make kind-delete-cluster
+        
--- a/Makefile
+++ b/Makefile
@@ -217,10 +217,20 @@ kustomize: ## Download kustomize locally if necessary.
 GINKGO = $(LOCAL_BIN)/ginkgo
 
 .PHONY: kind-bootstrap-cluster
-kind-bootstrap-cluster: kind-create-cluster install-crds kind-deploy-controller install-resources
+kind-bootstrap-cluster: kind-create-cluster install-crds webhook kind-deploy-controller install-resources
 
 .PHONY: kind-bootstrap-cluster-dev
 kind-bootstrap-cluster-dev: kind-create-cluster install-crds install-resources
+
+webhook:
+	-kubectl create ns $(KIND_NAMESPACE)
+	@echo installing cert-manager
+	kubectl apply -f https://github.com/cert-manager/cert-manager/releases/download/v1.12.0/cert-manager.yaml
+	@echo "wait until pods are up"
+	kubectl wait deployment -n cert-manager cert-manager --for condition=Available=True --timeout=90s
+	kubectl wait --for=condition=Ready pod -l app.kubernetes.io/instance=cert-manager -n cert-manager --timeout=30s 
+	sed 's/namespace: open-cluster-management/namespace: $(KIND_NAMESPACE)/g' deploy/webhook.yaml | kubectl apply -f-
+	
 
 .PHONY: kind-deploy-controller
 kind-deploy-controller: manifests
@@ -281,7 +291,7 @@ e2e-dependencies:
 
 .PHONY: e2e-test
 e2e-test: e2e-dependencies
-	$(GINKGO) -v --fail-fast $(E2E_TEST_ARGS) test/e2e
+	$(GINKGO) -v --fail-fast $(E2E_TEST_ARGS) --label-filter="!webhook" test/e2e
 
 .PHONY: e2e-test-coverage
 e2e-test-coverage: E2E_TEST_ARGS = --json-report=report_e2e.json --output-dir=.
@@ -298,6 +308,9 @@ e2e-run-instrumented: e2e-build-instrumented
 .PHONY: e2e-stop-instrumented
 e2e-stop-instrumented:
 	ps -ef | grep '$(IMG)' | grep -v grep | awk '{print $$2}' | xargs kill
+
+e2e-test-webhook:
+	$(GINKGO) -v --fail-fast --label-filter="webhook" test/e2e 
 
 .PHONY: e2e-debug
 e2e-debug:
--- a/README.md
+++ b/README.md
@@ -62,6 +62,15 @@ make e2e-dependencies
 make e2e-test
 ```
 
+### How to run webhook locally
+
+```bash
+--enable-webhooks=true
+```
+
+> **Limit**: If you want to run the webhook locally, you need to generate certificates and place them in `/tmp/k8s-webhook-server/serving-certs/tls.{crt,key}`. If you’re not running a local API server, you’ll also need to figure out how to proxy traffic from the remote cluster to your local webhook server. For this reason, Kubebuilder generally recommends disabling webhooks when doing your local code-run-test cycle. To disable it, please supply the `--enable-webhooks=false` argument when running the controller.
+> For more information, visit https://book.kubebuilder.io/cronjob-tutorial/running.html
+
 ### Clean up
 
 ```
@@ -92,6 +101,3 @@ The following environment variables can be set to configure the controller:
 - The `governance-policy-propagator` is part of the `open-cluster-management` community. For more information, visit:
   [open-cluster-management.io](https://open-cluster-management.io).
 
-<!---
-Date: 2022-03-2
--->
--- a/api/v1/policy_webhook.go
+++ b/api/v1/policy_webhook.go
@@ -1,0 +1,66 @@
+// Copyright (c) 2021 Red Hat, Inc.
+// Copyright Contributors to the Open Cluster Management project
+
+package v1
+
+import (
+	"errors"
+	"unicode/utf8"
+
+	"k8s.io/apimachinery/pkg/runtime"
+	ctrl "sigs.k8s.io/controller-runtime"
+	logf "sigs.k8s.io/controller-runtime/pkg/log"
+	"sigs.k8s.io/controller-runtime/pkg/webhook"
+	"sigs.k8s.io/controller-runtime/pkg/webhook/admission"
+)
+
+var (
+	// log is for logging in this package.
+	policylog = logf.Log.WithName("policy-validating-webhook")
+	errName   = errors.New("the combined length of the policy namespace and name " +
+		"<namespace>.<name> cannot exceed 63 characters")
+)
+
+func (r *Policy) SetupWebhookWithManager(mgr ctrl.Manager) error {
+	return ctrl.NewWebhookManagedBy(mgr).
+		For(r).
+		Complete()
+}
+
+//+kubebuilder:webhook:path=/validate-policy-open-cluster-management-io-v1-policy,mutating=false,failurePolicy=Ignore,sideEffects=None,groups=policy.open-cluster-management.io,resources=policies,verbs=create,versions=v1,name=policy.open-cluster-management.io.webhook,admissionReviewVersions=v1
+
+var _ webhook.Validator = &Policy{}
+
+// ValidateCreate implements webhook.Validator so a webhook will be registered for the type
+func (r *Policy) ValidateCreate() (admission.Warnings, error) {
+	policylog.Info("Validate policy creation request", "name", r.Name)
+
+	return r.validateName()
+}
+
+// ValidateUpdate implements webhook.Validator so a webhook will be registered for the type
+func (r *Policy) ValidateUpdate(_ runtime.Object) (admission.Warnings, error) {
+	return nil, nil
+}
+
+// ValidateDelete implements webhook.Validator so a webhook will be registered for the type
+func (r *Policy) ValidateDelete() (admission.Warnings, error) {
+	return nil, nil
+}
+
+// validate the policy name and namespace length
+func (r *Policy) validateName() (admission.Warnings, error) {
+	policylog.Info("Validating the policy name through a validating webhook")
+
+	// replicated policies don't need pass this validation
+	if _, ok := r.GetLabels()["policy.open-cluster-management.io/root-policy"]; ok {
+		return nil, nil
+	}
+
+	// 1 character for "."
+	if (utf8.RuneCountInString(r.Name) + utf8.RuneCountInString(r.Namespace)) > 62 {
+		return nil, errName
+	}
+
+	return nil, nil
+}
--- a/deploy/manager/manager.yaml
+++ b/deploy/manager/manager.yaml
@@ -2,16 +2,22 @@
 apiVersion: apps/v1
 kind: Deployment
 metadata:
+  labels:
+    webhook-origin: governance-policy-propagator
   name: governance-policy-propagator
 spec:
   replicas: 1
   selector:
     matchLabels:
       name: governance-policy-propagator
+      webhook-origin: governance-policy-propagator
   template:
     metadata:
+      annotations:
+        kubectl.kubernetes.io/default-container: governance-policy-propagator
       labels:
         name: governance-policy-propagator
+        webhook-origin: governance-policy-propagator
     spec:
       serviceAccountName: governance-policy-propagator
       containers:
@@ -27,7 +33,14 @@ spec:
             - containerPort: 8383
               protocol: TCP
               name: http
+            - containerPort: 9443
+              protocol: TCP
+              name: webhook-http
           imagePullPolicy: Always
+          volumeMounts:
+          - mountPath: /tmp/k8s-webhook-server/serving-certs
+            name: cert
+            readOnly: true
           env:
             - name: WATCH_NAMESPACE
               value: ""
@@ -37,6 +50,11 @@ spec:
                   fieldPath: metadata.name
             - name: OPERATOR_NAME
               value: "governance-policy-propagator"
+      volumes:
+      - name: cert
+        secret:
+          defaultMode: 420
+          secretName: propagator-webhook-server-cert
 ---
 kind: ClusterRoleBinding
 apiVersion: rbac.authorization.k8s.io/v1
--- a/deploy/webhook.yaml
+++ b/deploy/webhook.yaml
@@ -1,0 +1,61 @@
+apiVersion: v1
+kind: Service
+metadata:
+  name: propagator-webhook-service
+  namespace: open-cluster-management
+spec:
+  ports:
+  - port: 443
+    protocol: TCP
+    targetPort: 9443
+  selector:
+    webhook-origin: governance-policy-propagator
+---
+apiVersion: cert-manager.io/v1
+kind: Certificate
+metadata:
+  name: propagator-webhook-serving-cert
+  namespace: open-cluster-management
+spec:
+  dnsNames:
+  - propagator-webhook-service.open-cluster-management.svc
+  - propagator-webhook-service.open-cluster-management.svc.cluster.local
+  issuerRef:
+    kind: Issuer
+    name: propagator-webhook-selfsigned-issuer
+  secretName: propagator-webhook-server-cert
+---
+apiVersion: cert-manager.io/v1
+kind: Issuer
+metadata:
+  name: propagator-webhook-selfsigned-issuer
+  namespace: open-cluster-management
+spec:
+  selfSigned: {}
+---
+apiVersion: admissionregistration.k8s.io/v1
+kind: ValidatingWebhookConfiguration
+metadata:
+  annotations:
+    cert-manager.io/inject-ca-from: open-cluster-management/propagator-webhook-serving-cert
+  name: propagator-webhook-validating-configuration
+webhooks:
+- admissionReviewVersions:
+  - v1
+  clientConfig:
+    service:
+      name: propagator-webhook-service
+      namespace: open-cluster-management
+      path: /validate-policy-open-cluster-management-io-v1-policy
+  failurePolicy: Ignore
+  name: policy.open-cluster-management.io.webhook
+  rules:
+  - apiGroups:
+    - policy.open-cluster-management.io
+    apiVersions:
+    - v1
+    operations:
+    - CREATE
+    resources:
+    - policies
+  sideEffects: None
--- a/main.go
+++ b/main.go
@@ -90,12 +90,15 @@ func main() {
 	var enableLeaderElection bool
 	var probeAddr string
 	var keyRotationDays, keyRotationMaxConcurrency, policyMetricsMaxConcurrency, policyStatusMaxConcurrency uint
+	var enableWebhooks bool
 
 	pflag.StringVar(&metricsAddr, "metrics-bind-address", ":8383", "The address the metric endpoint binds to.")
 	pflag.StringVar(&probeAddr, "health-probe-bind-address", ":8081", "The address the probe endpoint binds to.")
 	pflag.BoolVar(&enableLeaderElection, "leader-elect", true,
 		"Enable leader election for controller manager. "+
 			"Enabling this will ensure there is only one active controller manager.")
+	pflag.BoolVar(&enableWebhooks, "enable-webhooks", true,
+		"Enable the policy validating webhook")
 	pflag.UintVar(
 		&keyRotationDays,
 		"encryption-key-rotation",
@@ -299,6 +302,13 @@ func main() {
 	}).SetupWithManager(mgr); err != nil {
 		log.Error(err, "Unable to create controller", "controller", rootpolicystatusctrl.ControllerName)
 		os.Exit(1)
+	}
+
+	if enableWebhooks {
+		if err = (&policyv1.Policy{}).SetupWebhookWithManager(mgr); err != nil {
+			log.Error(err, "unable to create webhook", "webhook", "Policy")
+			os.Exit(1)
+		}
 	}
 
 	//+kubebuilder:scaffold:builder
--- a/test/e2e/case17_policy_webhook_test.go
+++ b/test/e2e/case17_policy_webhook_test.go
@@ -1,0 +1,85 @@
+// Copyright (c) 2020 Red Hat, Inc.
+// Copyright Contributors to the Open Cluster Management project
+
+package e2e
+
+import (
+	"context"
+	"unicode/utf8"
+
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+	"open-cluster-management.io/governance-policy-propagator/test/utils"
+)
+
+const (
+	case17PolicyLongYaml       string = "../resources/case17_policy_webhook/case17_policy_long.yaml"
+	case17PolicyReplicatedYaml string = "../resources/case17_policy_webhook/" +
+		"case17_policy_replicate.yaml"
+	longNamesapce              string = "long-long-long-long-long-long-long"
+	case17PolicyReplicatedName string = "case17-test-policy-replicated-longlong"
+	errMsg                     string = `admission webhook "policy.open-cluster-management.io.webhook" denied the ` +
+		`request: the combined length of the policy namespace and name <namespace>.<name> ` +
+		`cannot exceed 63 characters`
+)
+
+var _ = Describe("Test policy webhook", Label("webhook"), Ordered, func() {
+	BeforeAll(func() {
+		_, err := utils.KubectlWithOutput("create",
+			"ns", longNamesapce,
+		)
+		Expect(err).ShouldNot(HaveOccurred())
+	})
+	AfterAll(func() {
+		// cleanup
+		_, err := utils.KubectlWithOutput("delete",
+			"ns", longNamesapce,
+			"--ignore-not-found",
+		)
+		Expect(err).ShouldNot(HaveOccurred())
+
+		_, err = utils.KubectlWithOutput("delete",
+			"policy", case17PolicyReplicatedName,
+			"-n", testNamespace,
+			"--ignore-not-found",
+		)
+		Expect(err).ShouldNot(HaveOccurred())
+	})
+	Describe("Test name + namespace over 63", func() {
+		It("Should the error message is presented", func() {
+			output, err := utils.KubectlWithOutput("apply",
+				"-f", case17PolicyLongYaml,
+				"-n", longNamesapce)
+			Expect(err).Should(HaveOccurred())
+			Expect(output).Should(ContainSubstring(errMsg))
+		})
+		It("Should replicated policy should not be validated", func() {
+			_, err := utils.KubectlWithOutput("apply",
+				"-f", case17PolicyReplicatedYaml,
+				"-n", testNamespace)
+			Expect(err).ShouldNot(HaveOccurred())
+			plr := utils.GetWithTimeout(
+				clientHubDynamic, gvrPlacementRule, case17PolicyReplicatedName+"-plr",
+				testNamespace, true, defaultTimeoutSeconds,
+			)
+			plr.Object["status"] = utils.GeneratePlrStatus("managed1")
+			_, err = clientHubDynamic.Resource(gvrPlacementRule).Namespace(testNamespace).UpdateStatus(
+				context.TODO(), plr, metav1.UpdateOptions{},
+			)
+			Expect(err).ToNot(HaveOccurred())
+
+			By("Replicated policy name and cluster namespace should be over 63 character")
+			Expect(utf8.RuneCountInString("managed1." + testNamespace +
+				"." + case17PolicyReplicatedName)).Should(BeNumerically(">", 63))
+
+			By("Replicated policy should be created")
+			plc := utils.GetWithTimeout(
+				clientHubDynamic, gvrPolicy, testNamespace+"."+case17PolicyReplicatedName,
+				"managed1", true, defaultTimeoutSeconds,
+			)
+			Expect(plc).ToNot(BeNil())
+		})
+	})
+})
--- a/PROJECT
+++ b/PROJECT
@@ -15,6 +15,9 @@ resources:
   group: policy
   kind: Policy
   path: open-cluster-management.io/governance-policy-propagator/api/v1
+  webhooks:
+    validation: true
+    webhookVersion: v1
   version: v1
 - api:
     crdVersion: v1
--- a/deploy/operator.yaml
+++ b/deploy/operator.yaml
@@ -231,16 +231,22 @@ subjects:
 apiVersion: apps/v1
 kind: Deployment
 metadata:
+  labels:
+    webhook-origin: governance-policy-propagator
   name: governance-policy-propagator
 spec:
   replicas: 1
   selector:
     matchLabels:
       name: governance-policy-propagator
+      webhook-origin: governance-policy-propagator
   template:
     metadata:
+      annotations:
+        kubectl.kubernetes.io/default-container: governance-policy-propagator
       labels:
         name: governance-policy-propagator
+        webhook-origin: governance-policy-propagator
     spec:
       containers:
       - args:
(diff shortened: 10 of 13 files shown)


build/common/scripts/setup-webhook.sh
```diff
@@ -0,0 +1,7 @@
+kubectl apply -f deploy/webhook.yaml
+while [ $? -ne 0 ]
+do 
+    sleep 3
+    echo "-----webhook install failed try again-----"
+    kubectl apply -f deploy/webhook.yaml
+done
```
yiraeChristineKim, line 7: What is this mark?

deploy/manager/manager.yaml
```diff
@@ -24,10 +36,14 @@ spec:
             - "--metrics-bind-address=:8383"
             - "--leader-elect"
           ports:
-            - containerPort: 8383
+            - containerPort: 9443
               protocol: TCP
               name: http
```
JustinKuli, line 41: We should probably mark both ports as used, `8383` for metrics, and now `9443` for the webhook.

build/common/scripts/setup-webhook.sh
JustinKuli: Do you mean the red minus in GitHub? That means there's no newline at the end of the file. Technically that means the line is incomplete: https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/V1_chap03.html#tag_03_206, but it's just a nitpick.

Or do you mean, why did kubebuilder add this script (I'm guessing that's where it came from)? I don't think we need a whole script, just try to apply the yaml once and give up in the make target if it fails... I definitely don't like the potentially infinite loop here.

deploy/manager/manager.yaml
```diff
@@ -3,15 +3,27 @@ apiVersion: apps/v1
...
 metadata:
   name: governance-policy-propagator
+  labels: 
+    app.kubernetes.io/component: manager
+    app.kubernetes.io/created-by: webhook
+    app.kubernetes.io/instance: controller-manager
+    app.kubernetes.io/managed-by: kustomize
+    app.kubernetes.io/name: deployment
+    app.kubernetes.io/part-of: webhook
+    control-plane: controller-manager
```
JustinKuli, line 13: Do we need to add any of these labels for the webhook to work? If so, let's make them a bit more specific to just our controller, if possible. Eg `app.kubernetes.io/name: governance-policy-propagator`

test/e2e/case17_policy_webhook_test.go
```diff
@@ -0,0 +1,45 @@
...
+	AfterEach(func() {
+		// cleanup
+		_, err := utils.KubectlWithOutput("delete",
+			"ns", longNamesapce,
+			"--ignore-not-found",
+		)
+		Expect(err).ShouldNot(HaveOccurred())
+	})
+	Describe("Test name + namespace over 63", func() {
+		It("Should the error message is presented", func() {
```
JustinKuli, line 37: Can we add another test for when the error doesn't happen? Since this is the only test running right now when the webhook is enabled, I think?

JustinKuli (changes requested):

test/e2e/case17_policy_webhook_test.go
yiraeChristineKim: I tried to add update case but namespace and name cannot be modified... I tried to add long namespace and short name but namespace itself has limitation so... would you give some idea?!

deploy/manager/manager.yaml
yiraeChristineKim: let me try! last time i removed then didn't work

deploy/manager/manager.yaml
yiraeChristineKim: you mean I need to change `"--metrics-bind-address=:8383"` to 9443?

deploy/manager/manager.yaml
yiraeChristineKim: app.kubernetes.io/name: looks like they use this as a resource kind

yiraeChristineKim: seems like woking well without label

deploy/manager/manager.yaml
yiraeChristineKim: I understand. I added one more item for metrics

api/v1/policy_webhook.go
```diff
@@ -0,0 +1,63 @@
...
+// log is for logging in this package.
+var policylog = logf.Log.WithName("policy-resource")
+
+func (r *Policy) SetupWebhookWithManager(mgr ctrl.Manager) error {
+	return ctrl.NewWebhookManagedBy(mgr).
+		For(r).
+		Complete()
+}
+
+//+kubebuilder:webhook:path=/validate-policy-open-cluster-management-io-v1-policy,mutating=false,failurePolicy=fail,sideEffects=None,groups=policy.open-cluster-management.io,resources=policies,verbs=create;update,versions=v1,name=policy.open-cluster-management.io.webhook,admissionReviewVersions=v1
```
mprahl, line 25: This should only handle creations since you can't update an object name after it's created.

api/v1/policy_webhook.go
```diff
@@ -0,0 +1,63 @@
...
+
+// ValidateCreate implements webhook.Validator so a webhook will be registered for the type
+func (r *Policy) ValidateCreate() (admission.Warnings, error) {
+	policylog.Info("validate create", "name", r.Name)
+
+	return r.validate()
+}
+
+// ValidateUpdate implements webhook.Validator so a webhook will be registered for the type
+func (r *Policy) ValidateUpdate(_ runtime.Object) (admission.Warnings, error) {
```
mprahl, line 37: After addressing my comment, I think this can be removed.

api/v1/policy_webhook.go
```diff
@@ -0,0 +1,63 @@
...
+	policylog.Info("validate update", "name", r.Name)
+
+	return nil, nil
+}
+
+// ValidateDelete implements webhook.Validator so a webhook will be registered for the type
+func (r *Policy) ValidateDelete() (admission.Warnings, error) {
```
mprahl, line 44: This isn't used right?

api/v1/policy_webhook.go
```diff
@@ -0,0 +1,63 @@
...
+
+	return nil, nil
+}
+
+// validate admits a pod if a specific annotation exists.
+func (r *Policy) validate() (admission.Warnings, error) {
+	policylog.Info("Validating Policy by webhook")
+
+	// 1 character for "."
+	if len(r.GetNamespace())+len(r.GetName()) > 62 {
```
mprahl, line 55: I don't think Kubernetes supports non-ASCII characters for names, but just in case, you should be using `utf8.RuneCountInString` as opposed to `len` since I'm fairly certain `len` measures the bytes used by the string and not the character count. In most cases, ASCII characters would be used which only use one byte each.

api/v1/policy_webhook.go
```diff
@@ -0,0 +1,63 @@
...
+		msg := "the combined length of namespace and policy name " +
+			"(namespaceName.policyName) should not exceed 63 characters"
```
mprahl, line 57:
```suggestion
			"(<namespace>.<policy>) should not exceed 63 characters"
```

build/common/scripts/setup-webhook.sh
```diff
@@ -0,0 +1,20 @@
...
+n=0
+try=10
+kubectl apply -f deploy/webhook.yaml
+
+while  [[ $? -ne 0 && $n -ne $try ]]
+do 
+    sleep 5
+    n=$((n+1))
+    echo "-----Webhook install failed, Please try again $n-----"
+    kubectl apply -f deploy/webhook.yaml
```
mprahl, line 11: Why do you need to continuously retry? What causes it to fail?

deploy/webhook.yaml
```diff
@@ -0,0 +1,117 @@
...
+rules:
+- nonResourceURLs:
+  - /metrics
+  verbs:
+  - get
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  name: webhook-proxy-role
```
mprahl, line 14: Is this ClusterRole needed?

deploy/webhook.yaml
```diff
@@ -0,0 +1,117 @@
...
+subjects:
+- kind: ServiceAccount
+  name: webhook-controller-manager
+  namespace: open-cluster-management
+---
+apiVersion: v1
+kind: Service
+metadata:
+  name: webhook-controller-manager-metrics-service
```
mprahl, line 45: I don't think we need metrics to the webhook. Could you please disabled that?

main.go
```diff
@@ -301,6 +301,13 @@ func main() {
 		os.Exit(1)
 	}
 
+	if os.Getenv("ENABLE_WEBHOOKS") != "false" {
```
mprahl, line 304: Why do you need the option to disable the webhook?

README.md
```diff
@@ -70,22 +87,19 @@ make kind-delete-cluster
 
 ### Updating Deployment resources
 
-Some of the deployment resources are generated by kubebuilder - the crds are generated into `./deploy/crds` and the rbac
-details from kubebuilder comments are compiled into `./deploy/rbac/role.yaml`. Other details are managed independently -
-in particular, the details in `./deploy/manager/manager.yaml`. When any of those details need to be changed, the main
-deployment yaml `./deploy/operator.yaml` must be regenerated through the `make generate-operator-yaml` target. The
-`./deploy/operator.yaml` SHOULD NOT be manually updated.
+Some of the deployment resources are generated by kubebuilder - the crds are generated into `./deploy/crds` and the rbac details from kubebuilder comments are compiled into `./deploy/rbac/role.yaml`. Other details are managed independently - in particular, the details in `./deploy/manager/manager.yaml`. When any of those details need to be changed, the main deployment yaml `./deploy/operator.yaml` must be regenerated through the `make generate-operator-yaml` target. The `./deploy/operator.yaml` SHOULD NOT be manually updated.
```
mprahl, line 90: Could you please undo these line length changes?

.github/workflows/kind.yml
```diff
@@ -27,69 +27,76 @@ jobs:
         # The "minimum" tag is set in the Makefile
         # KinD tags: https://hub.docker.com/r/kindest/node/tags
         kind:
-          - 'minimum'
-          - 'latest'
+          - "minimum"
```
mprahl, line 30: Could you please undo these style changes or at least put them in a separate commit?

main.go
mprahl: Oh, it's likely for when it's not run in the cluster.

Could you please create a new flag/argument instead of an environment variable for this?

api/v1/policy_webhook.go
mprahl, line 25: Do we actually want `failurePolicy=fail`? If the webhook is down, it would allow the invalid policy to be created but the Propagator eventually reports a failure so I think I'd rather be conservative here to not cause an outage.

build/common/scripts/setup-webhook.sh
yiraeChristineKim: Deploying webhook resources after cert-manager causes error. I guess cert-manager needs time to settle down. Instead sleep 10, multiple try is better way

deploy/webhook.yaml
yiraeChristineKim: The proxy role is needed. Without this not working

deploy/webhook.yaml
yiraeChristineKim: let me try

api/v1/policy_webhook.go
yiraeChristineKim:
```
type Validator interface {
	runtime.Object

	// ValidateCreate validates the object on creation.
	// The optional warnings will be added to the response as warning messages.
	// Return an error if the object is invalid.
	ValidateCreate() (warnings Warnings, err error)

	// ValidateUpdate validates the object on update. The oldObj is the object before the update.
	// The optional warnings will be added to the response as warning messages.
	// Return an error if the object is invalid.
	ValidateUpdate(old runtime.Object) (warnings Warnings, err error)

	// ValidateDelete validates the object on deletion.
	// The optional warnings will be added to the response as warning messages.
	// Return an error if the object is invalid.
	ValidateDelete() (warnings Warnings, err error)
}
```

This file is read only

yiraeChristineKim:
> 

I tried this but there is no way to get `namespace`

yiraeChristineKim:
> @yiraeChristineKim, rather than a webhook, could you please try using CEL to see if that would work? https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definitions/#validation-rules
> 
> It would only have effect on newer Kubernetes/OpenShift versions but I think it might be a reasonable limitation for now.
@mprahl 
I tried this but there is no way to get `namespace`

deploy/webhook.yaml
mprahl: Where is the proxy used?

api/v1/policy_webhook.go
mprahl: I'm saying that I would prefer the policy to still be created if the webhook is down since nothing damaging actually happens, it's just you get an error once the policy propagator has processed the policy.

api/v1/policy_webhook.go
```diff
@@ -0,0 +1,65 @@
...
+	policylog.Info("validate delete", "name", r.Name)
+
+	return nil, nil
+}
+
+// validate admits a pod if a specific annotation exists.
+func (r *Policy) validate() (admission.Warnings, error) {
+	policylog.Info("Validating Policy by webhook")
+
+	namespaceLen := len([]rune(r.GetNamespace()))
```
mprahl, line 54: @yiraeChristineKim you need to ensure this only runs on root policies, otherwise, this check will also erroneously apply to replicated policies.

For example, if my root policy is in the namespace `open-cluster-management-global-set` and the policy name is `policy-name`, the webhook would validate the length of `open-cluster-management-global-set.policy-name`.

Then say the policy applies to a cluster named `very-long-cluster-name-for-this-example`, then the webhook would erroneously check that the length of `very-long-cluster-name-for-this-example.open-cluster-management-global-set.policy-name`.

yiraeChristineKim: Now I fully understand!

.github/workflows/kind.yml
```diff
@@ -81,8 +81,13 @@ jobs:
 
     - name: Verify Deployment Configuration
       run: |
+        make webhook
```
mprahl, line 84: Should this be after `make kind-deploy-controller-dev` since the webhook isn't running until then?

.github/workflows/kind.yml
```diff
@@ -92,4 +97,4 @@ jobs:
     - name: Clean up cluster
       if: ${{ always() }}
       run: |
-        make kind-delete-cluster
+        make kind-delete-cluster
```
mprahl, line 100: Could you please add back the new line?

README.md
```diff
@@ -62,6 +62,15 @@ make e2e-dependencies
 make e2e-test
 ```
 
+### How to run webhook locally
+
+```bash
+--enable-webhooks=true
+```
+
+> **Limit**: If you want to run the webhook locally, you need to generate certifications and place them `/tmp/k8s-webhook-server/serving-certs/tls.{crt,key}`. If you’re not running a local API server, you’ll also need to figure out how to proxy traffic from the remote cluster to your local webhook server. For this reason, Kubebuilder generally recommends disabling webhooks when doing your local code-run-test cycle `export ENABLE_WEBHOOKS=false`
```
mprahl, line 71:
```suggestion
> **Limit**: If you want to run the webhook locally, you need to generate certifications and place them in `/tmp/k8s-webhook-server/serving-certs/tls.{crt,key}`. If you’re not running a local API server, you’ll also need to figure out how to proxy traffic from the remote cluster to your local webhook server. For this reason, Kubebuilder generally recommends disabling webhooks when doing your local code-run-test cycle. To disable it, please supply the  `--enable-webhooks=false` argument when running the controller.
```

api/v1/policy_webhook.go
```diff
@@ -0,0 +1,72 @@
...
+
+// ValidateDelete implements webhook.Validator so a webhook will be registered for the type
+func (r *Policy) ValidateDelete() (admission.Warnings, error) {
+	policylog.Info("validate delete", "name", r.Name)
+
+	return nil, nil
+}
+
+// validate admits a pod if a specific annotation exists.
+func (r *Policy) validate() (admission.Warnings, error) {
```
mprahl, line 51:
```suggestion
func (r *Policy) validateName() (admission.Warnings, error) {
```

api/v1/policy_webhook.go
```diff
@@ -0,0 +1,72 @@
...
+}
+
+// ValidateDelete implements webhook.Validator so a webhook will be registered for the type
+func (r *Policy) ValidateDelete() (admission.Warnings, error) {
+	policylog.Info("validate delete", "name", r.Name)
+
+	return nil, nil
+}
+
+// validate admits a pod if a specific annotation exists.
```
mprahl, line 50: Could you please update this comment?

api/v1/policy_webhook.go
```diff
@@ -0,0 +1,72 @@
...
+func (r *Policy) validate() (admission.Warnings, error) {
+	policylog.Info("Validating Policy by webhook")
```
mprahl, line 52:
```suggestion
	policylog.Info("Validating the policy name through a validating webhook")
```

api/v1/policy_webhook.go
```diff
@@ -0,0 +1,72 @@
...
+
+	// replicated policy should not pass this validation
+	labels := r.GetLabels()
+
+	if _, ok := labels["policy.open-cluster-management.io/root-policy"]; ok {
```
mprahl, line 57: Optional: It's not needed to store a copy of the labels in memory.

```suggestion
	if _, ok := r.GetLabels()["policy.open-cluster-management.io/root-policy"]; ok {
```

api/v1/policy_webhook.go
```diff
@@ -0,0 +1,72 @@
...
+		return nil, nil
+	}
+
+	namespaceLen := len([]rune(r.GetNamespace()))
+	nameLenLen := len([]rune(r.GetName()))
```
mprahl, line 62: Could you please use https://pkg.go.dev/unicode/utf8#RuneCountInString? Then we don't have to create rune slices.

api/v1/policy_webhook.go
```diff
@@ -0,0 +1,72 @@
...
+	// 1 character for "."
+	if (namespaceLen + nameLenLen) > 62 {
+		msg := "the combined length of the policy namespace and name " +
+			"<namespace>.<name> should not exceed 63 characters"
```
mprahl, line 66:
```suggestion
		msg := "the combined length of the policy namespace and name " +
			"(<namespace>.<name>) should not exceed 63 characters"
```

api/v1/policy_webhook.go
```diff
@@ -0,0 +1,72 @@
...
+
+	if _, ok := labels["policy.open-cluster-management.io/root-policy"]; ok {
+		return nil, nil
+	}
+
+	namespaceLen := len([]rune(r.GetNamespace()))
+	nameLenLen := len([]rune(r.GetName()))
+	// 1 character for "."
+	if (namespaceLen + nameLenLen) > 62 {
+		msg := "the combined length of the policy namespace and name " +
```
mprahl, line 65: Could you define this as a package level error? This allows `errors.Is` and `errors.As` functions to be used on it by the caller and it also means you don't have to allocate memory for this string and error object below every time the webhook is called.

deploy/manager/manager.yaml
mprahl: What's the outcome of this discussion?

deploy/manager/manager.yaml
```diff
@@ -2,16 +2,22 @@
 apiVersion: apps/v1
 kind: Deployment
 metadata:
+  labels:
+    control-plane: controller-manager
```
mprahl, line 6: Is this new label required?

deploy/webhook.yaml
mprahl: This question still applies. Where is the validating webhook checking user access?

test/e2e/case17_policy_webhook_test.go
```diff
@@ -0,0 +1,96 @@
...
+		Expect(err).ShouldNot(HaveOccurred())
+
+		_, err = utils.KubectlWithOutput("create",
+			"ns", longRepliNamesapce,
+		)
+		Expect(err).ShouldNot(HaveOccurred())
+
+		_, err = utils.KubectlWithOutput("apply",
+			"-f", "../resources/case17_policy_webhook/"+
+				"long-name-cluster.yaml",
```
mprahl, line 42: The cluster name does not impact the validating webhook since the replicated policy is not evaluated, only the root policy.

.github/workflows/kind.yml
```diff
@@ -83,6 +83,10 @@ jobs:
       run: |
         make build-images
         make kind-deploy-controller-dev
```
yiraeChristineKim, line 85:
```
kind-bootstrap-cluster: kind-create-cluster install-crds webhook kind-deploy-controller install-resources 
kind-bootstrap-cluster-dev: kind-create-cluster install-crds install-resources webhook
```

I added `webhook` before propagator controller deployed because our default is `--enable-webhooks=true
`, controller is looking for cert-manager when it initiates. If there isn't `/tmp/k8s-webhook-server/serving-certs/tls.{crt,key}`, image creating will be looping (fail)

deploy/manager/manager.yaml
yiraeChristineKim: webhook service need this label 
```
apiVersion: v1
kind: Service
metadata:
  name: webhook-webhook-service
  namespace: open-cluster-management
spec:
  ports:
  - port: 443
    protocol: TCP
    targetPort: 9443
  selector:
    control-plane: controller-manager

```

deploy/webhook.yaml
yiraeChristineKim: Kubebuilder name it as webhook-proxy-role. I tested removed cluster-role. Webhook then result is fail with 
```
 Internal error occurred: failed calling webhook "policy.open-cluster-management.io.webhook": failed to call webhook: Post "https://webhook-webhook-service.open-cluster-management.svc:443/validate-policy-open-cluster-management-io-v1-policy?timeout=10s": dial tcp 10.96.119.57:443: connect: connection refused
```

Hard to find a doc that explains why kubebuilder made this.
I will look for this more

.github/workflows/kind.yml
yiraeChristineKim: I added webhook before propagator controller deployed because our default is --enable-webhooks=true , controller is looking for cert-manager when it initiates. If there isn't /tmp/k8s-webhook-server/serving-certs/tls.{crt,key}, image creating will be looping (fail)

api/v1/policy_webhook.go
yiraeChristineKim: awesome! Thanks!

api/v1/policy_webhook.go
```diff
@@ -0,0 +1,70 @@
...
+		Complete()
+}
+
+//+kubebuilder:webhook:path=/validate-policy-open-cluster-management-io-v1-policy,mutating=false,failurePolicy=fail,sideEffects=None,groups=policy.open-cluster-management.io,resources=policies,verbs=create,versions=v1,name=policy.open-cluster-management.io.webhook,admissionReviewVersions=v1
+
+var _ webhook.Validator = &Policy{}
+
+// ValidateCreate implements webhook.Validator so a webhook will be registered for the type
+func (r *Policy) ValidateCreate() (admission.Warnings, error) {
+	policylog.Info("validate create", "name", r.Name)
```
mprahl, line 36:
```suggestion
	policylog.Info("Validate policy creation request", "name", r.Name)
```

api/v1/policy_webhook.go
```diff
@@ -0,0 +1,70 @@
...
+
+	return r.validateName()
+}
+
+// ValidateUpdate implements webhook.Validator so a webhook will be registered for the type
+func (r *Policy) ValidateUpdate(_ runtime.Object) (admission.Warnings, error) {
+	policylog.Info("validate update", "name", r.Name)
+
```
mprahl, line 44:
```suggestion
```

api/v1/policy_webhook.go
```diff
@@ -0,0 +1,70 @@
...
+	return nil, nil
+}
+
+// ValidateDelete implements webhook.Validator so a webhook will be registered for the type
+func (r *Policy) ValidateDelete() (admission.Warnings, error) {
+	policylog.Info("validate delete", "name", r.Name)
+
```
mprahl, line 51:
```suggestion
```

api/v1/policy_webhook.go
```diff
@@ -0,0 +1,70 @@
...
+	return nil, nil
+}
+
+// validate the policy name and namespace length
+func (r *Policy) validateName() (admission.Warnings, error) {
+	policylog.Info("Validating the policy name through a validating webhook")
+
+	// replicated policy should not pass this validation
```
mprahl, line 59:
```suggestion
	// replicated policies don't need this validation
```

deploy/webhook.yaml
```diff
@@ -0,0 +1,61 @@
+apiVersion: v1
+kind: Service
+metadata:
+  name: webhook-webhook-service
```
mprahl, line 4: Can this name be made unique since there are other controllers running the same namespace? We don't want conflicts.

deploy/webhook.yaml
```diff
@@ -0,0 +1,61 @@
...
+  - port: 443
+    protocol: TCP
+    targetPort: 9443
+  selector:
+    control-plane: controller-manager
+---
+apiVersion: cert-manager.io/v1
+kind: Certificate
+metadata:
+  name: webhook-serving-cert
```
mprahl, line 17: Can this name be made unique since there are other controllers running the same namespace? We don't want conflicts.

deploy/webhook.yaml
```diff
@@ -0,0 +1,61 @@
...
+  - webhook-webhook-service.open-cluster-management.svc.cluster.local
+  issuerRef:
+    kind: Issuer
+    name: webhook-selfsigned-issuer
+  secretName: webhook-server-cert
+---
+apiVersion: cert-manager.io/v1
+kind: Issuer
+metadata:
+  name: webhook-selfsigned-issuer
```
mprahl, line 31: Can this name be made unique since there are other controllers running the same namespace? We don't want conflicts.

deploy/webhook.yaml
```diff
@@ -0,0 +1,61 @@
...
+  namespace: open-cluster-management
+spec:
+  selfSigned: {}
+---
+apiVersion: admissionregistration.k8s.io/v1
+kind: ValidatingWebhookConfiguration
+metadata:
+  annotations:
+    cert-manager.io/inject-ca-from: open-cluster-management/webhook-serving-cert
+  name: webhook-validating-webhook-configuration
```
mprahl, line 41: Can this name be made unique since there are other controllers running the same namespace? We don't want conflicts.

deploy/webhook.yaml
```diff
@@ -0,0 +1,61 @@
...
+webhooks:
+- admissionReviewVersions:
+  - v1
+  clientConfig:
+    service:
+      name: webhook-webhook-service
+      namespace: open-cluster-management
+      path: /validate-policy-open-cluster-management-io-v1-policy
+  failurePolicy: Fail
```
mprahl, line 50: Like we had talked about, I think it's better to just ignore the failure for now since the Propagator will eventually catch the issue. This webhook is just providing immediate feedback for a better user experience but we don't want it to block policy creation if it's down.

main.go
```diff
@@ -90,12 +90,16 @@ func main() {
...
+	var enableWebhooks bool
 
 	pflag.StringVar(&metricsAddr, "metrics-bind-address", ":8383", "The address the metric endpoint binds to.")
 	pflag.StringVar(&probeAddr, "health-probe-bind-address", ":8081", "The address the probe endpoint binds to.")
 	pflag.BoolVar(&enableLeaderElection, "leader-elect", true,
 		"Enable leader election for controller manager. "+
 			"Enabling this will ensure there is only one active controller manager.")
+	pflag.BoolVar(&enableWebhooks, "enable-webhooks", true,
+		"Enable webhook for validating namespace + name "+
+			"less than 63 characters")
```
mprahl, line 102: Optional: Can the string fit on a single line?

test/e2e/case17_policy_webhook_test.go
mprahl: So can you just use the normal managed cluster that is part of the E2E tests instead of this long-name-cluster?

yiraeChristineKim: you mean like 'managed1' ?

mprahl: Yes, that sounds good.

deploy/manager/manager.yaml
yiraeChristineKim: I added for metrics but we don't need metrics so I removed it
```
            - containerPort: 8383
              protocol: TCP
              name: http
```

deploy/webhook.yaml
yiraeChristineKim: How about propagator-webhook-service

deploy/webhook.yaml
yiraeChristineKim: How about `webhook-propagator-serving-cert`?

mprahl: If `governance-policy-propagator-` prefixes fit, then it'd be good to be consistent. Otherwise, prefixing all with `propagator-` is good.

yiraeChristineKim: `governance-policy-propagator-` too long, let's go with  `propagator-`

api/v1/policy_webhook.go
```diff
@@ -0,0 +1,70 @@
...
+	if _, ok := r.GetLabels()["policy.open-cluster-management.io/root-policy"]; ok {
+		return nil, nil
+	}
+
+	// 1 character for "."
+	if (utf8.RuneCountInString(r.GetNamespace()) + utf8.RuneCountInString(r.GetName())) > 62 {
```
mprahl, line 65: Are you able to do this to avoid memory allocation on those function calls?
```suggestion
	if (utf8.RuneCountInString(r.Namespace) + utf8.RuneCountInString(r.Name)) > 62 {
```

deploy/webhook.yaml
yiraeChristineKim: Ok I will change it to `Ignore`

test/e2e/case17_policy_webhook_test.go
yiraeChristineKim: `case17PolicyReplicatedYaml` has name: `case17-test-policy-replicated`, namepace:`  policy-propagator-test`. I tested this because I wanted to see that the policy is created even though it is copied in long cluster name. name + namespace(long cluster name) > 63

api/v1/policy_webhook.go
yiraeChristineKim: let me try

Makefile
```diff
@@ -281,7 +291,7 @@ e2e-dependencies:
 
 .PHONY: e2e-test
 e2e-test: e2e-dependencies
-	$(GINKGO) -v --fail-fast $(E2E_TEST_ARGS) test/e2e
+	$(GINKGO) -v --fail-fast $(E2E_TEST_ARGS) --label-filter="!webhook" test/e2e
```
mprahl, line 294:
```suggestion
	$(GINKGO) -v --fail-fast $(E2E_TEST_ARGS) --label-filter="!webhook" test/e2e
```

api/v1/policy_webhook.go
```diff
@@ -0,0 +1,66 @@
...
+	"k8s.io/apimachinery/pkg/runtime"
+	ctrl "sigs.k8s.io/controller-runtime"
+	logf "sigs.k8s.io/controller-runtime/pkg/log"
+	"sigs.k8s.io/controller-runtime/pkg/webhook"
+	"sigs.k8s.io/controller-runtime/pkg/webhook/admission"
+)
+
+var (
+	// log is for logging in this package.
+	policylog = logf.Log.WithName("policy-resource")
```
mprahl, line 19:
```suggestion
	policylog = logf.Log.WithName("policy-validating-webhook")
```

deploy/manager/manager.yaml
mprahl: @yiraeChristineKim so we don't want metrics from the validating webhook but the Prometheus metrics exposed by the Policy Propagator we still want. So as Justin suggested, we'll need two ports. One for the Policy Propagator metrics of `8383` and one for the webhook of `9443`.

deploy/webhook.yaml
```diff
@@ -0,0 +1,61 @@
+apiVersion: v1
```
mprahl, line 1: Note that when this is added to ACM, you'll want to use OpenShift's mechanism of generating certificates instead of Cert Manager:
https://docs.openshift.com/container-platform/4.13/security/certificates/service-serving-certificate.html#add-service-certificate_service-serving-certificate
https://docs.openshift.com/container-platform/4.13/security/certificates/service-serving-certificate.html#add-service-certificate-validating-webhook_service-serving-certificate

No action is needed in this PR though.

main.go
```diff
@@ -90,12 +90,15 @@ func main() {
...
 	var keyRotationDays, keyRotationMaxConcurrency, policyMetricsMaxConcurrency, policyStatusMaxConcurrency uint
+	var enableWebhooks bool
 
 	pflag.StringVar(&metricsAddr, "metrics-bind-address", ":8383", "The address the metric endpoint binds to.")
 	pflag.StringVar(&probeAddr, "health-probe-bind-address", ":8081", "The address the probe endpoint binds to.")
 	pflag.BoolVar(&enableLeaderElection, "leader-elect", true,
 		"Enable leader election for controller manager. "+
 			"Enabling this will ensure there is only one active controller manager.")
+	pflag.BoolVar(&enableWebhooks, "enable-webhooks", true,
+		"Enable webhook for validating namespace + name less than 63 characters")
```
mprahl, line 101:
```suggestion
		"Enable the policy validating webhook")
```

test/e2e/case17_policy_webhook_test.go
```diff
@@ -0,0 +1,85 @@
...
+			)
+			plr.Object["status"] = utils.GeneratePlrStatus("managed1")
+			_, err = clientHubDynamic.Resource(gvrPlacementRule).Namespace(testNamespace).UpdateStatus(
+				context.TODO(), plr, metav1.UpdateOptions{},
+			)
+			Expect(err).ToNot(HaveOccurred())
+
+			By("Replicated policy name and cluster namespace should be over 63 character")
+			Expect(utf8.RuneCountInString(testNamespace+"."+case17PolicyReplicatedName) +
+				utf8.RuneCountInString("managed1")).Should(BeNumerically(">", 63))
```
mprahl, line 75:
```suggestion
			Expect(utf8.RuneCountInString("managed1."+testNamespace+"."+case17PolicyReplicatedName)).Should(BeNumerically(">", 63))
```

mprahl (commented):
This is looking really good! Just a few minor comments.

mprahl (approved):

mprahl:
@JustinKuli did you want to take another look at this?

deploy/webhook.yaml
```diff
@@ -0,0 +1,61 @@
...
+kind: Service
+metadata:
+  name: propagator-webhook-service
+  namespace: open-cluster-management
+spec:
+  ports:
+  - port: 443
+    protocol: TCP
+    targetPort: 9443
```
mprahl, line 10: Do you want to use the name (i.e. `webhook-http`) here instead of the port number?

yiraeChristineKim: I think `targetport` is enough

mprahl (approved):
/hold in case @JustinKuli wants to review it again

JustinKuli (approved):
LGTM

PR #123, opened 2023-07-25
Add webhook for limit 64 char and framework test fix

yiraeChristineKim:
The 63 character limit validation for policy name is not working while creating a Policy via CLI.
Policy creation - Policy name + namespace name can’t go longer than 63 From CLI, it possible to create a policy with a name like “policy-check-setupjob-setupjob-operator-installer” with no issue; but, while wanting to edit it, we can’t.
Getting this error: “The combined length of namespace and policy nmabe (namespaceName.policyName) should not exceed 63 characters (see attached screenshot)

- Extend timeout 
- Add `make webhook` for framework tests 

Ref: https://issues.redhat.com/browse/ACM-3558

Signed-off-by: Yi Rae Kim [yikim@redhat.com](mailto:yikim@redhat.com)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -223,13 +223,19 @@ webhook:
 	@echo installing cert-manager
 	kubectl apply -f https://github.com/cert-manager/cert-manager/releases/download/v1.12.0/cert-manager.yaml
 	@echo "wait until pods are up"
-	kubectl wait deployment -n cert-manager cert-manager --for condition=Available=True --timeout=90s
-	kubectl wait --for=condition=Ready pod -l app.kubernetes.io/instance=cert-manager -n cert-manager --timeout=30s 
+	kubectl wait deployment -n cert-manager cert-manager --for condition=Available=True --timeout=180s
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -223,13 +223,19 @@ webhook:
 	@echo installing cert-manager
 	kubectl apply -f https://github.com/cert-manager/cert-manager/releases/download/v1.12.0/cert-manager.yaml
 	@echo "wait until pods are up"
-	kubectl wait deployment -n cert-manager cert-manager --for condition=Available=True --timeout=90s
-	kubectl wait --for=condition=Ready pod -l app.kubernetes.io/instance=cert-manager -n cert-manager --timeout=30s 
+	-kubectl wait deployment -n cert-manager cert-manager --for condition=Available=True --timeout=180s
+	-kubectl wait --for=condition=Ready pod -l app.kubernetes.io/instance=cert-manager -n cert-manager --timeout=180s
```
JustinKuli, line 227: Why ignore the possible errors from these lines?

yiraeChristineKim: I did because of mininum, false case but I think it is better to remove that Thanks!

JustinKuli (approved):

PR #126, opened 2023-08-31
Add the database schema for the compliance events API

mprahl:
This defines the database schema, integrates with a database migration library, and automatically migrates the database when the governance-policy-database secret is defined in the controller namespace.

This feature is disabled by default but can be enabled with the -enable-compliance-events-store flag and by defining the WATCH_NAMESPACE_COMPLIANCE_EVENTS_STORE environment variable.

Relates:
https://issues.redhat.com/browse/ACM-6848


Makefile
```diff
@@ -218,13 +221,33 @@ kind-bootstrap-cluster: kind-create-cluster install-crds webhook kind-deploy-con
...
 	kubectl wait deployment -n cert-manager cert-manager --for condition=Available=True --timeout=180s
 	kubectl wait --for=condition=Ready pod -l app.kubernetes.io/instance=cert-manager -n cert-manager --timeout=180s 
+
+postgres: cert-manager
+	@echo "Installing Postgres"
+	-kubectl create ns $(KIND_NAMESPACE)
+	sed 's/open-cluster-management/$(KIND_NAMESPACE)/g' build/kind/postgres.yaml | kubectl apply --timeout=180s -f-
+
+	@echo "Waiting until the pods are up"
+	@sleep 3
```
yiraeChristineKim, line 237: Why we need this? we have `wait` command

mprahl: It's a hack in case you run `make postgres` again with some changes in the YAML. If you go too fast to the `kubectl wait` command, the Deployment hasn't started a new rollout yet so the `kubectl wait` command uses the old pod.

JustinKuli (commented):
Only small comments. I don't really know much about the SQL, but for what it's worth it doesn't look wrong to me 😆

build/kind/postgres.yaml
```diff
@@ -0,0 +1,98 @@
...
+              subPath: data
+            - mountPath: /var/lib/postgresql/tls
+              name: postgres-cert
+              readOnly: true
+      volumes:
+        - name: postgres-cert
+          secret:
+            secretName: postgres-cert
+            # Postgres requires limited permissions on the private key.
+            defaultMode: 0o440
```
yiraeChristineKim, line 95: Is this correct? unusual too me

controllers/complianceeventsapi/migrations/000001_compliance_history_initial_tables.up.sql
```diff
@@ -0,0 +1,57 @@
...
+   spec_hash CHAR[40],
+   CONSTRAINT fk_parent_policy_id
+      FOREIGN KEY(parent_policy_id) 
+	  REFERENCES parent_policies(id),
+   UNIQUE (kind, api_group, name, parent_policy_id, spec_hash)
+);
+
+CREATE INDEX IF NOT EXISTS idx_policies_spec_hash ON policies (spec_hash);
+
+CREATE TABLE IF NOT EXISTS compliance_events(
```
gparvin, line 38: I'm wondering about compliance operator which could have a standard, category control for the overall policy, but compliance events may be for more specific categories and controls due to the specific rule failure.  Just wondering if we may want to account for a scenario like that.

controllers/complianceeventsapi/migrations/000001_compliance_history_initial_tables.up.sql
```diff
@@ -0,0 +1,57 @@
+BEGIN;
+
+CREATE TABLE IF NOT EXISTS clusters(
+   id serial PRIMARY KEY,
+   name TEXT NOT NULL,
```
gparvin, line 5: It may be in the design and I don't remember, are we using the namespaced name so that's why there's no namespace here?

controllers/complianceeventsapi/migrations/000001_compliance_history_initial_tables.up.sql
```diff
@@ -0,0 +1,57 @@
...
+	  REFERENCES policies(id),
+   CONSTRAINT fk_cluster_id
+      FOREIGN KEY(cluster_id) 
+	  REFERENCES clusters(id)
+);
+
+CREATE INDEX IF NOT EXISTS idx_compliance_events_compliance ON compliance_events (compliance);
+CREATE INDEX IF NOT EXISTS idx_compliance_events_timestamp ON compliance_events (timestamp);
+
+COMMIT;
```
gparvin, line 57: Some other questions:
- how do I tell if a policy was deleted or disabled from the data above?
- can I tell if the policy stayed the same but the source changed?
- should we log the policy remediation so changes there are reflected?
- I expect we could expand to include PolicyAutomation that is triggered in the future -- plus any other integrations we may support in the future...
- I think it's good for now to not focus much on policy placement for this historical data, but I wonder just like with the policy being deleted or disabled if we should integrate in "some" form of placement history.  With placement becoming more complex, this can be discussed later.

build/kind/postgres.yaml
mprahl: I believe I had a YAML lint error when I didn't use the octal notation. This is the equivalent of `chmod 440` on the files though.

controllers/complianceeventsapi/migrations/000001_compliance_history_initial_tables.up.sql
```diff
@@ -0,0 +1,57 @@
...
+   UNIQUE (name, categories, controls, standards)
+);
+
+CREATE TABLE IF NOT EXISTS policies(
+   id serial PRIMARY KEY,
+   kind TEXT NOT NULL,
+   api_group TEXT NOT NULL,
+   name TEXT NOT NULL,
+   parent_policy_id INT NOT NULL,
+   spec TEXT,
```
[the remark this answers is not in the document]
mprahl: It's a good idea, though the latter query can be covered by querying for compliance events for a policy of kind `ConfigurationPolicy` but it's not as clear. I'll add this to the `compliance_events` table.

controllers/complianceeventsapi/migrations/000001_compliance_history_initial_tables.up.sql
mprahl: I think you meant to post this on the `parent_policies` table. Yes, we are using the `<namespace>.<name>` format for the name.

I think it makes sense to have a nullable `namespace` column on the `policies` table though to later support integrations with namespaced policies such as Kyverno.

controllers/complianceeventsapi/migrations/000001_compliance_history_initial_tables.up.sql
mprahl: 1. I added this to the design in this [PR|https://github.com/open-cluster-management-io/enhancements/pull/101]. It'll be a separate compliance event from the Spec Sync controller.
2. Could you please rephrase this question?
3. Perhaps, but at the moment I'm thinking not for PolicyAutomation because we are recording the policy state and if the PolicyAutomation changes this state, the compliance event will be recorded through the lens of the policy.
4. My solution for the disabled case is the Spec Sync controller would send an additional compliance event where the `compliance` is set to `deleted`. This does not track the placement but tracks the actual state on the cluster which is better because a cluster could not receive its placement update until much later if the cluster is disconnected from the hub. Having this separate compliance in the `compliance_events` table can give you a sense of when a policy was deleted and then reapplied.

yiraeChristineKim (approved):

controllers/complianceeventsapi/migrations/000001_compliance_history_initial_tables.up.sql
gparvin: I don't remember what I was thinking with the comment:
- can I tell if the policy stayed the same but the source changed?

Regarding remediation:  
- should we log the policy remediation so changes there are reflected? I think I'm ok ignoring this comment for now too.

gparvin (commented):
Thanks for addressing my comments.  It looks like they are all addressed except a couple comments I made that I don't think are important or I don't remember my concern anymore.

PR #128, opened 2023-09-14
7332 placement change reconciler

JustinKuli:
This should help performance at scale, especially if any placements are using selective enforcement, because reconciles coming from placement changes will not re-create all replicated policies: they will still create new replicated policies, and delete ones which are no longer covered by a placement, but they will only consider the remediationAction on existing replicated policies.


controllers/propagator/placement_change_reconclier.go
```diff
@@ -0,0 +1,190 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/event"
+	"sigs.k8s.io/controller-runtime/pkg/handler"
+	"sigs.k8s.io/controller-runtime/pkg/predicate"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+
+	policiesv1 "open-cluster-management.io/governance-policy-propagator/api/v1"
+	"open-cluster-management.io/governance-policy-propagator/controllers/common"
+)
+
+//+kubebuilder:rbac:groups=policy.open-cluster-management.io,resources=placementbindings,verbs=get;list;watch;create;update;patch;delete
```
mprahl, line 22: I think you'll need to add placement decision here

mprahl, line 22: You'll also need access to policies in the kubebuilder marker too.

controllers/propagator/placement_change_reconclier.go
```diff
@@ -0,0 +1,190 @@
...
+
+	// placementRuleMapper maps from a PlacementRule to the PlacementBindings associated with it
+	placementRuleMapper := func(ctx context.Context, object client.Object) []reconcile.Request {
+		log := log.WithValues("placementRuleName", object.GetName(), "namespace", object.GetNamespace())
+		log.V(2).Info("Reconcile Request for PlacementRule")
+
+		pbList := &policiesv1.PlacementBindingList{}
+		lopts := &client.ListOptions{Namespace: object.GetNamespace()}
+
+		opts := client.MatchingFields{"placementRef.name": object.GetName()}
```
mprahl, line 69: Are you able to also filter by PlacementRef.Kind == "PlacementRule"?

controllers/propagator/placement_change_reconclier.go
```diff
@@ -0,0 +1,190 @@
...
+		if err := r.List(ctx, pbList, lopts); err != nil {
+			return nil
+		}
+
+		var result []reconcile.Request
+		// loop through pbs find the matching ones
+		for _, pb := range pbList.Items {
+			match := pb.PlacementRef.APIGroup == appsv1.SchemeGroupVersion.Group &&
+				pb.PlacementRef.Kind == "PlacementRule" &&
+				pb.PlacementRef.Name == object.GetName()
```
mprahl, line 81: No harm in having this but doesn't the list option from before ensure this?

controllers/propagator/placement_change_reconclier.go
```diff
@@ -0,0 +1,190 @@
...
+		// get the Placement name from the PlacementDecision
+		placementName := object.GetLabels()["cluster.open-cluster-management.io/placement"]
+		if placementName == "" {
+			return nil
+		}
+
+		pbList := &policiesv1.PlacementBindingList{}
+		lopts := &client.ListOptions{Namespace: object.GetNamespace()}
+
+		opts := client.MatchingFields{"placementRef.name": placementName}
```
mprahl, line 105: Similar comment to before. Are you able to filter on `placementRef.kind` as well?

controllers/propagator/placement_change_reconclier.go
JustinKuli: I debated not splitting the kubebuilder markers at all - these all get considered together when making the ClusterRole, so it doesn't really make a difference where they are. Or maybe I should split it more, so that the propagator has its required permissions separately listed?

If I split [this marker](https://github.com/open-cluster-management-io/governance-policy-propagator/blob/4178f22748c05eafb42a4b2ed2fe72391a319d70/controllers/propagator/policy_controller.go#L38C1-L39C1) so that the placement-change-reconciler gets the `placement` and `placementdecision` parts, then the generated ClusterRole will have separate entries for those and `managedclusters`. I was trying to skip the work of updating the role in charts and things.

controllers/propagator/propagation.go
```diff
@@ -555,8 +560,13 @@ func (r *Propagator) handleDecisions(
 
 			processedResults++
 
-			for refObject := range result.TemplateRefObjs {
-				allTemplateRefObjs[refObject] = true
+			if checkFullSpec {
+				// NOTE: new replicated policies may still appear when we aren't checking the full specs,
+				// which *could* bring new template references... will that be resolved by an individual
+				// replicated policy reconciler with a watch source for those?
```
mprahl, line 566: Yeah, this is a problem. The dependency watch library needs the ability to append watches.

I think you are on the right track with the root policy handler. This would just be for definition updates of the root policy. The issue is that an update to a replicated policy, even in this new controller, actually triggers a reconcile on the root policy handler from what I can tell from the code. When we added the root policy status controller, this was a point of debate of if we should ignore changes to the replicated policies and we decided not to in the event a user accidentally or maliciously modified a replicated policy. This is a detail I had forgotten about that a reconcile on the root policy that updates a replicated policy actually triggers two total reconciles. That's another inefficiency to address.

Instead of splitting out the placement updates to be a reconciler on the root policy, I think it should be a reconcile on a replicated policy/cluster decision. The following would cause reconciles:

- The placement/placement rule/placement decision binding is created, updated, or deleted. A reconcile request is made for every new/deleted cluster decisions. This could be based on existing propagated policies in the cache to see what was there previously. Alternatively, you could cache that the cluster decisions per root policy in memory.
- The root policy reconciler still handles root policy definition changes, but it doesn't actually perform the work of building replicated policies. We would add concurrency to the new propagated policy reconciler and remove the custom concurrency of handling decisions in the root policy reconciler. The root policy reconciler would have an event source (i.e. Go channel) that feeds reconcile requests to the new propagated policy reconciler. This root policy reconciler would still add filler status for new replicated policies to be created to not cause the root policy status controller to update the root policy status for every new replicated policy being created by the new replicated policy reconciler. Perhaps there is a more clever way here.
- A change to a replicated policy but we need to find a way to distinguish between a user change and a controller change. Perhaps the managed fields could allow us not to duplicate the request if the last update was from this controller.

The following additional changes would be needed:
- All custom locking is removed since deduplication of requests is already handled by controller-runtime's queue.
- The hub template watches would be per replicated policy and trigger a reconcile on the replicated policy. This would solve one of our other inefficiencies.
- You could optionally optimize for the case where selective policy enforcement is used to not reexecute templates but I think that is not needed.

controllers/propagator/placement_change_reconclier.go
JustinKuli: I was half just following what was already in one of the mappers, and bringing it into the others. But based on this doc comment, I think it should only use one filter: https://github.com/kubernetes-sigs/controller-runtime/blob/116a1b831fffe7ccc3c8145306c3e1a3b1b14ffa/pkg/client/options.go#L422

controllers/propagator/placement_change_reconclier.go
JustinKuli: Yes, but because I wasn't 100% sure if the list from the cache would honor that filter, I wanted to keep it here.

controllers/propagator/propagation.go
JustinKuli: I think you're right: a per-replicated-policy reconcile will be better. I was hoping that the double-reconcile issue here (where this new reconciler does a change which just triggers the root-policy reconcile) could be addressed by a replicated-policy reconciler anyway, but now I see that would still be unnecessary duplication of work. Even with additional changes/fixes, it doesn't seem like my approach will gain us much.

I'm worried that calculating what replicated policies need reconciles based on placement changes will put more work in the `EventHandler` / mapper functions than is intended by controller-runtime, but it's hard for me to tell if that's actually a valid concern.

I think a new bottleneck could come up which is that each replicated policy reconcile will need to calculate its correct decision and any remediationAction overrides... From sketching out different cases of changes, I'll only feel confident it can get it right if it looks through all of the relevant bindings, placements, and decisions for its root policy. But, those will all be in the cache, so it's not a *terrible* problem. It's just that the current approach nicely calculates that for all the replicated policies at once.

The double-reconcile problem will still appear. A placement change will trigger a reconcile on the replicated policy, and if that makes any updates it will trigger a new reconcile. I don't see a good way to filter that out, maybe it's a necessary evil, or we can solve it later. It's still better than an update to one replicated policy effectively causing a re-reconcile for all of the replicated policies.

I don't think a replicated-policy-reconciler would be able to skip some template resolution. But I think with the identified improvements to `go-template-utils`, that will be ok: [ACM-7398](https://issues.redhat.com/browse/ACM-7398), [ACM-7402](https://issues.redhat.com/browse/ACM-7402).

JustinKuli:
I think this will change enough to merit a new PR.

controllers/propagator/propagation.go
JustinKuli: So the new proposed reconciler would watch:
- placement stuff (like the one in this PR, but map to replicated policies)
- replicated policies
- dependency-watches for objects used in a replicated policy's template resolution
- a source that would come from the root-policy-controller telling it that the root spec has changed.

I think we need to figure out where the PolicySet changes reconcile to... Originally I had that in this proposed reconciler, but now I forget why I removed that.

mprahl: @JustinKuli, I just checked and we can't use managed fields like I originally thought to detect the double reconcile.

Perhaps you could keep a cache of resourceVersion entries for the last handled replicated policies? Then a predicate on watching replicated policies could use it.

This would work because dependency watches and the root policy controller source would bypass this check.

PR #130, opened 2023-09-27
Split replicated policy reconciles

JustinKuli:
The first 2 commits are refactors that are not all necessary for the other part of the PR. I can slim down those commits if it helps this PR review. I would recommend looking through the commits individually.

The new replicated policy reconciler helps with two things:
- When a replicated policy is updated, other related replicated policies are not recalculated.
- The root policy reconciler no longer creates/updates/deletes replicated policies itself, via custom concurrency logic. Instead, it simply sends events to the new reconciler for those policies to recalculate their correct states.

**_It does not_** "intelligently" handle updates to replication decisions, eg to not recalculate the replicated policy on managed clusters that should not have changed.

This also adds configurable concurrency to the root policy reconciliation, and the new reconciler has configurable concurrency. The new default root policy concurrency is `2`, and the replicated policy concurrency is `10` (up from the default of `5` that we had in our bespoke concurrency logic).


JustinKuli:
For reference, the resource version caching reduced the number of reconciles done in the e2e tests by about 35%, from [577](https://github.com/JustinKuli/governance-policy-propagator-1/actions/runs/6317785644/job/17155347889#step:9:9390) to [364](https://github.com/JustinKuli/governance-policy-propagator-1/actions/runs/6329609741/job/17190249192#step:9:8705)

JustinKuli:
I am worried that the test failure is because the replicated policy reconcile is now happening too fast? From the failing test:
```
Expected
  ... "label-vendor-test":"auto-detect\n", "label-vendor-test-two":"Fake\n" ...
to equal
  ... "label-vendor-test":"Fake\n", "label-vendor-test-two":"Fake\n" ...
... 

both of these are from the managed cluster label - the first uses `.ManagedClusterLabels`, the second uses a lookup. So I wonder if the update to the ManagedCluster triggered the reconcile, but when the ManagedCluster was grabbed from the cache wasn't updated yet... because the ManagedCluster that the template-utils grab for the lookup *is* correct.

So I'm not really sure what to do about that. It seems to be happening inconsistently. Previously, I think if this happened, it would get handled by the duplicate reconcile.

**Comment by JustinKuli** [comment]
Added some changes after more discussion with @mprahl. In particular, a race condition where the predicates might not correctly filter out events caused by the reconciler itself (causing an unnecessary duplicate reconcile) has been addressed by adding a set of locks.

AFAIK, there are not any more currently requested changes to this PR; it should be ready for another look.

File: controllers/propagator/replicatedpolicy_controller.go [hunk 1]
```diff
@@ -0,0 +1,487 @@
...
+
+			return reconcile.Result{}, err
+		}
+
+		if !replicatedExists {
+			version := safeWriteLoad(r.ResourceVersions, rsrcVersKey)
+			defer version.Unlock()
+
+			// Store this to ensure the cache matches a known possible state for this situation
+			version.resourceVersion = "deleted"
```
mprahl, line 93: Is this cache write needed here since what you did shouldn't trigger a reconcile?

controllers/propagator/replicatedpolicy_controller.go
```diff
@@ -0,0 +1,487 @@
...
+				if !k8serrors.IsNotFound(err) {
+					log.Error(err, "Failed to delete the disabled replicated policy, requeueing")
+
+					return reconcile.Result{}, err
+				}
+			}
+
+			log.Info("Disabled replicated policy deleted")
+
+			return reconcile.Result{}, nil
```
mprahl, line 138: Won't this trigger a double reconcile?

controllers/propagator/replicatedpolicy_controller.go
```diff
@@ -0,0 +1,487 @@
...
+			if !k8serrors.IsNotFound(err) {
+				log.Error(err, "Failed to delete the orphaned replicated policy, requeueing")
+
+				return reconcile.Result{}, err
+			}
+		}
+
+		log.Info("Orphaned replicated policy deleted")
+
+		return reconcile.Result{}, nil
```
mprahl, line 123: Won't this trigger a double reconcile?

controllers/propagator/replicatedpolicy_controller.go
```diff
@@ -0,0 +1,487 @@
...
+				if !k8serrors.IsNotFound(err) {
+					log.Error(err, "Failed to remove the replicated policy for this managed cluster, requeueing")
+
+					return reconcile.Result{}, err
+				}
+			}
+
+			log.Info("Removed replicated policy from managed cluster")
+
+			return reconcile.Result{}, nil
```
mprahl, line 173: Won't this cause a double reconcile?

controllers/propagator/replicatedpolicy_controller.go
mprahl: Nevermind, I see that cleanUpReplicated handles that.

mprahl (approved):
Nice job! This looks really good and thank you for all of your hard work on this.

PR #131, opened 2023-09-28
Fix duplicated event generation when root policy is disabled

zyjjay:
Currently the propagator performs a double reconcile when the root policy is disabled, which leads to duplicated events being created. The work-around to prevent the aforementioned duplicated events is to check for the existence of replicated policies in the managed cluster namespaces. When the root policy is disabled, the first reconcile will remove the replicated policies and on the second reconcile a new check is now added to determine the existence of replicated policies, which allow the controller to safely skip over the logic that emits the second duplicated event.

ref: https://issues.redhat.com/browse/ACM-6122

### DIFF
--- a/controllers/propagator/propagation.go
+++ b/controllers/propagator/propagation.go
@@ -124,8 +124,9 @@ func plcDeletionWrapper(
 	}
 }
 
-// cleanUpPolicy will delete all replicated policies associated with provided policy.
-func (r *PolicyReconciler) cleanUpPolicy(instance *policiesv1.Policy) error {
+// cleanUpPolicy will delete all replicated policies associated with provided policy and return a boolean
+// indicating all of the replicated policies were deleted, if any
+func (r *PolicyReconciler) cleanUpPolicy(instance *policiesv1.Policy) (bool, error) {
 	log := log.WithValues("policyName", instance.GetName(), "policyNamespace", instance.GetNamespace())
 	replicatedPlcList := &policiesv1.PolicyList{}
 
@@ -148,7 +149,7 @@ func (r *PolicyReconciler) cleanUpPolicy(instance *policiesv1.Policy) error {
 			),
 		)
 
-		return err
+		return false, err
 	}
 
 	err = r.List(
@@ -157,13 +158,13 @@ func (r *PolicyReconciler) cleanUpPolicy(instance *policiesv1.Policy) error {
 	if err != nil {
 		log.Error(err, "Failed to list the replicated policies")
 
-		return err
+		return false, err
 	}
 
 	if len(replicatedPlcList.Items) == 0 {
 		log.V(2).Info("No replicated policies to delete.")
 
-		return nil
+		return false, nil
 	}
 
 	log.V(2).Info(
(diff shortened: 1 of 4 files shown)


yiraeChristineKim (dismissed):
Look good to me but hold for others

dhaiducek (changes requested):
@zyjjay Please add a link to the Jira in the commit message so that it propagates to `stolostron`.

controllers/propagator/propagation.go
```diff
@@ -1122,3 +1130,27 @@ func (r *PolicyReconciler) isPolicyInPolicySet(policyName, policySetName, namesp
 
 	return false
 }
+
+// Checks if any replicated policies exists on the hub
```
dhaiducek, line 1134:
```suggestion
// Checks if any replicated policies exist on the hub
```

test/e2e/case1_propagation_test.go
```diff
@@ -26,6 +28,91 @@ const (
...
+					clientHub,
+					testNamespace,
+					case1PolicyName,
+					"PolicyPropagation",
+					"disabled",
+					defaultTimeoutSeconds,
+				))
+
+				return numEvents
+			}, defaultTimeoutSeconds, 1).Should(BeNumerically("<", 2))
```
dhaiducek, line 112: Should this just return `utils.GetMatchingEvents` and use `HaveLen(1)` instead of `BeNumerically()`? Or are there cases where there are zero events?

dhaiducek (commented):
This solution works, but I'm wondering about whether we're duplicating code since both `cleanUpPolicy` and the new function are listing replicated policies (which would happen twice when there are policies to clean up). It might be preferable to have `cleanUpPolicy` return a boolean instead of whether any replicated policies were cleaned up and conditionally emit the event based on that?

test/e2e/case1_propagation_test.go
zyjjay: There are cases where the events array starts with zero elements so to avoid creating a possibly flaky test I used `BeNumerically()`

zyjjay:
> This solution works, but I'm wondering about whether we're duplicating code since both `cleanUpPolicy` and the new function are listing replicated policies (which would happen twice when there are policies to clean up). It might be preferable to have `cleanUpPolicy` return a boolean instead of whether any replicated policies were cleaned up and conditionally emit the event based on that?

I think you're right, since `cleanUpPolicy` already checks for replicated policies, I can just have it return an additional boolean representing the removal of all the replicated policies. I also kept the error return type to satisfy the original behavior of `handleRootPolicy`. 

My initial concern with this change was with regards to `RemoveWatcher`  in `cleanUpPolicy`. I wasn't sure if it would return an error on the second reconcile since the first reconcile would have removed the referenced watcher already, and if it did, it will skip the logic to check for replicated policies. I read the code implementation, tested it locally, and ran the e2e to make sure it wouldn't do that and from what I can gather, it seems to not return an error on the second reconcile. Thanks for catching that!

test/e2e/case1_propagation_test.go
dhaiducek: Ah got it--I missed that it was inside of a `Consistently()` rather than `Eventually()`. 🙂

dhaiducek (approved):

yiraeChristineKim (approved):

PR #132, opened 2023-09-29
Fix 62 ns+name accepted error

yiraeChristineKim:
Actual: namespace + name = 62 chars accepted.

Expect: Should be blocked with error message

Ref: https://issues.redhat.com/browse/ACM-7744

### DIFF
--- a/api/v1/policy_webhook.go
+++ b/api/v1/policy_webhook.go
@@ -58,7 +58,7 @@ func (r *Policy) validateName() (admission.Warnings, error) {
 	}
 
 	// 1 character for "."
-	if (utf8.RuneCountInString(r.Name) + utf8.RuneCountInString(r.Namespace)) > 62 {
+	if (utf8.RuneCountInString(r.Name) + utf8.RuneCountInString(r.Namespace)) >= 62 {
 		return nil, errName
 	}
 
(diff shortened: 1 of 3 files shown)


dhaiducek (commented):
I'm trying to wrap my head around the logic here. I think it was correct originally? 62 characters should be okay:
The limit is 63 characters (ref: https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/#syntax-and-character-set), so:
- This range is invalid:
  `name + namespace + dot > 63`
- Since the dot is 1 character:
  `name + namespace > 62`

So I think the current logic is correct and there's something else going on here to cause the webhook to not work?

dhaiducek:
@yiraeChristineKim Do you agree that this PR can be closed or is there something I'm misunderstanding here?

yiraeChristineKim:
Last time we discussed this... name + namespace = 62 should be accepted. so console was changed

dhaiducek:
`name + namespace = 62` should be okay--that's the maximum. And I think that aligns with the console update:
>The combined length of the policy namespace and name must not exceed 62 characters

https://github.com/stolostron/console/pull/2983/files#diff-bfd4f591207b297ccb35598011a8427e8f66bce78a53bdbb2a7ef0c7bf41dd11

PR #135, opened 2023-10-11
Ignore synced policies for hosted clusters

JustinKuli:
Fixes issues noticed by [CI run in stolostron](https://github.com/stolostron/governance-policy-propagator/pull/459), but not here. In particular, an issue identified by SonarCloud, and an issue with hosted mode clusters.

### DIFF
--- a/controllers/propagator/replicatedpolicy_controller.go
+++ b/controllers/propagator/replicatedpolicy_controller.go
@@ -97,13 +97,14 @@ func (r *ReplicatedPolicyReconciler) Reconcile(ctx context.Context, request ctrl
 			return reconcile.Result{}, nil
 		}
 
-		// do not handle a replicated policy which does not belong to the current cluster
 		inClusterNS, err := common.IsInClusterNamespace(r.Client, request.Namespace)
 		if err != nil {
 			return reconcile.Result{}, err
 		}
 
 		if !inClusterNS {
+			// "Hub of hubs" scenario: this cluster is managed by another cluster,
+			// which has the root policy for the policy being reconciled.
 			log.V(1).Info("Found a replicated policy in non-cluster namespace, skipping it")
 
 			return reconcile.Result{}, nil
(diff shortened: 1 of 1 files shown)


controllers/propagator/replicatedpolicy_controller.go
```diff
@@ -97,13 +97,14 @@ func (r *ReplicatedPolicyReconciler) Reconcile(ctx context.Context, request ctrl
 			return reconcile.Result{}, nil
 		}
 
-		// do not handle a replicated policy which does not belong to the current cluster
```
yiraeChristineKim, line 100: This change is handle replicated policies which belong to the current cluster?

controllers/propagator/replicatedpolicy_controller.go
```diff
@@ -346,12 +360,12 @@ func (r *ReplicatedPolicyReconciler) singleClusterDecision(
...
 
 	// Process all placement bindings without subFilter
-	for _, pb := range pbList.Items {
+	for i, pb := range pbList.Items {
 		if pb.SubFilter == policiesv1.Restricted {
 			continue
 		}
 
-		found, err := r.isSingleClusterInDecisions(ctx, &pb, rootPlc.GetName(), clusterName)
+		found, err := r.isSingleClusterInDecisions(ctx, &pbList.Items[i], rootPlc.GetName(), clusterName)
```
yiraeChristineKim, line 368: Is this for saving memory?

JustinKuli: This is the vulnerability/smell that SonarCloud noticed.

When you take the address of a loop variable like this was doing, you don't get what you'd necessarily expect. Golang kind of re-uses the loop variable in different iterations, so you end up with the same address each time, I think. I believe they want to change it in Go more generally, although I might be confusing some details, and this might not be exactly the same situation: https://go.dev/blog/loopvar-preview

controllers/propagator/replicatedpolicy_controller.go
JustinKuli: The special case being considered here, where it needs to skip dealing with a policy, is caused by the current cluster being a managed cluster for some other hub. In that hub-of-hubs situation, there can be a synced policy on this current cluster, which this reconciler would think is invalid because it can't find that root policy.

controllers/propagator/replicatedpolicy_controller.go
yiraeChristineKim: oh that... got it!! Good catch

yiraeChristineKim (approved):

PR #133, opened 2023-10-03
ACM-7397: Reduce the reconciles caused from policy set updates

JeffeyL:
Currently, a reconcile request is made for every single policy in the policy set on update. A custom event handler was added to only reconcile the differing policies between the new and old policy sets. `policySetPredicateFuncs` was also removed, and the logic was added to the `policySetMapper` function.

Ref: https://issues.redhat.com/browse/ACM-7397

PR #136, opened 2023-10-17
Update only affected replicated policies

yiraeChristineKim:
Refactoring propagated policy controller to update only targeted  replicated policies to reduce meaningless reconcile and improve performance.

Whenever there is a PlacementRule, Placement/PlacementDecision, or PlacementBinding change, it causes the root policy to be regenerated for every managed cluster, even those not affected.

This should be separated out to a separate controller in the Policy Propagator so that a placement update only affects the impacted clusters. For example, if a cluster is added to a placement, the propagated policy should generated just for that one cluster.

This is related to [ACM-7332](https://issues.redhat.com/browse/ACM-7332).

Ref: https://issues.redhat.com/browse/ACM-7403


controllers/propagator/propagation.go
```diff
@@ -54,233 +51,12 @@ type clusterDecision struct {
 	PolicyOverrides policiesv1.BindingOverrides
 }
 
-type decisionSet map[appsv1.PlacementDecision]bool
```
yiraeChristineKim, line 57: move to common

controllers/propagator/replicatedpolicy_controller.go
```diff
@@ -511,3 +528,61 @@ func (r *ReplicatedPolicyReconciler) isSingleClusterInDecisions(
...
+				}, policySet); err != nil {
+					isCorrectPolicyset := false
+
+					for _, plc := range rootPlc.Status.Placement {
+						if plc.PolicySet == sub.Name {
+							isCorrectPolicyset = true
+						}
+					}
+
+					if !isCorrectPolicyset {
```
yiraeChristineKim, line 561: Incorrect policy set like non exist

controllers/propagator/replicatedpolicy_controller.go
```diff
@@ -511,3 +528,61 @@ func (r *ReplicatedPolicyReconciler) isSingleClusterInDecisions(
...
+						log.V(2).Info("Incorrect PolicySet", "policySetName", sub.Name)
+
+						return false, nil
```
yiraeChristineKim, line 564: continue reconcile

controllers/propagator/replicatedpolicy_controller.go
```diff
@@ -511,3 +528,61 @@ func (r *ReplicatedPolicyReconciler) isSingleClusterInDecisions(
...
+					}
+
+					if k8serrors.IsNotFound(err) {
+						log.Info("PolicySet deleted so delete related replicated policy")
+
+						if err := r.cleanUpReplicated(ctx, replicatedPolicy); err != nil {
+							if !k8serrors.IsNotFound(err) {
+								return false, err
```
yiraeChristineKim, line 572: requeue

controllers/common/common.go
```diff
@@ -254,3 +322,26 @@ func TypeConverter(a, b interface{}) error {
 
 	return json.Unmarshal(js, b)
 }
+
+// Select objects that are deleted or created
+func GetAffectedOjbs[T comparable](oldObjs []T, newObjs []T) []T {
```
mprahl, line 327:
```suggestion
func GetAffectedObjs[T comparable](oldObjs []T, newObjs []T) []T {
```

controllers/common/common.go
```diff
@@ -245,6 +245,74 @@ func FullNameForPolicy(plc *policiesv1.Policy) string {
...
+						Namespace: pd.ClusterName,
+					}})
+				}
+			}
+		}
+	}
+
+	var unique []reconcile.Request
+
+	table := map[reconcile.Request]bool{}
```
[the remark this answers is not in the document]
yiraeChristineKim: Sorry, maybe my misunderstood, this map is for getting unique element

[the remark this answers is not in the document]
mprahl: Right, so if you set it directly in the map you can skip the intermediary of `result` since you'll be guaranteeing uniqueness as you add the items in.

controllers/propagator/replicatepolicy_pr_eventHandler.go
```diff
@@ -0,0 +1,168 @@
...
+	}
+
+	for _, pr := range rootPolicyResults {
+		for _, decision := range affectedDecisions {
+			q.Add(reconcile.Request{NamespacedName: types.NamespacedName{
+				Namespace: decision.ClusterNamespace,
+				Name:      pr.Namespace + "." + pr.Name,
+			}})
+		}
+	}
```
mprahl, line 94: This code is more or less duplicated three times. Could you move some of this to helper functions?

controllers/propagator/replicatepolicy_pr_eventHandler.go
```diff
@@ -0,0 +1,168 @@
...
+
+	return result
+}
+
+///////////////////////////////////
```
mprahl, line 168:
```suggestion
```

controllers/common/common.go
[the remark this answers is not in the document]
yiraeChristineKim: ohoh I see what you mean! Thanks let me try

controllers/propagator/replicatepolicy_pb_eventHandler.go
```diff
@@ -0,0 +1,94 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/handler"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+
+	policiesv1 "open-cluster-management.io/governance-policy-propagator/api/v1"
+	"open-cluster-management.io/governance-policy-propagator/controllers/common"
+)
+
+// EnqueueReqsFromMapFuncByBinding maps a PlacementBinding to the targeted RepPolicies that are either directly in its
+// subjects list, or are in a PolicySet which is a subject of this PlacementBinding.
+func EnqueueReqsFromMapFuncByBinding(c client.Client) handler.EventHandler {
```
JustinKuli, line 18: Maybe we can rename these types and functions to something like `HandlerForBinding`

controllers/common/common_status_update.go
```diff
@@ -0,0 +1,394 @@
+// Copyright (c) 2023 Red Hat, Inc.
```
mprahl, line 1:
```suggestion
```

controllers/propagator/replicatepolicy_pb_eventHandler.go
```diff
@@ -0,0 +1,94 @@
...
+	evt event.CreateEvent, q workqueue.RateLimitingInterface,
+) {
+	e.mapAndEnqueue(ctx, q, evt.Object)
+}
+
+// Update implements EventHandler. Update only targeted(modified) objects
+func (e *enqueueReqsFromMapFuncByBinding) Update(ctx context.Context,
+	evt event.UpdateEvent, q workqueue.RateLimitingInterface,
+) {
+	log.Info("Detect placementBinding update and targeting replicated-policy to be changed")
```
JustinKuli, line 41: Make this at least V(1), I think.

JustinKuli (changes requested):
It mostly looks good! I'm just worried about the PolicySet situation

controllers/common/common_status_update.go
```diff
@@ -0,0 +1,394 @@
...
+	sort.Slice(placements, func(i, j int) bool {
+		return placements[i].PlacementBinding < placements[j].PlacementBinding
+	})
+
+	return decisions, placements, nil
+}
+
+type DecisionSet map[appsv1.PlacementDecision]bool
+
+// getClusterDecisions identifies all managed clusters which should have a replicated policy using RootPolicy
```
mprahl, line 251:
```suggestion
// getClusterDecisions identifies all managed clusters which should have a replicated policy using the  root policy
```

controllers/common/common_status_update.go
```diff
@@ -0,0 +1,394 @@
...
+func GetClusterDecisions(
+	c client.Client,
+	instance *policiesv1.Policy,
```
mprahl, line 254: Optional:
```suggestion
	rootPolicy *policiesv1.Policy,
```

controllers/common/common_status_update.go
```diff
@@ -0,0 +1,394 @@
...
+	clusterv1beta1 "open-cluster-management.io/api/cluster/v1beta1"
+	appsv1 "open-cluster-management.io/multicloud-operators-subscription/pkg/apis/apps/placementrule/v1"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	policiesv1 "open-cluster-management.io/governance-policy-propagator/api/v1"
+	policiesv1beta1 "open-cluster-management.io/governance-policy-propagator/api/v1beta1"
+)
+
+// RootStatusUpdat update Root policy status with bound decisions and placements
+func RootStatusUpdate(c client.Client, rootPolicy *policiesv1.Policy) error {
```
mprahl, line 24: I think you should have a context as an argument since you'll have a context from the event handler or reconcile request that can be shared with the API (actually uses the cache) requests in calling functions.

controllers/common/common_status_update.go
```diff
@@ -0,0 +1,394 @@
...
+		return err
+	}
+
+	return nil
+}
+
+// GetPolicyPlacementDecisions retrieves the placement decisions for a input PlacementBinding when
+// the policy is bound within it. It can return an error if the PlacementBinding is invalid, or if
+// a required lookup fails.
+func GetPolicyPlacementDecisions(c client.Client,
```
mprahl, line 66: This should accept a context similar to my comment on `RootPolicyStatus`.

controllers/common/common_status_update.go
```diff
@@ -0,0 +1,394 @@
...
+	decisions, err = GetDecisions(c, pb)
+
+	return decisions, placements, err
+}
+
+// GetAllClusterDecisions calculates which managed clusters should have a replicated policy, and
+// whether there are any BindingOverrides for that cluster. The placements array it returns is
+// sorted by PlacementBinding name. It can return an error if the PlacementBinding is invalid, or if
+// a required lookup fails.
+func GetAllClusterDecisions(
```
mprahl, line 155: This should accept a context similar to my comment on `RootPolicyStatus`.

controllers/common/common_status_update.go
```diff
@@ -0,0 +1,394 @@
...
+		return placements[i].PlacementBinding < placements[j].PlacementBinding
+	})
+
+	return decisions, placements, nil
+}
+
+type DecisionSet map[appsv1.PlacementDecision]bool
+
+// getClusterDecisions identifies all managed clusters which should have a replicated policy using RootPolicy
+func GetClusterDecisions(
```
mprahl, line 252: This should accept a context similar to my comment on `RootPolicyStatus`.

controllers/common/common_status_update.go
```diff
@@ -0,0 +1,394 @@
...
+	}
+
+	return placements, decisions, nil
+}
+
+// CalculatePerClusterStatus lists up all policies replicated from the input policy, and stores
+// their compliance states in the result list. The result is sorted by cluster name. An error
+// will be returned if lookup of a replicated policy fails, but all lookups will still be attempted.
+func CalculatePerClusterStatus(
+	c client.Client,
```
mprahl, line 290: This should accept a context similar to my comment on `RootPolicyStatus`.

controllers/common/common_status_update.go
```diff
@@ -0,0 +1,394 @@
...
+	}
+
+	sort.Slice(status, func(i, j int) bool {
+		return status[i].ClusterName < status[j].ClusterName
+	})
+
+	return status, lookupErr
+}
+
+func IsPolicyInPolicySet(c client.Client, policyName, policySetName, namespace string) bool {
```
mprahl, line 335: This should accept a context similar to my comment on `RootPolicyStatus`.

controllers/common/common_status_update.go
```diff
@@ -0,0 +1,394 @@
...
+			Namespace: rootPolicy.Namespace,
+			Name:      rootPolicy.Name,
+		}, rootPolicy)
+	if err != nil {
+		log.Error(err, "Failed to refresh the cached policy. Will use existing policy.")
+	}
+
+	// make a copy of the original status
+	originalCPCS := make([]*policiesv1.CompliancePerClusterStatus, len(rootPolicy.Status.Status))
+	copy(originalCPCS, rootPolicy.Status.Status)
```
mprahl, line 49: Where is the copy used?

yiraeChristineKim: I copy and pasted, it is used at rootpolicy_controller

yiraeChristineKim: I had a question too lol why copy there?

controllers/common/common_status_update.go
mprahl, line 24: Is this function actually needed to be split off? I only see it used in the root policy status sync controller.

controllers/propagator/replicatedpolicy_controller.go
```diff
@@ -511,3 +524,61 @@ func (r *ReplicatedPolicyReconciler) isSingleClusterInDecisions(
...
+					if !isCorrectPolicyset {
+						log.V(2).Info("Incorrect PolicySet", "policySetName", sub.Name)
+
+						return false, nil
+					}
+
+					if k8serrors.IsNotFound(err) {
+						log.Info("PolicySet deleted so delete related replicated policy")
+
+						if err := r.cleanUpReplicated(ctx, replicatedPolicy); err != nil {
```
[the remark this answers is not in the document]
yiraeChristineKim: how can we distinguish mistakes or missing? cuz without this e2e test fail

controllers/propagator/replicatepolicy_pb_eventHandler.go
```diff
@@ -0,0 +1,94 @@
...
+	return &enqueueReqsFromMapFuncByBinding{
+		c,
+	}
+}
+
+type (
+	enqueueReqsFromMapFuncByBinding struct {
+		c client.Client
+	}
+)
```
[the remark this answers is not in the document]
yiraeChristineKim: oh I referred to original enqueuemapfunc

controllers/propagator/replicatepolicy_pb_eventHandler.go
yiraeChristineKim: `EnqueueReqsFromMapFuncByBinding` -> `HandlerForBinding` ?

controllers/common/common_status_update.go
mprahl: Yeah, it'd be nice to reuse this function in the root policy controller and return the copy for it to call `cleanUpOrphanedRplPolicies` with. Then we wouldn't have code duplication.

controllers/common/common_status_update.go
yiraeChristineKim: yes but like GetAllClusterDecisions, GetPolicyPlacementDecisions, etc used in rootpolicycontroller

yiraeChristineKim: oh you mean only RootStatusUpdate can be moved
 to as a RootPolicyStatusReconciler? OH that  is good idea

controllers/propagator/replicatedpolicy_controller.go
```diff
@@ -126,6 +127,18 @@ func (r *ReplicatedPolicyReconciler) Reconcile(ctx context.Context, request ctrl
 		return reconcile.Result{}, nil
 	}
 
+	// Handle replicated policy, which is related to policyset
+	isHandled, err := r.handlePolicySet(ctx, rootPolicy, replicatedPolicy)
```
mprahl, line 131: Is this actually needed? I think the code below should notice the case where the policy only had a placement bound through the policy set and clean it up appropriately.

yiraeChristineKim: This func cleanup replicated policy connected to policyset... without this, breaks e2e test. I moved all placement watch from root to replicated policy. that is why happen

controllers/common/common_status_update.go
mprahl: Yeah, you can just keep the code in RootPolicyStatusReconcile as is I think since RootStatusUpdate isn't reused in your new code.

controllers/propagator/replicatedpolicy_controller.go
JustinKuli: Which e2e test breaks? I basically disabled this bit, and ran the tests and they seemed to pass (one is re-running for a PolicyAutomation test that mysteriously failed): https://github.com/JustinKuli/governance-policy-propagator-1/pull/10

JustinKuli: What I would expect is that when a PolicySet is deleted, a reconcile is queued for each of the root policies in that set. Each of those then figures out all of the clusters the policy should be on, and when it does that it would notice that the PolicySet was gone, and not consider the placements that were previously associated with that set.

controllers/common/common.go
```diff
@@ -245,6 +248,62 @@ func FullNameForPolicy(plc *policiesv1.Policy) string {
...
+						Namespace: pd.ClusterName,
+					}}
+
+					table[req] = true
+				}
+			}
+		}
+	}
+
+	return maps.Keys(table)
```
yiraeChristineKim, line 304: changed as requeted

controllers/propagator/rootpolicy_setup.go
```diff
@@ -88,119 +78,46 @@ func rootPolicyNonStatusUpdates() predicate.Funcs {
...
+		result = append(result, request)
 	}
+
+	return result
 }
 
-// bindingForPolicy triggers reconciliation if the binding has any Policies or PolicySets in its
-// subjects list.
-func bindingForPolicy() predicate.Funcs {
+// policySetPolicyListChanged triggers reconciliation if the list of policies in the PolicySet has
```
yiraeChristineKim, line 104: This is from rebase

controllers/propagator/rootpolicy_setup.go
```diff
@@ -88,119 +78,46 @@ func rootPolicyNonStatusUpdates() predicate.Funcs {
...
 		},
 		DeleteFunc: func(e event.DeleteEvent) bool {
-			//nolint:forcetypeassert
-			pbObj := e.Object.(*policiesv1.PlacementBinding)
-
-			return common.IsForPolicyOrPolicySet(pbObj)
+			return true
 		},
 	}
 }
```
yiraeChristineKim, line 123: This is from rebase

JustinKuli (approved):
/hold

My comments are barely more than nits. I think the logic of all of this is good, so we don't need to wait for those changes to merge this. I'm holding this in case Matt has any other requests or wants to look over it again.

controllers/common/common.go
```diff
@@ -254,3 +313,68 @@ func TypeConverter(a, b interface{}) error {
...
+
+	return result
+}
+
+type PlacementRefKinds string
+
+const (
+	Placement     PlacementRefKinds = "Placement"
+	PlacementRule PlacementRefKinds = "PlacementRule"
+)
```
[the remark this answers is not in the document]
yiraeChristineKim: Do we handle any other case rather than placementRule or placement for PlacementRef?

controllers/common/common_status_update.go
```diff
@@ -0,0 +1,401 @@
...
+	err = c.Get(ctx,
+		types.NamespacedName{
+			Namespace: rootPolicy.Namespace,
+			Name:      rootPolicy.Name,
+		}, rootPolicy)
+	if err != nil {
+		log.Error(err, "Failed to refresh the cached policy. Will use existing policy.")
+	}
+
+	comlianceState := CalculateRootCompliance(cpcs)
```
mprahl, line 46:
```suggestion
	complianceState := CalculateRootCompliance(cpcs)
```

controllers/common/common_status_update.go
```diff
@@ -0,0 +1,401 @@
...
+	rootPolicy.Status.Status = cpcs
+	rootPolicy.Status.ComplianceState = comlianceState
+	rootPolicy.Status.Placement = placements
+
+	err = c.Status().Update(ctx, rootPolicy)
+	if err != nil {
+		return nil, err
+	}
+
+	return decisions, err
```
mprahl, line 63:
```suggestion

	return decisions, err
```

mprahl, line 63:
```suggestion
	return decisions, nil
```

controllers/common/common_status_update.go
```diff
@@ -0,0 +1,401 @@
...
+		}
+	}
+
+	// If there are no placements, then the PlacementBinding is not for this Policy.
+	if len(placements) == 0 {
+		return nil, nil, nil
+	}
+
+	// If the policy is disabled, don't return any decisions, so that the policy isn't put on any clusters
+	if instance.Spec.Disabled {
```
mprahl, line 145: Is there a reason to not return this in the beginning of the function? Perhaps @JustinKuli remembers.

JustinKuli: I think there are tests that check the status.placements even when it's disabled?

controllers/common/common.go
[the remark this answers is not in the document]
JustinKuli: I guess I'm just saying I prefer the static check against a regular string.

controllers/propagator/rootpolicy_setup.go
```diff
@@ -43,17 +42,8 @@ func (r *RootPolicyReconciler) SetupWithManager(mgr ctrl.Manager, maxConcurrentR
...
-			&policiesv1.PlacementBinding{},
-			handler.EnqueueRequestsFromMapFunc(mapBindingToPolicies(mgr.GetClient())),
-			builder.WithPredicates(bindingForPolicy())).
-		Watches(
-			&appsv1.PlacementRule{},
-			handler.EnqueueRequestsFromMapFunc(mapPlacementRuleToPolicies(mgr.GetClient()))).
-		Watches(
-			&clusterv1beta1.PlacementDecision{},
-			handler.EnqueueRequestsFromMapFunc(mapPlacementDecisionToPolicies(mgr.GetClient()))).
+			handler.EnqueueRequestsFromMapFunc(mapPolicySetToPolicies),
```
mprahl, line 45: This is overwriting the changes made by #133

controllers/common/common.go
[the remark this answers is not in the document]
yiraeChristineKim: Before it was
 ```
		for i, pb := range pbList.Items {
			if pb.PlacementRef.APIGroup != appsv1.SchemeGroupVersion.Group ||
				pb.PlacementRef.Kind != "PlacementRule" || pb.PlacementRef.Name != object.GetName() {
				continue
			}

			result = append(result, common.GetPoliciesInPlacementBinding(ctx, c, &pbList.Items[i])...)
		}
```

File: controllers/common/common.go [hunk 32]
```diff
@@ -245,6 +247,68 @@ func FullNameForPolicy(plc *policiesv1.Policy) string {
...
+						Namespace: pd.ClusterName,
+					}}
+
+					table[req] = true
+				}
+			}
+		}
+	}
+
+	result := []reconcile.Request{}
```
mprahl, line 303: This will save on memory allocations and copies.
```suggestion
	result := make([]reconcile.Request, 0, len(table))
```

controllers/common/common.go
```diff
@@ -245,6 +247,68 @@ func FullNameForPolicy(plc *policiesv1.Policy) string {
...
+	ctx context.Context, c client.Client, pb *policiesv1.PlacementBinding,
+) []reconcile.Request {
+	decisions, err := GetDecisions(ctx, c, pb)
+	if err != nil {
+		return []reconcile.Request{}
+	}
+	// Use this for removing duplicated policies
+	table := map[reconcile.Request]bool{}
+
+	for _, subject := range pb.Subjects {
```
mprahl, line 263: Instead of duplicating this code, could you please call `GetPoliciesInPlacementBinding` and then use that result and `decisions` to generate the reconcile requests for all replicated policies?

controllers/common/common.go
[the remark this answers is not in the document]
mprahl: @yiraeChristineKim I think the issue is that it doesn't get you much since I can create a `PlacementRefKinds` with `PlacementRefKinds("some-random-string")`. That being said, I have seen this used in other libraries and I don't mind it, though it's more of a hint of accepted constants as opposed to any real validation.

controllers/common/common.go
mprahl: I do think that it'd be nice if `GetPoliciesInPlacementBinding` ensured uniqueness, though the queue used by controller-runtime will consolidate them nearly 100% of the time.

controllers/rootpolicystatus/root_policy_status_controller.go
```diff
@@ -91,53 +104,8 @@ func (r *RootPolicyStatusReconciler) Reconcile(ctx context.Context, request ctrl
 
 	log.Info("Updating the root policy status")
```
mprahl, line 105: Can you move this log to inside `RootStatusUpdate` since we don't know if the status will actually get updated?

controllers/common/common.go
```diff
@@ -254,3 +318,68 @@ func TypeConverter(a, b interface{}) error {
...
+type PlacementRefKinds string
+
+const (
+	Placement     PlacementRefKinds = "Placement"
+	PlacementRule PlacementRefKinds = "PlacementRule"
+)
+
+// GetRootPolicyResult find and filter placementbindings which have namespace and placementRef.name.
+// Gather all root policies under placementbindings
+func GetRootPolicyResult(ctx context.Context, c client.Client,
```
mprahl, line 354:
```suggestion
func GetRootPolicyRequests(ctx context.Context, c client.Client,
```

controllers/common/common_status_update.go
```diff
@@ -0,0 +1,401 @@
...
+// GetAllClusterDecisions calculates which managed clusters should have a replicated policy, and
+// whether there are any BindingOverrides for that cluster. The placements array it returns is
+// sorted by PlacementBinding name. It can return an error if the PlacementBinding is invalid, or if
+// a required lookup fails.
+func GetAllClusterDecisions(
+	ctx context.Context,
+	c client.Client,
+	instance *policiesv1.Policy, pbList *policiesv1.PlacementBindingList,
+) (
+	decisions map[appsv1.PlacementDecision]policiesv1.BindingOverrides, placements []*policiesv1.Placement, err error,
```
mprahl, line 163: Where are the binding overrides returned from this function used?

mprahl: @JustinKuli , it looks like you wrote this originally and this PR moved it. I think the binding overrides are not used by any of the callers and instead, the binding overrides are taken into account in `singleClusterDecision`.

yiraeChristineKim: here `GetClusterDecisions` ?

controllers/common/common_status_update.go
```diff
@@ -0,0 +1,401 @@
...
+	decisions, err = GetDecisions(ctx, c, pb)
+
+	return decisions, placements, err
+}
+
+// GetAllClusterDecisions calculates which managed clusters should have a replicated policy, and
+// whether there are any BindingOverrides for that cluster. The placements array it returns is
+// sorted by PlacementBinding name. It can return an error if the PlacementBinding is invalid, or if
+// a required lookup fails.
+func GetAllClusterDecisions(
```
mprahl, line 158: I find this function name really confusing because there is a function called `GetClusterDecisions` which is the only code that calls this function. Perhaps renaming it, or if indeed my comment about the binding overrides returned not being used, the remaining code would be small enough to put in `GetClusterDecisions` and not need this function at all.

controllers/rootpolicystatus/filters.go
```diff
@@ -28,3 +36,56 @@ func policyStatusPredicate() predicate.Funcs {
...
+// mapBindingToPolicies maps a placementBinding to all the Policies in its policies list.
+func mapBindingToPolicies(c client.Client) handler.MapFunc {
+	return func(ctx context.Context, object client.Object) []reconcile.Request {
+		log := log.WithValues("placementBinding", object.GetName(), "namespace", object.GetNamespace())
+		log.V(2).Info("Reconcile Request for placementBinding")
+
+		//nolint:forcetypeassert
+		pb := object.(*policiesv1.PlacementBinding)
+
+		return common.GetPoliciesInPlacementBinding(ctx, c, pb)
```
mprahl, line 49: It occurred to me that that this function doesn't ignore the placement bindings that have `SubFilter` set to `Restricted`. @JustinKuli , what are you thoughts on this?

controllers/common/common_status_update.go
JustinKuli: @mprahl, that sounds right. IIRC, in the refactor PR I was changing so many other things that I wanted to try and keep some things consistent. The overrides information comes naturally while looking through all of the decisions, and isn't much extra data to be shuffling around, so I kept it. I wouldn't expect it to be used by the root policy reconcilers

controllers/common/common_status_update.go
yiraeChristineKim: GetAllClusterDecisions handles binding and subfilter. it is only used in GetClusterDecisions. I guess that the original writer intended GetAllClusterDecisions only for placementbinding. so rename it?!!

controllers/rootpolicystatus/filters.go
JustinKuli: It shouldn't just ignore them always though, right? I think it would need to lookup other bindings for each candidate policy to try and determine what is correct... which is the calculation the root policy reconcile would have to make anyway. So I'm in favor of ignoring the subfilter here, and possibly flagging some policies to be reconciled that don't strictly need to be.

I know the whole intent with the epic is to reduce unnecessary reconciles, but I don't think we need to guarantee that we now do the absolutely minimum number of reconciles. Especially when incorrectly skipping something (because of a bug or oversight) will cause other problems.

mprahl: Yeah, this should be light-weight enough since it's just the root policy status reconciler and if nothing changed, there's no HTTP request made.

controllers/common/common_status_update.go
```diff
@@ -0,0 +1,403 @@
...
+	appsv1 "open-cluster-management.io/multicloud-operators-subscription/pkg/apis/apps/placementrule/v1"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	policiesv1 "open-cluster-management.io/governance-policy-propagator/api/v1"
+	policiesv1beta1 "open-cluster-management.io/governance-policy-propagator/api/v1beta1"
+)
+
+// RootStatusUpdate updates the root policy status with bound decisions, placements, and cluster status.
+func RootStatusUpdate(ctx context.Context, c client.Client, rootPolicy *policiesv1.Policy) (DecisionSet, error) {
+	log.Info("Updating the root policy status")
```
mprahl, line 24: This needs to be put after the `if reflect.DeepEqual(rootPolicy.Status.Status, cpcs) &&` if statement in case you return without doing a status update.

Also, please include the policy name and namespace a key/value pairs in the log message.

controllers/common/common.go
```diff
@@ -245,6 +257,33 @@ func FullNameForPolicy(plc *policiesv1.Policy) string {
...
+	ctx context.Context, c client.Client, pb *policiesv1.PlacementBinding,
+) []reconcile.Request {
+	decisions, err := GetDecisions(ctx, c, pb)
+	if err != nil {
+		return []reconcile.Request{}
+	}
+	// Use this for removing duplicated policies
+	rootPolicyRequest := GetPoliciesInPlacementBinding(ctx, c, pb)
+
+	result := []reconcile.Request{}
```
mprahl, line 273: To reduce memory allocations.
```suggestion
	result := make([]reconcile.Request, 0, len(rootPolicyRequest) * len(decisions))
```

controllers/common/common.go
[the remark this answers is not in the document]
yiraeChristineKim: I thought golang enum more like typescript or Java enum. I didn't know that I could create any random string with that type. Thanks for letting me know

mprahl:
@yiraeChristineKim, the logs are quite confusing now because the code moved in `common.go` for some reason has a logger with the name `policy-automation`. Perhaps this would be best:

```diff
diff --git a/controllers/common/ansible.go b/controllers/automation/ansible.go
similarity index 97%
rename from controllers/common/ansible.go
rename to controllers/automation/ansible.go
index 4464679..a3bf21a 100644
--- a/controllers/common/ansible.go
+++ b/controllers/automation/ansible.go
@@ -1,6 +1,6 @@
 // Copyright Contributors to the Open Cluster Management project
 
-package common
+package automation
 
 import (
        "context"
@@ -15,21 +15,17 @@ import (
        "k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
        "k8s.io/apimachinery/pkg/runtime/schema"
        "k8s.io/client-go/dynamic"
-       ctrl "sigs.k8s.io/controller-runtime"
 
        policyv1beta1 "open-cluster-management.io/governance-policy-propagator/api/v1beta1"
 )
 
 const (
-       ControllerName             string = "policy-automation"
        PolicyAutomationLabel      string = "policy.open-cluster-management.io/policyautomation-name"
        PolicyAutomationGeneration string = "policy.open-cluster-management.io/policyautomation-generation"
        // policyautomation-ResouceVersion
        PolicyAutomationResouceV string = "policy.open-cluster-management.io/policyautomation-resource-version"
 )
 
-var log = ctrl.Log.WithName(ControllerName)
-
 var ansibleJobRes = schema.GroupVersionResource{
        Group: "tower.ansible.com", Version: "v1alpha1",
        Resource: "ansiblejobs",
diff --git a/controllers/automation/policyautomation_controller.go b/controllers/automation/policyautomation_controller.go
index 481d670..e7aa37c 100644
--- a/controllers/automation/policyautomation_controller.go
+++ b/controllers/automation/policyautomation_controller.go
@@ -302,7 +302,7 @@ func (r *PolicyAutomationReconciler) Reconcile(
        }
 
        if policyAutomation.Annotations["policy.open-cluster-management.io/rerun"] == "true" {
-               AjExist, err := common.MatchPAResouceV(policyAutomation,
+               AjExist, err := MatchPAResouceV(policyAutomation,
                        r.DynamicClient, policyAutomation.GetResourceVersion())
                if err != nil {
                        log.Error(err, "Failed to compare Ansible job's resourceVersion")
@@ -323,7 +323,7 @@ func (r *PolicyAutomationReconciler) Reconcile(
 
                violationContext, _ := r.getViolationContext(ctx, policy, targetList, policyAutomation)
 
-               err = common.CreateAnsibleJob(
+               err = CreateAnsibleJob(
                        policyAutomation,
                        r.DynamicClient,
                        "manual",
@@ -373,7 +373,7 @@ func (r *PolicyAutomationReconciler) Reconcile(
                        if len(targetList) > 0 {
                                log.Info("Creating An Ansible job", "targetList", targetList)
                                violationContext, _ := r.getViolationContext(ctx, policy, targetList, policyAutomation)
-                               err = common.CreateAnsibleJob(policyAutomation, r.DynamicClient, "scan",
+                               err = CreateAnsibleJob(policyAutomation, r.DynamicClient, "scan",
                                        violationContext)
                                if err != nil {
                                        return reconcile.Result{RequeueAfter: requeueAfter}, err
@@ -395,7 +395,7 @@ func (r *PolicyAutomationReconciler) Reconcile(
                        if len(targetList) > 0 {
                                log.Info("Creating an Ansible job", "targetList", targetList)
 
-                               AjExist, err := common.MatchPAGeneration(policyAutomation,
+                               AjExist, err := MatchPAGeneration(policyAutomation,
                                        r.DynamicClient, policyAutomation.GetGeneration())
                                if err != nil {
                                        log.Error(err, "Failed to get Ansible job's generation")
@@ -406,7 +406,7 @@ func (r *PolicyAutomationReconciler) Reconcile(
                                        return reconcile.Result{}, nil
                                }
                                violationContext, _ := r.getViolationContext(ctx, policy, targetList, policyAutomation)
-                               err = common.CreateAnsibleJob(
+                               err = CreateAnsibleJob(
                                        policyAutomation,
                                        r.DynamicClient,
                                        string(policyv1beta1.Once),
@@ -517,7 +517,7 @@ func (r *PolicyAutomationReconciler) Reconcile(
                                }
                                log.Info("Creating An Ansible job", "trimmedTargetList", trimmedTargetList)
                                violationContext, _ := r.getViolationContext(ctx, policy, trimmedTargetList, policyAutomation)
-                               err = common.CreateAnsibleJob(
+                               err = CreateAnsibleJob(
                                        policyAutomation,
                                        r.DynamicClient,
                                        string(policyv1beta1.EveryEvent),
diff --git a/controllers/common/common.go b/controllers/common/common.go
index cc40cb8..60d1b68 100644
--- a/controllers/common/common.go
+++ b/controllers/common/common.go
@@ -15,6 +15,7 @@ import (
        clusterv1 "open-cluster-management.io/api/cluster/v1"
        clusterv1beta1 "open-cluster-management.io/api/cluster/v1beta1"
        appsv1 "open-cluster-management.io/multicloud-operators-subscription/pkg/apis/apps/placementrule/v1"
+       ctrl "sigs.k8s.io/controller-runtime"
        "sigs.k8s.io/controller-runtime/pkg/client"
        "sigs.k8s.io/controller-runtime/pkg/reconcile"
 
@@ -29,7 +30,10 @@ const (
        RootPolicyLabel       string = APIGroup + "/root-policy"
 )
 
-var ErrInvalidLabelValue = errors.New("unexpected format of label value")
+var (
+       log                  = ctrl.Log.WithName("common")
+       ErrInvalidLabelValue = errors.New("unexpected format of label value")
+)
 
 // IsInClusterNamespace check if policy is in cluster namespace
 func IsInClusterNamespace(c client.Client, ns string) (bool, error) {
```

mprahl (approved):
This looks good to me! Could you please squash your commits since the last commit now includes other refactoring work/improvements?

@JustinKuli , could you please take a look on Monday morning?

JustinKuli (approved):
/unhold

PR #137, opened 2023-10-17
Cache hub template queries

mprahl:
This leverages the new caching functionality in go-template-utils so that API queries are not duplicated.

As a side-effect, this should cause less reconciles because when starting a watch due to a hub template, it does not cause an initial reconcile merely from the watch being created.

Relates:
https://issues.redhat.com/browse/ACM-7402
https://issues.redhat.com/browse/ACM-7398

### DIFF
--- a/controllers/propagator/encryption.go
+++ b/controllers/propagator/encryption.go
@@ -10,7 +10,7 @@ import (
 	"fmt"
 	"time"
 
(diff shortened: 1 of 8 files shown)


JustinKuli (approved):
This seems so simple after all the library updates!

PR #144, opened 2023-11-07
Handle other changes to PlacementBinding

JustinKuli:
To reduce the number of unnecessary reconciles on replicated policies, when a PlacementBinding is updated, only the added or removed policies are reconciled. However, this meant that when other fields are updated in a PlacementBinding, the affected policies were not being updated accordingly. Now, when there are other changes, all possibly affected replicated policies will be reconciled.

Refs:
 - https://issues.redhat.com/browse/ACM-8515

### DIFF
--- a/controllers/propagator/replicatepolicy_pb_eventHandler.go
+++ b/controllers/propagator/replicatepolicy_pb_eventHandler.go
@@ -49,7 +49,22 @@ func (e *handlerForBinding) Update(ctx context.Context,
 	oldRepPolcies := common.GetRepPoliciesInPlacementBinding(ctx, e.c, oldObj)
 	newRepPolcies := common.GetRepPoliciesInPlacementBinding(ctx, e.c, newObj)
 
-	// Send only affected replicated policies
+	// If the BindingOverrides or subFilter has changed, all new and old policies need to be checked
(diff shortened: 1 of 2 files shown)


controllers/propagator/replicatepolicy_pb_eventHandler.go
```diff
@@ -49,7 +49,22 @@ func (e *handlerForBinding) Update(ctx context.Context,
 	oldRepPolcies := common.GetRepPoliciesInPlacementBinding(ctx, e.c, oldObj)
 	newRepPolcies := common.GetRepPoliciesInPlacementBinding(ctx, e.c, newObj)
 
-	// Send only affected replicated policies
+	// If the BindingOverrides or subFilter has changed, all new and old policies need to be checked
+	if newObj.BindingOverrides.RemediationAction != oldObj.BindingOverrides.RemediationAction ||
+		newObj.SubFilter != oldObj.SubFilter {
+		for _, obj := range newRepPolcies {
```
yiraeChristineKim, line 55: Should we make separate... bindingOverrides already set but user changed only subfilter then it doesn't fall into here.. ??

yiraeChristineKim (approved):

mprahl (approved):

PR #145, opened 2023-11-09
Modest Memory Improvements

mprahl:
See each commit for more details.

### DIFF
--- a/main.go
+++ b/main.go
@@ -232,6 +232,16 @@ func main() {
 				&corev1.Secret{}: {
 					Field: fields.SelectorFromSet(fields.Set{"metadata.name": propagatorctrl.EncryptionKeySecret}),
 				},
+				&policyv1.Policy{}: {
+					Transform: func(obj interface{}) (interface{}, error) {
+						policy := obj.(*policyv1.Policy)
+						// Remove unused large fields
+						delete(policy.Annotations, "kubectl.kubernetes.io/last-applied-configuration")
+						policy.ManagedFields = nil
+
+						return policy, nil
+					},
+				},
 			},
 		},
 	}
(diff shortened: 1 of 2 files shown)


yiraeChristineKim:
good good curious Just how much this is improved compared to previous

main.go
```diff
@@ -232,6 +232,27 @@ func main() {
 				&corev1.Secret{}: {
 					Field: fields.SelectorFromSet(fields.Set{"metadata.name": propagatorctrl.EncryptionKeySecret}),
 				},
+				&clusterv1.ManagedCluster{}: {
+					Transform: func(obj interface{}) (interface{}, error) {
+						cluster := obj.(*clusterv1.ManagedCluster)
+						// All that ManagedCluster objects are used for is to check their existence to see if a
+						// namespace is a cluster namespace.
+						guttedCluster := &clusterv1.ManagedCluster{}
+						guttedCluster.SetName(cluster.Name)
```
yiraeChristineKim, line 241: Label not used? I found this `ManagedClusterLabels map[string]string`

mprahl:
> good good curious Just how much this is improved compared to previous

For real results, we'll have to wait for a performance and scale test but in my fake one, it does reduce the memory allocations by several GB but it doesn't move the needle much on actual memory in use after it becomes idle.

main.go
mprahl: It's a bit confusing but that ManagedCluster isn't queried by the controller-runtime caching client but it's coming from the policy templating client (which uses the caching dependency watcher library):
https://github.com/open-cluster-management-io/governance-policy-propagator/blob/a06836bba155f9ca30edd0ded268d38d222baf3f/controllers/propagator/propagation.go#L285

JustinKuli (approved):
Three nice improvements!

PR #146, opened 2023-11-20
Sync common Makefile

dhaiducek:


mprahl (dismissed):

dhaiducek:
For some reason `make manifests` generates a `_.yaml` file in the CRD directory when the `controller-gen` command is run. Unfortunately, I couldn't pin down why that's happening, so it'll be a mystery to solve another day.

CI is passing now (after resolving a conflict with adding the Kind config arguments).

Makefile
```diff
@@ -29,19 +29,8 @@ export TESTARGS ?= $(TESTARGS_DEFAULT)
 VERSION ?= $(shell cat COMPONENT_VERSION 2> /dev/null)
 IMAGE_NAME_AND_VERSION ?= $(REGISTRY)/$(IMG)
 # Handle KinD configuration
-KIND_NAME ?= test-hub
+CLUSTER_NAME ?= hub
 KIND_NAMESPACE ?= open-cluster-management
-KIND_VERSION ?= latest
-# Set the Kind version tag
-ifeq ($(KIND_VERSION), minimum)
```
yiraeChristineKim, line 36: No more minimum test?

yiraeChristineKim (approved):

PR #149, opened 2023-12-01
Add retries to some templating related failures

mprahl:
In particular, after the replicated policy controller was created, multiple replicated policies for the same cluster can now be processed at the same time. This led to race conditions of each goroutine trying to create the policy encryption key for the first time.

This adds more retries to make hub templates more resilient.

Relates:
https://issues.redhat.com/browse/ACM-8744

### DIFF
--- a/controllers/propagator/propagation.go
+++ b/controllers/propagator/propagation.go
@@ -5,6 +5,7 @@ package propagator
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"strconv"
 	"strings"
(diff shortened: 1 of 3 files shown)


controllers/propagator/propagation.go
```diff
@@ -42,6 +43,19 @@ type Propagator struct {
 	ReplicatedPolicyUpdates chan event.GenericEvent
 }
 
+type RetryableError struct {
```
JustinKuli, line 46: Can we just use this as a special type of error, and we check `errors.Is(RetryableError)`? I'm hesitant to change the return value away from just `error` for things

mprahl: Good idea!

PR #152, opened 2023-12-06
Compliance Event Recording API

JustinKuli:
Implement compliance event POST api

    The server creates new cluster/policy/parent-policy rows as necessary
    for the incoming events. When creating a new policy row, the spec is
    required, but future events for the same policy can use the hash, which
    is returned in the response. If (in the future) we must support policies
    that don't have a spec, we might use a sentinel value.

    Fields omitted in the request are interpreted as empty values: for
    example, not specifying `standards` in a policy means the policy has
    an empty `standards` array, *not* that the server can assign the event
    to any policy matching the other fields.

    The server keeps a local cache of foreign keys that a compliance event
    will need, the keys to those caches are based on the UNIQUE fields in
    the db table schema. It seems like a lot of work, but if the DB is a
    network hop away, the cache can probably save a lot of time.

    The `goqu` library is used here for the queries, but might be removed.

    It was tempting to get all of the get*ForeignKey functions to something
    more generic, but they all have some different queries, and abstracting
    that doesn't seem useful at this point. A common interface might be
    nice in the future.

    Refs:
     - https://issues.redhat.com/browse/ACM-6850


controllers/complianceeventsapi/migrations/000001_compliance_history_initial_tables.up.sql
```diff
@@ -27,7 +27,8 @@ CREATE TABLE IF NOT EXISTS policies(
    parent_policy_id INT,
    spec TEXT,
    -- SHA1 hash
-   spec_hash CHAR[40],
+   spec_hash CHAR(40),
+   -- should spec_hash be NOT NULL ?
```
JustinKuli, line 31: @mprahl I was wondering if all fields which are in the UNIQUE clause/thing below should be required...

mprahl: @JustinKuli it's a good question but I think my intent was to allow importing the existing policy history, which we don't have the historical spec values for.

If we decide against doing that, then I think it'd be better as a required field.

controllers/complianceeventsapi/migrations/000001_compliance_history_initial_tables.up.sql
```diff
@@ -44,7 +45,7 @@ CREATE TABLE IF NOT EXISTS compliance_events(
    compliance TEXT NOT NULL,
    message TEXT NOT NULL,
    timestamp TIMESTAMP NOT NULL,
-   metadata TEXT,
+   metadata JSON,
```
mprahl, line 48: Good call. I originally left it open to TEXT so that we could support SQLite too but we seem to be leaning towards just supporting Postgres.

controllers/complianceeventsapi/server.go
```diff
@@ -0,0 +1,334 @@
...
+	defer s.Lock.Unlock()
+
+	if s.isRunning {
+		return nil
+	}
+
+	mux := http.NewServeMux()
+
+	s.server = &http.Server{
+		Addr:    ":5480",
```
mprahl, line 44: This should be configurable. It should also default to listen only on localhost as well.

controllers/complianceeventsapi/server.go
```diff
@@ -0,0 +1,334 @@
...
+
+	if err := s.server.Shutdown(context.TODO()); err != nil {
+		log.Error(err, "Error stopping compliance events api server")
+	}
+
+	s.isRunning = false
+}
+
+func postComplianceEvent(goquDB *goqu.Database) func(http.ResponseWriter, *http.Request) {
+	return func(w http.ResponseWriter, r *http.Request) {
```
mprahl, line 99: What's the benefit of returning a function in this case?

controllers/complianceeventsapi/server.go
```diff
@@ -0,0 +1,334 @@
...
+		if err := json.Unmarshal(body, reqEvent); err != nil {
+			log.Error(err, "error unmarshalling")
+			http.Error(w, "Incorrectly formatted request body", http.StatusBadRequest)
+
+			return
+		}
+
+		if err := reqEvent.Validate(); err != nil {
+			log.Error(err, "error validating")
+			http.Error(w, "Invalid request body", http.StatusBadRequest)
```
mprahl, line 119: Could you include the error in the response?

controllers/complianceeventsapi/server.go
```diff
@@ -0,0 +1,334 @@
...
+
+	s.isRunning = false
+}
+
+func postComplianceEvent(goquDB *goqu.Database) func(http.ResponseWriter, *http.Request) {
+	return func(w http.ResponseWriter, r *http.Request) {
+		body, err := io.ReadAll(r.Body)
+		if err != nil {
+			log.Error(err, "error reading")
+			http.Error(w, "Error reading request body", http.StatusBadRequest)
```
mprahl, line 103: I'd prefer if the error return JSON to match successful output and it gives us more flexibility later.

For example, `{"message": "Error reading request body"}`

controllers/complianceeventsapi/server.go
```diff
@@ -0,0 +1,334 @@
...
+
+		reqEvent.Event.PolicyID = policyFK
+
+		insert := goquDB.Insert("compliance_events").Rows(reqEvent.Event).Executor()
+
+		_, err = insert.Exec()
+		if err != nil {
+			log.Error(err, "error inserting compliance event")
+			http.Error(w, "Internal Error", http.StatusInternalServerError)
+		}
```
mprahl, line 162: On success, the status code should be 201 and the returned JSON should likely be the serialized form of `reqEvent`. I can also handle the return value in my task on the GET HTTP endpoint.

controllers/complianceeventsapi/server.go
```diff
@@ -0,0 +1,334 @@
...
+
+	// If the row already exists
+	if found {
+		clusterKeyCache.Store(cluster.ClusterID, foundCluster.KeyID)
+
+		return foundCluster.KeyID, nil
+	}
+
+	// Otherwise, create a new row in the table
+	insert := goquDB.Insert("clusters").Returning("id").Rows(cluster).Executor()
```
mprahl, line 190: How is this guarding against SQL injection?

controllers/complianceeventsapi/server.go
```diff
@@ -0,0 +1,334 @@
...
+		WriteTimeout: 15 * time.Second,
+		IdleTimeout:  15 * time.Second,
+	}
+
+	db, err := sql.Open("postgres", dbURL)
+	if err != nil {
+		return err
+	}
+
+	goquDB := goqu.New("postgres", db)
```
mprahl, line 58: It kind of looks like the project is on life support. There are a bunch of open PRs with no responses and then there was one release recently in the last two years with one change.

controllers/complianceeventsapi/server.go
```diff
@@ -0,0 +1,334 @@
...
+	parentPolicyKeyCache.Store(parKey, *id)
+
+	return *id, nil
+}
+
+func getPolicyForeignKey(ctx context.Context, goquDB *goqu.Database, pol Policy) (int, error) {
+	// Fill in missing fields that can be inferred from other fields
+	if pol.SpecHash == nil {
+		var buf bytes.Buffer
+		if err := json.Compact(&buf, []byte(*pol.Spec)); err != nil {
```
mprahl, line 260: Is this deterministic? For example, are struct fields always marshaled in the same order?

mprahl: I guess it'd be up to the API caller to make this deterministic.

controllers/complianceeventsapi/types.go
```diff
@@ -0,0 +1,264 @@
...
+
+	switch s := src.(type) {
+	case string:
+		source = []byte(s)
+	case []byte:
+		source = s
+	case nil:
+		source = nil
+	default:
+		return errors.New("incompatible type for JsonMap")
```
mprahl, line 260:
```suggestion
		return errors.New("incompatible type for JSONMap")
```

controllers/complianceeventsapi/types.go
```diff
@@ -0,0 +1,264 @@
...
+	Severity  string
+}
+
+type JSONMap map[string]interface{}
+
+func (j JSONMap) Value() (driver.Value, error) {
+	return json.Marshal(j)
+}
+
+func (j *JSONMap) Scan(src interface{}) error {
```
mprahl, line 249: Could you please document this? When does this get used?

test/e2e/case18_compliance_api_test.go
```diff
@@ -112,4 +110,757 @@ var _ = Describe("Test policy webhook", Label("compliance-events-api"), Ordered,
...
 	})
+
+	Describe("Test POSTing Events", func() {
+		Describe("POST one valid event with including all the optional fields", func() {
+			payload := []byte(`{
+				"cluster": {
+					"name": "cluster1",
+					"cluster_id": "test1-cluster1-fake-uuid-1"
+				},
+				"parent_policy": {
```
mprahl, line 121: Do you think this would be better as a key under the `policy` map?

test/e2e/case18_compliance_api_test.go
```diff
@@ -112,4 +110,757 @@ var _ = Describe("Test policy webhook", Label("compliance-events-api"), Ordered,
...
+					Expect(err).ToNot(HaveOccurred())
+
+					Expect(id).NotTo(Equal(0))
+					Expect(clusterID).NotTo(Equal(0))
+					Expect(policyID).NotTo(Equal(0))
+					Expect(compliance).To(Equal("NonCompliant"))
+					Expect(message).To(Equal("configmaps [etcd] not found in namespace default"))
+					Expect(timestamp).To(Equal("2023-01-01T01:01:01.111Z"))
+					Expect(metadata).To(HaveKeyWithValue("test", true))
+					// Expect(*metadata).To(Equal(`{"test":true}`))
```
mprahl, line 267:
```suggestion
```

test/e2e/case18_compliance_api_test.go
```diff
@@ -112,4 +110,757 @@ var _ = Describe("Test policy webhook", Label("compliance-events-api"), Ordered,
...
+					var (
+						id       int
+						kind     string
+						apiGroup string
+						name     string
+						ns       *string
+						pid      *int
+						spec     *string
+						specHash *string
+						severity *string
```
mprahl, line 212: So you have to use a pointer for potentially null values?

test/e2e/case18_compliance_api_test.go
```diff
@@ -112,4 +110,757 @@ var _ = Describe("Test policy webhook", Label("compliance-events-api"), Ordered,
...
+			payload2 := []byte(`{
+				"cluster": {
+					"name": "cluster3",
+					"cluster_id": "test2-cluster3-fake-uuid-3"
+				},
+				"policy": {
+					"apiGroup": "policy.open-cluster-management.io",
+					"kind": "ConfigurationPolicy",
+					"name": "etcd-encryption2",
+					"specHash": "cb84fe29e44202e3aeb46d39ba46993f60cdc6af"
```
mprahl, line 305: I think this shouldn't be allowed. Either the specHash is known in the database and if not, you must provide the spec.

controllers/complianceeventsapi/migrations/000001_compliance_history_initial_tables.up.sql
mprahl: @JustinKuli let me clarify, I meant direct insertions in the database from the script instead of through the API.

Now that I think of it, let's go ahead and make it required and we can put in a dummy value for when it's historically imported.

controllers/complianceeventsapi/server.go
JustinKuli: Originally I was going to use it directly in the `mux.HandleFunc`... later I added the `switch` for the HTTP method, and that makes this not make as much sense any more. I'll clean that up, good catch.

controllers/complianceeventsapi/server.go
JustinKuli: I vaguely remembered some advice at some point like "don't leak errors in your APIs" (as a potential security concern), but actually researching it now I've found some good advice. I'll go over these returns again with that in mind.

controllers/complianceeventsapi/server.go
JustinKuli: It should be easy enough for me to put it in here.

controllers/complianceeventsapi/server.go
JustinKuli: My understanding is that the `.Rows(cluster)` part handles any required escaping to prevent injections. Do you think it's missing something?

controllers/complianceeventsapi/server.go
JustinKuli: That's a good point. The docs were pretty good, and I didn't vet whether it was still very active. Do you think it's bad enough to prevent using it initially here? IMO, the query builder is something we could potentially switch out in a future PR (and the tests don't use the query builder, which could help the swap)

controllers/complianceeventsapi/server.go
JustinKuli: I'll check on this.

test/e2e/case18_compliance_api_test.go
JustinKuli: Mostly I don't think it would matter much. The current scheme is kind of nice because the stuff in the `policy` basically matches the stuff in a k8s `policy` object (which does not have this proposed kind of embedding of the parent policy)

test/e2e/case18_compliance_api_test.go
JustinKuli: I see what you mean... we don't want an entry in the db that only has the hash, and doesn't have the full spec.

On the other hand, it feels a little weird that the user needs to know what was already in the database, or that a request might fail at one point, but then succeed later with no other changes from them (because in the meantime, something else could have put that policy spec in the database).

test/e2e/case18_compliance_api_test.go
```diff
@@ -112,4 +110,757 @@ var _ = Describe("Test policy webhook", Label("compliance-events-api"), Ordered,
...
+				Expect(ids).To(HaveLen(2))
+				Expect(ids[0]).ToNot(Equal(ids[1]))
+				Expect(names[0]).To(Equal(names[1]))
+				Expect(pids[0]).ToNot(Equal(pids[1]))
+				Expect(hashes[0]).To(Equal(hashes[1]))
+			})
+		})
+
+		Describe("POST invalid events", func() {
+			_ = []byte(`{
```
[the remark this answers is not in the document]
JustinKuli: Sorry, that was the event I was pulling pieces out of for the other tests, I didn't mean to leave it in there.

controllers/complianceeventsapi/server.go
JustinKuli: Per our conversation, the plan is to merge this with `goqu`, and explore other options. Ideally we remove it soon (either for standard library go only, or another query builder), but we want to un-block anything else that is waiting for the initial API.

controllers/complianceeventsapi/server.go
JustinKuli: It doesn't unmarshal and then marshal again, so the order in the request does matter. Switching the order of the keys changes the hash later.

I think we should somehow discourage users from trying to do the hash themselves: if they don't know what it is, they should just submit the whole spec, and the response should have the hash in case they want to submit more events for the same policy. But I'm not sure the ideal design to accomplish that.

test/e2e/case18_compliance_api_test.go
mprahl: I think the normal use-case is to provide `spec` and not `specHash`, however, to save on bandwidth, you can cache the `specHash` to reuse for future API usage.

In our use-case, the propagator will likely prepopulate the `parent_policies` and `policies` tables so to save on bandwidth in edge scenarios so the managed cluster just needs to send `specHash`.

Alternatively, we could have the propagator set the database ID for the `policies` table entry on the `policy-templates` entries and use that from the managed cluster side.

controllers/complianceeventsapi/server.go
mprahl: I agree to your last comment. I think we need an algorithm that orders the JSON consistently so the hash is always the same from the same JSON.

I don't really see a way around this unless we just use the `policy-templates` entry as the "official" spec and the template sync sets the spec hash as an annotation or something on the policy template (e.g. ConfigurationPolicy).

test/e2e/case18_compliance_api_test.go
JustinKuli: Confirmed (and maybe saved you an experiment): if a column is NULL and you don't try to put it into a pointer, you get an error like this:
```
sql: Scan error on column index 8, name \"severity\": converting NULL to string is unsupported
```

On the other hand, I think you can always use a pointer - as far as I can tell, I could use a pointer for any of the `id` fields. It's just slightly annoying to de-reference later, and makes me worry about nil pointers.

JustinKuli:
@mrpahl, I believe I have addressed your comments. Let me know if I missed anything, or if you have any more 😀

controllers/complianceeventsapi/server.go
mprahl: The only info I could find of this online is this:
https://github.com/doug-martin/goqu/issues/279#issuecomment-1367819827

I haven't looked at the code though.

controllers/complianceeventsapi/migrations/000001_compliance_history_initial_tables.up.sql
mprahl: @JustinKuli I think JSONB might be better here for efficient querying.

test/e2e/case18_compliance_api_test.go
```diff
@@ -112,4 +110,758 @@ var _ = Describe("Test policy webhook", Label("compliance-events-api"), Ordered,
...
+					Expect(err).ToNot(HaveOccurred())
+
+					Expect(id).NotTo(Equal(0))
+					count++
+				}
+
+				Expect(count).To(Equal(1))
+			})
+
+			It("Should have only created one policy in a table", func() {
```
mprahl, line 507: Actually, this isn't quite right because even though they share similar values, the parent policy and severity are part of the uniqueness constraint, so there should be separate rows.

This stems from my misunderstanding of how nulls are treated in the uniqueness constraints. I'll have a PR up with this fixed.

JustinKuli: Ok, I think I also had some confusion: I was interpreting omitted fields as "I don't care about these fields, find anything that matches the rest of my request" as opposed to "that field must be empty".

controllers/complianceeventsapi/server.go
mprahl: I think this is solved with the idea of setting the `policy` ID and `parent_policy` ID as annotations on the `ConfigurationPolicy`.

test/e2e/case18_compliance_api_test.go
```diff
@@ -112,4 +110,886 @@ var _ = Describe("Test policy webhook", Label("compliance-events-api"), Ordered,
...
+					"name": "cluster1",
+					"cluster_id": "test1-cluster1-fake-uuid-1"
+				},
+				"parent_policy": {
+					"name": "policies.etcd-encryption1",
+					"categories": ["cat-1", "cat-2"],
+					"controls": ["ctrl-1"],
+					"standards": ["stand-1"]
+				},
+				"policy": {
```
mprahl, line 127: This is missing "severity".

test/e2e/case18_compliance_api_test.go
```diff
@@ -112,4 +110,886 @@ var _ = Describe("Test policy webhook", Label("compliance-events-api"), Ordered,
...
+					Expect(err).ToNot(HaveOccurred())
+
+					Expect(id).NotTo(Equal(0))
+					count++
+				}
+
+				Expect(count).To(Equal(1))
+			})
+
+			It("Should have only created one policy in a table", func() {
```
mprahl, line 513: This should be 2 policies right? Since severity is set on one but not the other?

controllers/complianceeventsapi/server.go
```diff
@@ -0,0 +1,366 @@
...
+	_ "github.com/lib/pq"
+)
+
+var (
+	clusterKeyCache      sync.Map
+	parentPolicyKeyCache sync.Map
+	policyKeyCache       sync.Map
+)
+
+type databaseAPIServer struct {
```
mprahl, line 28: I think omitting "database" from the struct name makes it clearer that this isn't a database.
```suggestion
type complianceAPIServer struct {
```

test/e2e/case18_compliance_api_test.go
JustinKuli: In this test, payload1 has the severity inside the spec. Since payload2 uses specHash, the severity can't be extracted in the same way, and specifying the severity separately is necessary.

Realistically, the caller should probably always specify severity...

controllers/complianceeventsapi/server.go
```diff
@@ -0,0 +1,366 @@
...
+func (s *databaseAPIServer) Stop() {
+	s.Lock.Lock()
+	defer s.Lock.Unlock()
+
+	if !s.isRunning {
+		return
+	}
+
+	if err := s.server.Shutdown(context.TODO()); err != nil {
+		log.Error(err, "Error stopping compliance events api server")
```
mprahl, line 95: Why not return the error in this case?

test/e2e/case18_compliance_api_test.go
mprahl: Oh I see.... I'm thinking that we shouldn't try to parse the severity in this API. What do you think?

I could go either way but I lean towards not trying to parse it.

controllers/complianceeventsapi/server.go
JustinKuli: It's in a different goroutine

controllers/complianceeventsapi/types.go
```diff
@@ -0,0 +1,266 @@
...
+}
+
+func (e EventDetails) Validate() error {
+	errs := make([]error, 0)
+
+	if e.Compliance == "" {
+		errs = append(errs, fmt.Errorf("%w: event.compliance", errRequiredFieldNotProvided))
+	} else {
+		switch e.Compliance {
+		case "Compliant", "compliant", "NonCompliant", "noncompliant":
```
mprahl, line 92: What do you think about just allowing `compliant` and `noncompliant` so we don't have to account for different cases in queries later on?

test/e2e/case18_compliance_api_test.go
JustinKuli: Obviously it's already coded to parse the severity, so I thought it was a good idea 😆 

I suppose requiring the severity separately in some situations, but not others, could be annoying to the user.

controllers/complianceeventsapi/types.go
```diff
@@ -0,0 +1,266 @@
...
+	for i, c := range p.Controls {
+		ctrls[i] = fmt.Sprintf("%q", c) // double-quoted and safely escaped with Go syntax
+	}
+
+	stands := make([]string, len(p.Standards))
+	for i, c := range p.Standards {
+		stands[i] = fmt.Sprintf("%q", c) // double-quoted and safely escaped with Go syntax
+	}
+
+	return parentPolicyKey{
```
mprahl, line 142: I think this or similar would be more efficient as a key.
```go
fmt.Sprintf("%s;%#v;%#v;%#v", p.Name, p.Categories, p.Controls, p.Standards)
```

controllers/complianceeventsapi/server.go
JustinKuli: Ah, sorry, my brain read the wrong line (line 76 above it is in a separate go routine).

You're right, it seems like it can return the error.

controllers/complianceeventsapi/types.go
```diff
@@ -0,0 +1,266 @@
...
+
+	if p.SpecHash != nil {
+		key.SpecHash = *p.SpecHash
+	}
+
+	if p.Severity != nil {
+		key.Severity = *p.Severity
+	}
+
+	return key
```
mprahl, line 230: Similar comment to above, what do you think about generating a uniquely identifying string instead? I think this is less important in this case since you aren't allocating new slices for the conversion as you were for the parent policy.

test/e2e/case18_compliance_api_test.go
mprahl: Yeah, I think reducing the "magic" is better in this instance.  :smile:

controllers/complianceeventsapi/types.go
JustinKuli: I agree in principle, but I think our resources most often use Compliant and NonCompliant, so that's what I'll make this accept. Unless you have a really good reason to use all lowercase?

mprahl: @JustinKuli I think just allowing a single variation of the value is what I'm getting at. If we want to use `Compliant` and `NonCompliant`, that sounds good to me, though `NonCompliant` is not correct English and is strange for non-OCM use-cases if this gets reused. :laughing:

controllers/complianceeventsapi/types.go
JustinKuli: `%#v` was giving a difference between nil and empty, but `%v` worked fine, FYI.

controllers/complianceeventsapi/types.go
JustinKuli: The pointers would still basically need to be handled this way, so I don't think changing it will really help.

mprahl (approved):

PR #155, opened 2023-12-18
Add partial indexes to account for null values

mprahl:
In Postgres 15+, NULLS NOT DISTINCT is supported so that a null value in
the same column is seen as the same value. This is not standard SQL and
Postgres 15 is likely too new to require. This is a workaround to
include partial indexes when some null values are provided.

Note that the first commit helps debug in a test/dev environment.

### DIFF
--- a/build/kind/postgres.yaml
+++ b/build/kind/postgres.yaml
@@ -68,6 +68,10 @@ spec:
             - ssl_key_file=/var/lib/postgresql/tls/tls.key
             - -c
             - ssl_ca_file=/var/lib/postgresql/tls/ca.crt
+            - -c
+            - log_statement=all
+            - -c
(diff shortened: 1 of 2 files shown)


build/kind/postgres.yaml
```diff
@@ -68,6 +68,10 @@ spec:
             - ssl_key_file=/var/lib/postgresql/tls/tls.key
             - -c
             - ssl_ca_file=/var/lib/postgresql/tls/ca.crt
+            - -c
+            - log_statement=all
+            - -c
+            - log_destination=stderr
```
JustinKuli, line 74: I could just check this myself I guess, but is this just a "echo all incoming requests" kind of thing?

JustinKuli (approved):

PR #156, opened 2023-12-19
Drop `ClusterNamespace` internally to prep for Placement

dhaiducek:
Also:
- Add `e2e-test-coverage-foreground` target
- Add kubeconfig to `.gitignore`
- Skip `GuttedObject` for generation (This was resulting in an empty `_.yaml` CRD file being generated)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -97,3 +97,4 @@ gosec.json
 vendor/
 bin
 testbin/*
+kubeconfig_*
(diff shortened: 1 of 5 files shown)


mprahl:
@dhaiducek I think you're right that it's not necessary but it is a breaking change.... so I'm not sure it's worth it.

dhaiducek:
> @dhaiducek I think you're right that it's not necessary but it is a breaking change.... so I'm not sure it's worth it.

Okay. Yeah, I guess usability probably outweighs my desire to rip it out. 😅 
The other commits are relevant though, so I'll pare it down.

dhaiducek:
Repurposed and ready for review!

mprahl (approved):

PR #157, opened 2023-12-20
Update compliance database schema

mprahl:
This adds a namespace field to the parent policy table so that the root policy namespace and name is recorded instead of the replicated policy name which includes the root policy namespace in it. This is more intuitive since the user interacts with the root policy and not often with the replicated policy.

This also moves the parent policy foreign key to the compliance_events table. This is better as a point in time record of the parent policy associated with the policy that generated the compliance event. It does make filtering more efficient later to see any compliance event associated with a parent policy. It also allows less data duplication on the policies table.

As part of this refactoring, goku was removed in favor of a minimal translation layer of database queries to structs and vice versa.


controllers/complianceeventsapi/migrations/000001_compliance_history_initial_tables.up.sql
```diff
@@ -12,63 +12,55 @@ CREATE INDEX IF NOT EXISTS idx_clusters_name ON clusters (name);
 CREATE TABLE IF NOT EXISTS parent_policies(
    id serial PRIMARY KEY,
    name TEXT NOT NULL,
+   namespace TEXT NOT NULL,
    categories TEXT [],
    controls TEXT [],
    standards TEXT [],
    UNIQUE (name, categories, controls, standards)
```
JustinKuli, line 19: Does this also need `namespace`?

test/e2e/case18_compliance_api_test.go
```diff
@@ -244,23 +248,26 @@ var _ = Describe("Test policy webhook", Label("compliance-events-api"), Ordered,
...
-					err := rows.Scan(&id, &clusterID, &policyID, &compliance, &message, &timestamp,
+					err := rows.Scan(&id, &clusterID, &policyID, &parentPolicyID, &compliance, &message, &timestamp,
 						&metadata, &reportedBy)
 					Expect(err).ToNot(HaveOccurred())
 
 					Expect(id).NotTo(Equal(0))
 					Expect(clusterID).NotTo(Equal(0))
 					Expect(policyID).NotTo(Equal(0))
+					Expect(parentPolicyID).NotTo(BeNil())
+					Expect(parentPolicyID).NotTo(Equal(0))
```
JustinKuli, line 270:
```suggestion
					Expect(*parentPolicyID).NotTo(Equal(0))
```

test/e2e/case18_compliance_api_test.go
```diff
@@ -550,23 +562,26 @@ var _ = Describe("Test policy webhook", Label("compliance-events-api"), Ordered,
...
```
JustinKuli, line 584:
```suggestion
					Expect(*parentPolicyID).NotTo(Equal(0))
```

controllers/complianceeventsapi/migrations/000001_compliance_history_initial_tables.up.sql
mprahl: Yeah, my thought is that OCM policies in the `policies` table won't set the namespace because it's always the cluster namespace to prevent data duplication. The root/parent policy namespace will be a separate column rather than the original plan of `<namespace>.<name>`. This will make filtering easier and allow us to have namespaced based RBAC for viewing compliance events in the future if desired.

mprahl: I've updated the design PR to reflect that:
https://github.com/open-cluster-management-io/enhancements/pull/105

controllers/complianceeventsapi/server.go
```diff
@@ -291,46 +239,37 @@ func getPolicyForeignKey(ctx context.Context, goquDB *goqu.Database, pol Policy)
...
-
-	// If the row already exists
-	if found {
-		policyKeyCache.Store(polKey, foundPolicy.KeyID)
-
-		return foundPolicy.KeyID, nil
-	}
+	if pol.Spec == "" {
+		row := db.QueryRowContext(
+			ctx, fmt.Sprintf("SELECT spec FROM %s WHERE spec_hash = $1 LIMIT 1", policyTableName), pol.SpecHash,
```
JustinKuli, line 244: Tangent/opinion: this `Sprintf` feels weird. How much do we gain by making the table name change-able later?

I think I'd rather either have a magic string here that includes the table name, or the whole query as a constant defined elsewhere. I like that the actual query is obvious here, and a magic string would keep that nicely. But, having the whole query defined somewhere else would let us keep all of these kinds of queries defined together, possibly making it easier to review whether they are following best practices. Naming the constants might be clunky, but if they were just magic strings, I expect IDEs would be able to reveal the actual query when hovered over (VS Code does at least), which I think would otherwise be the main downside.

controllers/complianceeventsapi/types.go
```diff
@@ -50,10 +70,34 @@ func (ce ComplianceEvent) Validate() error {
...
+		ce.Event.ClusterID = ce.Cluster.KeyID
+	}
+
+	if ce.Event.PolicyID == 0 {
+		ce.Event.PolicyID = ce.Policy.KeyID
+	}
+
+	insertQuery, insertArgs := ce.Event.InsertQuery()
+
+	row := db.QueryRowContext(ctx, insertQuery+"RETURNING id", insertArgs...) //nolint:execinquery
```
JustinKuli, line 84: The silly linter should recognize RETURNING...

Another tangent/discussion: it's not as easy as I would like to see the exact query this uses, partly because the `InsertQuery` calls through two other functions... It's nice that multiple things can use `generateInsertQuery`, but I'm wondering about the overall cost/benefit.

controllers/complianceeventsapi/migrations/000001_compliance_history_initial_tables.up.sql
mprahl: Good catch! Thanks!

controllers/complianceeventsapi/server.go
mprahl: Sure, I'll add it directly but I like having the query directly here than as a constant.

mprahl:
@JustinKuli the latest commit makes the SQL queries static. Let me know what you think of this approach!

controllers/complianceeventsapi/types.go
```diff
@@ -236,3 +385,130 @@ func (j *JSONMap) Scan(src interface{}) error {
...
+		field.Set(src)
+	}
+
+	return nil
+}
+
+// getOrCreate will translate the input object to an INSERT SQL query. When the input object already exists in the
+// database, a SELECT query is performed. The primary key is set on the input object when it is inserted or gotten
+// from the database. The INSERT first then SELECT approach is a clean way to account for race conditions of multiple
+// goroutines creating the same row.
```
JustinKuli, line 475: Neat! I don't think I fully appreciated this on my first review.

controllers/complianceeventsapi/types.go
```diff
@@ -19,6 +22,13 @@ var (
 	errInvalidInput             = errors.New("invalid input")
 )
 
+type dbRow interface {
+	FromRow(*sql.Rows) error
+	GetOrCreate(ctx context.Context, db *sql.DB) error
+	InsertQuery() (string, []any)
+	SelectQuery(returnedColumns ...string) (string, []any)
+}
```
JustinKuli, line 30: I think this interface can/should just be InsertQuery and SelectQuery.

The one current use of the interface is in the `getOrCreate` function, which only uses the methods listed above. It feels especially odd to include GetOrCreate in the interface, because all those implementations use `getOrCreate`, and are in some sense circular.

That said, I'd be fine with accepting the PR as-is.

controllers/complianceeventsapi/types.go
```diff
@@ -236,3 +385,130 @@ func (j *JSONMap) Scan(src interface{}) error {
...
+
+		fields[fieldName] = field
+	}
+
+	return fields
+}
+
+// fromRow assigns the output from `rows.Scan` to the input `obj`. Note that this does not explicitly support
+// nested structs.
+func fromRow(rows *sql.Rows, obj any) error {
```
JustinKuli, line 437: Especially if you slimmed the `dbRow` interface, we could rename this function to `ScanInto`. To me, it feels slightly more representative of what's going on.

JustinKuli (approved):
/hold

I added some optional comments. Feel free to un-hold if you'd rather address them in a future PR (or not at all)

controllers/complianceeventsapi/types.go
mprahl: When writing the GET HTTP endpoint, I found myself not using this because I always had a single large query with several joins and just referenced the fields directly in a single scan, so I think I'll just remove this.

mprahl:
@JustinKuli alright, all the complicated "cool" stuff has been removed from the PR now. :laughing: I kept them in the commit history in case we decided to go back that route.

JustinKuli (approved):

JustinKuli:
/unhold

Thanks (and sorry) for all the back-and-forth, but I think this looks really great now

PR #161, opened 2024-01-05
Add the API to get compliance events

mprahl:
This adds a list API endpoint at /api/v1/compliance-events and allows an individual compliance event to be retrieved with
/api/v1/compliance-events/<id>.

The list API endpoint supports sorting, pagination, and the flag of `includeSpec` which is disabled by default.

Relates:
https://issues.redhat.com/browse/ACM-6857


controllers/complianceeventsapi/server.go
```diff
@@ -92,6 +181,357 @@ func (s *complianceAPIServer) Stop() error {
 	return nil
 }
 
+// splitQueryValue will parse a string and split on unescaped commas. Empty values are discarded.
+func splitQueryValue(value string) []string {
```
mprahl, line 185: Note that a normal string split would work just fine in this case because no sort options have a comma in them but I plan to reuse this in my next PR for filter query arguments.

controllers/complianceeventsapi/server.go
```diff
@@ -92,6 +181,357 @@ func (s *complianceAPIServer) Stop() error {
...
+	values := []string{}
+
+	var currentVal string
+	var previousChar rune
+
+	for _, char := range value {
+		if char == ',' && previousChar != '\\' {
```
yiraeChristineKim, line 192: Do you have a unit test for this? I am confused that previousChar could be '\\'? Isn't '\\' 2 chars?

controllers/complianceeventsapi/server.go
```diff
@@ -92,6 +181,357 @@ func (s *complianceAPIServer) Stop() error {
...
+
+	pageStr := queryArgs.Get("page")
+
+	if pageStr == "" {
+		parsed.Page = 1
+	} else {
+		var err error
+
+		parsed.Page, err = strconv.ParseUint(pageStr, 10, 64)
+		if err != nil || parsed.Page == 0 {
```
yiraeChristineKim, line 242: Oh What if user add minus number such as -10

JustinKuli (commented):
Looking good overall! One extra question that sparked in my mind during this: will there automatically be a metric (either in our container, or through something in kubernetes) that says how many requests have succeeded/errored? In particular, I'm thinking that it will be necessary to monitor when there is an unusual number of internal errors or something like that.

controllers/complianceeventsapi/server.go
mprahl: '\\' is actual one character because the first backslash is an escape character. I realized this function wasn't removing the escape character so in my push I'll be adding a unit test to catch this.

test/e2e/case18_compliance_api_test.go
```diff
@@ -412,7 +499,372 @@ var _ = Describe("Test the compliance events API", Label("compliance-events-api"
...
+				Expect(err).ToNot(HaveOccurred())
+
+				resp, err := httpClient.Do(req)
+				Expect(err).ToNot(HaveOccurred())
+
+				defer resp.Body.Close()
+
+				Expect(resp.StatusCode).To(Equal(http.StatusBadRequest))
+
+				body, err := io.ReadAll(resp.Body)
```
yiraeChristineKim, line 822: It doesn't really matter but I found https://daryl-ng.medium.com/why-you-should-avoid-ioutil-readall-in-go-e6be4de180f8

controllers/complianceeventsapi/server.go
```diff
@@ -92,6 +181,357 @@ func (s *complianceAPIServer) Stop() error {
...
+			log.Error(err, "Failed to unmarshal the database results")
+			writeErrMsgJSON(w, "Internal Error", http.StatusInternalServerError)
+
+			return
+		}
+
+		complianceEvents = append(complianceEvents, *ce)
+	}
+
+	row := db.QueryRowContext(r.Context(), `SELECT COUNT(*) FROM compliance_events;`)
```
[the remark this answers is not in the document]
mprahl: There aren't any filters yet, just sort options. The next PR will add the filter options and this query also gets adjusted with those.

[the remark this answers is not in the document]
JustinKuli: Oh, right :D

controllers/complianceeventsapi/server.go
yiraeChristineKim: Unit test added?

mprahl: Sorry, I forgot to `git add` the new test file.

yiraeChristineKim (approved):

PR #162, opened 2024-01-15
Add filtering options to the compliance events API

mprahl:
Relates:
https://issues.redhat.com/browse/ACM-6871

### DIFF
--- a/controllers/complianceeventsapi/server.go
+++ b/controllers/complianceeventsapi/server.go
@@ -20,8 +20,9 @@ import (
 	"github.com/lib/pq"
 )
 
-// init dynamically parses the database columns of each struct type to create a mapping of user provided sort options
-// to the equivalent SQL column. ErrInvalidSortOption is also defined with the available sort options to choose from.
+// init dynamically parses the database columns of each struct type to create a mapping of user provided sort/filter
+// options to the equivalent SQL column. ErrInvalidSortOption, ErrInvalidQueryArg, and validQueryArgs are also defined
+// with the available sort/query options to choose from.
 func init() {
 	tableToStruct := map[string]any{
 		"clusters":          Cluster{},
@@ -39,7 +40,7 @@ func init() {
 
 	// ID is a special case since it's displayed at the top-level in the JSON but is actually in the compliance_events
 	// table.
-	sortOptionsToSQL = map[string]string{"id": "compliance_events.id"}
+	queryOptionsToSQL = map[string]string{"id": "compliance_events.id"}
 	sortOptionsKeys := []string{"id"}
 
 	for tableName, tableStruct := range tableToStruct {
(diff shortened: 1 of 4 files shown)


controllers/complianceeventsapi/server.go
```diff
@@ -224,67 +249,123 @@ func splitQueryValue(value string) []string {
...
-		if err != nil || parsed.Page == 0 {
-			return nil, fmt.Errorf("%w: page must be a positive integer", ErrInvalidQueryArg)
+			continue
 		}
-	}
 
-	sortArgs := splitQueryValue(queryArgs.Get("sort"))
+		switch arg {
+		case "direction":
+			directionArg := queryArgs.Get("direction")
```
JustinKuli, line 293: I think all of the others have just used `value` instead of getting it again - was this deliberate?

controllers/complianceeventsapi/server.go
```diff
@@ -454,31 +619,34 @@ func getComplianceEvents(db *sql.DB, w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
-	query := fmt.Sprintf(`%s
+	whereClause, filterValues := getWhereClause(queryArgs)
+
+	query := fmt.Sprintf(`%s%s
```
JustinKuli, line 624: Can we add an example of what the first part of this query looks like? It would've been nice if the WHERE could be in the string below, but I think that isn't possible because that whole part is optional. But as a result, it's slightly harder to see what the query will look like when it's put together.

JustinKuli (commented):
Looks good overall! Just a couple small things

controllers/complianceeventsapi/server.go
```diff
@@ -82,15 +84,38 @@ func init() {
 	ErrInvalidSortOption = fmt.Errorf(
 		"an invalid sort option was provided, choose from: %s", strings.Join(sortOptionsKeys, ", "),
 	)
+
+	// Reuse sortOptionsKeys since it's no longer being used. This overlaps with all the filterable columns.
+	validQueryArgs = sortOptionsKeys
```
yiraeChristineKim, line 89: I don't get it this part. line 75 	
```
sortOptionsKeys = append(sortOptionsKeys, sortOptionsKeys)
			validQueryArgs = append(validQueryArgs, queryOption)
```
`sortOptionsKeys` append `sortOptionsKeys` same as  `validQueryArgs`

but now validQueryArgs is replaced by sortOptionsKeys?

controllers/complianceeventsapi/server.go
```diff
@@ -224,67 +249,123 @@ func splitQueryValue(value string) []string {
...
+	for arg := range queryArgs {
+		valid := false
 
-	if perPageStr == "" {
-		parsed.PerPage = 20
-	} else {
-		var err error
+		for _, validQueryArg := range validQueryArgs {
+			if arg == validQueryArg {
+				valid = true
```
yiraeChristineKim, line 267: Is it ok only one `validQueryArg`  is valid? 
```

		valid := true

		for _, validQueryArg := range validQueryArgs {
			if arg == validQueryArg {
				valid = false
```

I guess this is right... Could you explain this

controllers/complianceeventsapi/server.go
yiraeChristineKim: vote to Justin

mprahl:
The last commit is to fix gosec scan failure in Stolostron.

yiraeChristineKim (approved):

yiraeChristineKim (approved):

PR #143, opened 2023-10-31
Run PolicyAutomation tests separately in CI

JustinKuli:
The PolicyAutomation tests were taking about half of the total e2e test duration. This excludes those tests from the usual `make e2e-test` target, which will accelerate both the CI and local development. Those tests are now run in a separate (concurrent) GH job in the CI workflow, and the test coverage information is merged in another new job.

This also combines the unit tests and linting tasks into one job, which means that a linting or formatting error will not prevent the e2e tests from running.

Refs:
 - https://issues.redhat.com/browse/ACM-8239

PR #163, opened 2024-01-17
Don't skip SonarCloud run on `push`

dhaiducek:
Followup to:
- #143


dhaiducek:
There's no Sonar check here, so this actually doesn't make sense. This will be addressed in:
- https://github.com/stolostron/governance-policy-propagator/pull/537

PR #164, opened 2024-01-23
Add authorization to the compliance event API endpoints

yiraeChristineKim:
I need to ensure that the compliance history can only be written by authorized service accounts and viewed by users with appropriate ACM access. 

**Definition of Done for Engineering Story Owner (Checklist)**
- The read API endpoint is restricted based on a user's "get" access to corresponding ManagedCluster objects. This result will filter the requests returned in the SQL queries.

Ref: https://issues.redhat.com/browse/ACM-6866
Signed-off-by: Yi Rae Kim <yikim@redhat.com>

### DIFF
--- a/.github/workflows/kind.yml
+++ b/.github/workflows/kind.yml
@@ -78,6 +78,11 @@ jobs:
     - name: Checkout Governance Policy Propagator
       uses: actions/checkout@v4
 
+    - name: Set up Go
+      uses: actions/setup-go@v5
+      with:
+        go-version-file: go.mod
+
     - name: Create K8s KinD Cluster - ${{ matrix.kind }}
       env:
         KIND_VERSION: ${{ matrix.kind }}
@@ -148,6 +153,12 @@ jobs:
     steps:
     - name: Checkout Governance Policy Propagator
       uses: actions/checkout@v4
+
+    - name: Set up Go
+      uses: actions/setup-go@v5
+      id: go
+      with:
+        go-version-file: go.mod
 
     - name: Create K8s KinD Cluster - ${{ matrix.kind }}
       env:
--- a/controllers/complianceeventsapi/auth.go
+++ b/controllers/complianceeventsapi/auth.go
@@ -7,38 +7,66 @@ import (
 	"encoding/json"
 	"errors"
 	"net/http"
+	"slices"
 	"strings"
 
+	"github.com/stolostron/rbac-api-utils/pkg/rbac"
 	authzv1 "k8s.io/api/authorization/v1"
 	k8serrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime/schema"
 	apiserverx509 "k8s.io/apiserver/pkg/authentication/request/x509"
 	"k8s.io/apiserver/pkg/server/dynamiccertificates"
 	"k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/rest"
 )
 
+func getManagedClusterRules(userChangedConfig *rest.Config, managedClusterNames []string,
+) (map[string][]string, error) {
+	kclient, err := kubernetes.NewForConfig(userChangedConfig)
+	if err != nil {
+		log.Error(err, "Failed to create a Kubernetes client with the user token")
+
+		return nil, err
+	}
+
+	managedClusterGR := schema.GroupResource{
+		Group:    "cluster.open-cluster-management.io",
+		Resource: "managedclusters",
+	}
+
+	// key is managedClusterName and value is verbs.
+	// ex: {"managed1": ["get"], "managed2": ["list","create"], "managed3": []}
+	return rbac.GetResourceAccess(kclient, managedClusterGR, managedClusterNames, "")
+}
+
+func canGetManagedCluster(userChangedConfig *rest.Config, managedClusterName string,
+) (bool, error) {
+	allRules, err := getManagedClusterRules(userChangedConfig, []string{managedClusterName})
+	if err != nil {
+		return false, err
+	}
+
+	return getAccessByClusterName(allRules, managedClusterName), nil
+}
+
+func getAccessByClusterName(allManagedClusterRules map[string][]string, clusterName string) bool {
+	starRules, ok := allManagedClusterRules["*"]
+	if ok && slices.Contains(starRules, "get") || slices.Contains(starRules, "*") {
+		return true
+	}
+
+	rules, ok := allManagedClusterRules[clusterName]
+	if ok && slices.Contains(rules, "get") || slices.Contains(rules, "*") {
+		return true
+	}
+
+	return false
+}
+
 // parseToken will return the token string in the Authorization header.
 func parseToken(req *http.Request) string {
 	return strings.TrimSpace(strings.TrimPrefix(req.Header.Get("Authorization"), "Bearer"))
-}
-
-// getClientFromToken will generate a Kubernetes client using the input config and token. No authentication data from
-// the input config is used in the generated client.
-func getClientFromToken(cfg *rest.Config, token string) (*kubernetes.Clientset, error) {
-	userConfig := &rest.Config{
-		Host:    cfg.Host,
-		APIPath: cfg.APIPath,
-		TLSClientConfig: rest.TLSClientConfig{
-			CAFile:     cfg.TLSClientConfig.CAFile,
-			CAData:     cfg.TLSClientConfig.CAData,
-			ServerName: cfg.TLSClientConfig.ServerName,
-			Insecure:   cfg.TLSClientConfig.Insecure,
-		},
-		BearerToken: token,
-	}
-
-	return kubernetes.NewForConfig(userConfig)
 }
 
 // canRecordComplianceEvent will perform certificate or token authentication and perform a subject access review to
@@ -104,12 +132,12 @@ func canRecordComplianceEvent(
 		return review.Status.Allowed, nil
 	}
 
-	token := parseToken(req)
-	if token == "" {
-		return false, ErrNotAuthorized
+	userConfig, err := getUserKubeConfig(cfg, req)
+	if err != nil {
+		return false, err
 	}
 
-	userClient, err := getClientFromToken(cfg, token)
+	userClient, err := kubernetes.NewForConfig(userConfig)
 	if err != nil {
 		return false, err
 	}
@@ -135,7 +163,7 @@ func canRecordComplianceEvent(
 		log.V(0).Info(
 			"The user is not authorized to record a compliance event",
 			"cluster", clusterName,
-			"user", getTokenUsername(token),
+			"user", getTokenUsername(userConfig.BearerToken),
 		)
 	}
 
@@ -188,4 +216,26 @@ func getCertAuthenticator(clientAuthCAs []byte) (*apiserverx509.Authenticator, e
 	authenticator := apiserverx509.NewDynamic(p.VerifyOptions, apiserverx509.CommonNameUserConversion)
 
 	return authenticator, nil
+}
+
+func getUserKubeConfig(config *rest.Config, r *http.Request) (*rest.Config, error) {
+	userConfig := &rest.Config{
+		Host:    config.Host,
+		APIPath: config.APIPath,
+		TLSClientConfig: rest.TLSClientConfig{
+			CAFile:     config.TLSClientConfig.CAFile,
+			CAData:     config.TLSClientConfig.CAData,
+			ServerName: config.TLSClientConfig.ServerName,
+			// For testing
+			Insecure: config.TLSClientConfig.Insecure,
+		},
+	}
+
+	userConfig.BearerToken = parseToken(r)
+
+	if userConfig.BearerToken == "" {
+		return nil, ErrNotAuthorized
+	}
+
+	return userConfig, nil
 }
--- a/controllers/complianceeventsapi/server.go
+++ b/controllers/complianceeventsapi/server.go
@@ -15,6 +15,7 @@ import (
 	"net/http"
 	"net/url"
 	"reflect"
+	"slices"
 	"sort"
 	"strconv"
 	"strings"
@@ -128,6 +129,8 @@ var (
 	ErrInvalidQueryArgValue = errors.New("invalid query argument")
 	ErrInvalidQueryArg      error
 	ErrNotAuthorized        = errors.New("not authorized")
+	// The user has no access to any managed cluster
+	ErrNoAccess = errors.New("the user has no access")
 )
 
 type ComplianceAPIServer struct {
@@ -235,7 +238,16 @@ func (s *ComplianceAPIServer) Start(ctx context.Context, serverContext *Complian
 
 		switch r.Method {
 		case http.MethodGet:
-			getComplianceEvents(serverContext.DB, w, r)
+			// To verify each request independently
+			userConfig, err := getUserKubeConfig(s.cfg, r)
+			if err != nil {
+				if errors.Is(err, ErrNotAuthorized) {
+					writeErrMsgJSON(w, "The Authorization header is not set", http.StatusForbidden)
+				}
+
+				return
+			}
+			getComplianceEvents(serverContext.DB, w, r, userConfig)
 		case http.MethodPost:
 			postComplianceEvent(serverContext.DB, s.cfg, authenticatedClient, authenticator, w, r)
 		default:
@@ -261,12 +273,32 @@ func (s *ComplianceAPIServer) Start(ctx context.Context, serverContext *Complian
 			return
 		}
 
-		getSingleComplianceEvent(serverContext.DB, w, r)
+		// To verify each request independently
+		userConfig, err := getUserKubeConfig(s.cfg, r)
+		if err != nil {
+			if errors.Is(err, ErrNotAuthorized) {
+				writeErrMsgJSON(w, "The Authorization header is not set", http.StatusForbidden)
+			}
+
+			return
+		}
+
+		getSingleComplianceEvent(serverContext.DB, w, r, userConfig)
 	})
 
 	mux.HandleFunc("/api/v1/reports/compliance-events", func(w http.ResponseWriter, r *http.Request) {
 		// This header is for error writings
 		w.Header().Set("Content-Type", "application/json")
+
+		// To verify each request independently
+		userConfig, err := getUserKubeConfig(s.cfg, r)
+		if err != nil {
+			if errors.Is(err, ErrNotAuthorized) {
+				writeErrMsgJSON(w, "The Authorization header is not set", http.StatusForbidden)
+			}
+
+			return
+		}
 
 		if r.Method != http.MethodGet {
 			writeErrMsgJSON(w, "Method not allowed", http.StatusMethodNotAllowed)
@@ -274,7 +306,7 @@ func (s *ComplianceAPIServer) Start(ctx context.Context, serverContext *Complian
 			return
 		}
 
-		getComplianceEventsCSV(serverContext.DB, w, r)
+		getComplianceEventsCSV(serverContext.DB, w, r, userConfig)
 	})
 
 	serveErr := make(chan error)
@@ -348,7 +380,9 @@ func splitQueryValue(value string) []string {
 
 // parseQueryArgs will parse the HTTP request's query arguments and convert them to a usable format for constructing
 // the SQL query. All defaults are set and any invalid query arguments result in an error being returned.
-func parseQueryArgs(queryArgs url.Values, isCSV bool) (*queryOptions, error) {
+func parseQueryArgs(ctx context.Context, queryArgs url.Values, db *sql.DB,
+	userConfig *rest.Config, isCSV bool,
+) (*queryOptions, error) {
 	parsed := &queryOptions{
 		Direction:    "desc",
 		Page:         1,
@@ -469,6 +503,95 @@ func parseQueryArgs(queryArgs url.Values, isCSV bool) (*queryOptions, error) {
 			// Standard string filtering
 			parsed.Filters[sqlName] = splitQueryValue(value)
 		}
+	}
+
+	parsed, err := setAuthorizedClusters(ctx, db, parsed, userConfig)
+	if err != nil {
+		// ErrNoAccess needs queryOptions
+		return parsed, err
+	}
+
+	return parsed, nil
+}
+
+// setAuthorizedClusters verifies that if a cluster filter is provided,
+// the user has access to this filter. If no cluster filter is provided,
+// it sets the cluster filter to all managed clusters the user has access to.
+// If the user has no access, then ErrNoAccess is returned.
+func setAuthorizedClusters(ctx context.Context, db *sql.DB, parsed *queryOptions,
+	userConfig *rest.Config,
+) (*queryOptions, error) {
+	unAuthorizedClusters := []string{}
+
+	// Get all managedCluster rules
+	allRules, err := getManagedClusterRules(userConfig, nil)
+	if err != nil {
+		return parsed, err
+	}
+
+	if slices.Contains(allRules["*"], "get") || slices.Contains(allRules["*"], "*") {
+		return parsed, nil
+	}
+
+	tempClusterIDs := parsed.Filters["clusters.cluster_id"]
+	clusterIDs := make([]string, len(tempClusterIDs))
+
+	// Deep copy parsed.Filters["clusters.cluster_id"]
+	copy(clusterIDs, tempClusterIDs)
+
+	// Reset clusters.cluster_id
+	parsed.Filters["clusters.cluster_id"] = []string{}
+	// Convert id to name
+	for _, id := range clusterIDs {
+		clusterName, err := getClusterNameFromID(ctx, db, id)
+		if err != nil {
+			if errors.Is(err, sql.ErrNoRows) {
+				// Filter out invalid cluster IDs from the query
+				continue
+			}
+
+			log.Error(err, "Failed to get cluster name from cluster ID", "ID", id)
+
+			return parsed, err
+		}
+
+		if !getAccessByClusterName(allRules, clusterName) {
+			unAuthorizedClusters = append(unAuthorizedClusters, id)
+		} else {
+			parsed.Filters["clusters.cluster_id"] = append(parsed.Filters["clusters.cluster_id"], id)
+		}
+	}
+
+	parsedClusterNames := parsed.Filters["clusters.name"]
+	for _, clusterName := range parsedClusterNames {
+		if !getAccessByClusterName(allRules, clusterName) {
+			unAuthorizedClusters = append(unAuthorizedClusters, clusterName)
+		}
+	}
+
+	// There is no cluster.id or clusterName arg from the url query
+	// In other words, the user requests all they have access to.
+	if len(clusterIDs) == 0 && len(parsedClusterNames) == 0 {
+		starRules, ok := allRules["*"]
+		if ok && slices.Contains(starRules, "get") || slices.Contains(starRules, "*") {
+			return parsed, nil
+		}
+
+		for mcName := range allRules {
+			// Should have "GET" auth for all managedCluster
+			if getAccessByClusterName(allRules, mcName) {
+				parsed.Filters["clusters.name"] = append(parsed.Filters["clusters.name"], mcName)
+			}
+		}
+
+		if len(parsed.Filters["clusters.name"]) == 0 {
+			return parsed, ErrNoAccess
+		}
+	}
+
+	if len(unAuthorizedClusters) > 0 {
+		return parsed, fmt.Errorf("%w: the following cluster filters are not authorized: %s",
+			ErrNotAuthorized, strings.Join(unAuthorizedClusters, ", "))
 	}
 
 	return parsed, nil
@@ -599,7 +722,9 @@ func scanIntoComplianceEvent(rows Scannable, includeSpec bool) (*ComplianceEvent
 }
 
 // getSingleComplianceEvent handles the GET API endpoint for a single compliance event by ID.
-func getSingleComplianceEvent(db *sql.DB, w http.ResponseWriter, r *http.Request) {
+func getSingleComplianceEvent(db *sql.DB, w http.ResponseWriter,
+	r *http.Request, config *rest.Config,
+) {
 	eventIDStr := strings.TrimPrefix(r.URL.Path, "/api/v1/compliance-events/")
 
 	eventID, err := strconv.ParseUint(eventIDStr, 10, 64)
@@ -613,7 +738,7 @@ func getSingleComplianceEvent(db *sql.DB, w http.ResponseWriter, r *http.Request
 
 	row := db.QueryRowContext(r.Context(), query, eventID)
 	if row.Err() != nil {
-		log.Error(err, "Failed to query for the compliance event", "eventID", eventID)
+		log.Error(row.Err(), "Failed to query for the compliance event", "eventID", eventID)
 		writeErrMsgJSON(w, "Internal Error", http.StatusInternalServerError)
 
 		return
@@ -633,6 +758,22 @@ func getSingleComplianceEvent(db *sql.DB, w http.ResponseWriter, r *http.Request
 		return
 	}
 
+	// Check auth for managedCluster GET verb
+	isAllowed, err := canGetManagedCluster(config, complianceEvent.Cluster.Name)
+	if err != nil {
+		log.Error(err, `Failed to get the "get" authorization for the cluster`,
+			"cluster", complianceEvent.Cluster.Name)
+		writeErrMsgJSON(w, "Internal Error", http.StatusInternalServerError)
+
+		return
+	}
+
+	if !isAllowed {
+		writeErrMsgJSON(w, "Forbidden", http.StatusForbidden)
+
+		return
+	}
+
 	jsonResp, err := json.Marshal(complianceEvent)
 	if err != nil {
 		log.Error(err, "Failed marshal the compliance event", "eventID", eventID)
@@ -644,6 +785,17 @@ func getSingleComplianceEvent(db *sql.DB, w http.ResponseWriter, r *http.Request
 	if _, err = w.Write(jsonResp); err != nil {
 		log.Error(err, "Error writing success response")
 	}
+}
+
+func getClusterNameFromID(ctx context.Context, db *sql.DB, clusterID string) (name string, err error) {
+	err = db.QueryRowContext(ctx,
+		`SELECT name FROM clusters WHERE cluster_id = $1`, clusterID,
+	).Scan(&name)
+	if err != nil {
+		return "", err
+	}
+
+	return name, nil
 }
 
 // getWhereClause will convert the input queryOptions to a WHERE statement and return the filter values for a prepared
@@ -731,10 +883,52 @@ func getWhereClause(options *queryOptions) (string, []any) {
 }
 
 // getComplianceEvents handles the list API endpoint for compliance events.
-func getComplianceEvents(db *sql.DB, w http.ResponseWriter, r *http.Request) {
-	queryArgs, err := parseQueryArgs(r.URL.Query(), false)
+func getComplianceEvents(db *sql.DB, w http.ResponseWriter,
+	r *http.Request, userConfig *rest.Config,
+) {
+	queryArgs, err := parseQueryArgs(r.Context(), r.URL.Query(), db, userConfig, false)
 	if err != nil {
-		writeErrMsgJSON(w, err.Error(), http.StatusBadRequest)
+		if errors.Is(err, ErrNotAuthorized) {
+			writeErrMsgJSON(w, err.Error(), http.StatusForbidden)
+
+			return
+		}
+
+		if errors.Is(err, ErrNoAccess) {
+			response := ListResponse{
+				Data: []ComplianceEvent{},
+				Metadata: metadata{
+					Page:    queryArgs.Page,
+					Pages:   0,
+					PerPage: queryArgs.PerPage,
+					Total:   0,
+				},
+			}
+
+			jsonResp, err := json.Marshal(response)
+			if err != nil {
+				log.Error(err, "Failed to marshal an empty response")
+				writeErrMsgJSON(w, "Internal Error", http.StatusInternalServerError)
+
+				return
+			}
+
+			if _, err = w.Write(jsonResp); err != nil {
+				log.Error(err, "Error writing empty response")
+				writeErrMsgJSON(w, "Internal Error", http.StatusInternalServerError)
+			}
+
+			return
+		}
+
+		if errors.Is(err, ErrInvalidQueryArg) || errors.Is(err, ErrInvalidQueryArgValue) ||
+			errors.Is(err, ErrInvalidSortOption) {
+			writeErrMsgJSON(w, err.Error(), http.StatusBadRequest)
+
+			return
+		}
+
+		writeErrMsgJSON(w, err.Error(), http.StatusInternalServerError)
 
 		return
 	}
@@ -810,6 +1004,9 @@ LEFT JOIN policies ON compliance_events.policy_id = policies.id` + whereClause /
 
 	if _, err = w.Write(jsonResp); err != nil {
 		log.Error(err, "Error writing success response")
+		writeErrMsgJSON(w, "Internal Error", http.StatusInternalServerError)
+
+		return
 	}
 }
 
@@ -962,10 +1159,51 @@ func getComplianceEventsQuery(whereClause string, queryArgs *queryOptions) strin
 	)
 }
 
-func getComplianceEventsCSV(db *sql.DB, w http.ResponseWriter, r *http.Request) {
-	queryArgs, err := parseQueryArgs(r.URL.Query(), true)
-	if err != nil {
-		writeErrMsgJSON(w, err.Error(), http.StatusBadRequest)
+func setCSVResponseHeaders(w http.ResponseWriter) {
+	w.Header().Set("Content-Disposition", "attachment; filename=reports.csv")
+	w.Header().Set("Content-Type", "text/csv")
+	// It's going to be divided into chunks. if the user don't get it all at once,
+	// the user can receive one by one in the meantime
+	w.Header().Set("Transfer-Encoding", "chunked")
+}
+
+func getComplianceEventsCSV(db *sql.DB, w http.ResponseWriter, r *http.Request,
+	userConfig *rest.Config,
+) {
+	var writer *csv.Writer
+
+	queryArgs, queryArgsErr := parseQueryArgs(r.Context(), r.URL.Query(), db, userConfig, true)
+	if queryArgs != nil {
+		headers := getCsvHeader(queryArgs.IncludeSpec)
+
+		writer = csv.NewWriter(w)
+
+		err := writer.Write(headers)
+		if err != nil {
+			log.Error(err, "Failed to write csv header")
+			writeErrMsgJSON(w, "Internal Error", http.StatusInternalServerError)
+
+			return
+		}
+	}
+
+	if queryArgsErr != nil {
+		if errors.Is(queryArgsErr, ErrNoAccess) {
+			setCSVResponseHeaders(w)
+
+			writer.Flush()
+
+			return
+		}
+
+		if errors.Is(queryArgsErr, ErrInvalidQueryArg) || errors.Is(queryArgsErr, ErrInvalidQueryArgValue) ||
+			errors.Is(queryArgsErr, ErrInvalidSortOption) {
+			writeErrMsgJSON(w, queryArgsErr.Error(), http.StatusBadRequest)
+
+			return
+		}
+
+		writeErrMsgJSON(w, queryArgsErr.Error(), http.StatusInternalServerError)
 
 		return
 	}
@@ -989,18 +1227,6 @@ func getComplianceEventsCSV(db *sql.DB, w http.ResponseWriter, r *http.Request) 
 
 	defer rows.Close()
 
-	headers := getCsvHeader(queryArgs.IncludeSpec)
-
-	writer := csv.NewWriter(w)
-
-	err = writer.Write(headers)
-	if err != nil {
-		log.Error(err, "Failed to write csv header")
-		writeErrMsgJSON(w, "Internal Error", http.StatusInternalServerError)
-
-		return
-	}
-
 	for rows.Next() {
 		ce, err := scanIntoComplianceEvent(rows, queryArgs.IncludeSpec)
 		if err != nil {
@@ -1021,11 +1247,7 @@ func getComplianceEventsCSV(db *sql.DB, w http.ResponseWriter, r *http.Request) 
 		}
 	}
 
-	w.Header().Set("Content-Disposition", "attachment; filename=reports.csv")
-	w.Header().Set("Content-Type", "text/csv")
-	// It's going to be divided into chunks. if the user don't get it all at once,
-	// the user can receive one by one in the meantime
-	w.Header().Set("Transfer-Encoding", "chunked")
+	setCSVResponseHeaders(w)
 
 	writer.Flush()
 }
--- a/go.mod
+++ b/go.mod
@@ -68,6 +68,7 @@ require (
 	github.com/prometheus/procfs v0.12.0 // indirect
 	github.com/shopspring/decimal v1.3.1 // indirect
 	github.com/spf13/cast v1.5.1 // indirect
+	github.com/stolostron/rbac-api-utils v0.0.0-20230213163759-159deac7d398
 	go.uber.org/atomic v1.11.0 // indirect
 	go.uber.org/multierr v1.11.0 // indirect
 	go.uber.org/zap v1.26.0 // indirect
--- a/test/e2e/case18_compliance_api_test.go
+++ b/test/e2e/case18_compliance_api_test.go
@@ -19,11 +19,14 @@ import (
 	"github.com/lib/pq"
 	. "github.com/onsi/ginkgo/v2"
 	. "github.com/onsi/gomega"
+	v1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/rest"
 	ctrllog "sigs.k8s.io/controller-runtime/pkg/log"
 
 	"open-cluster-management.io/governance-policy-propagator/controllers/complianceeventsapi"
+	"open-cluster-management.io/governance-policy-propagator/test/utils"
 )
 
 const eventsEndpoint = "http://localhost:8385/api/v1/compliance-events"
@@ -34,6 +37,17 @@ var httpClient = http.Client{
 		TLSClientConfig: &tls.Config{InsecureSkipVerify: true},
 	},
 }
+
+var (
+	token         string
+	wrongSAToken  string
+	subsetSAToken string
+)
+
+const (
+	wrongSAYaml  = "../resources/case18_compliance_api_test/wrong_service_account.yaml"
+	subsetSAYaml = "../resources/case18_compliance_api_test/subset_service_account.yaml"
+)
 
 func getTableNames(db *sql.DB) ([]string, error) {
 	tableNameRows, err := db.Query("SELECT tablename FROM pg_tables WHERE schemaname = current_schema()")
@@ -130,6 +144,14 @@ var _ = Describe("Test the compliance events API", Label("compliance-events-api"
 		})
 
 		Expect(err).ToNot(HaveOccurred())
+
+		By("Add a new wrong-service account")
+		utils.Kubectl("apply", "-f", wrongSAYaml, "--kubeconfig="+kubeconfigHub)
+		utils.Kubectl("apply", "-f", subsetSAYaml, "--kubeconfig="+kubeconfigHub)
+
+		token = k8sConfig.BearerToken
+		wrongSAToken = getToken(ctx, "default", "wrong-sa")
+		subsetSAToken = getToken(ctx, "default", "subset-sa")
 	})
 
 	Describe("Test the database migrations", func() {
@@ -152,8 +174,8 @@ var _ = Describe("Test the compliance events API", Label("compliance-events-api"
 		Describe("POST one valid event with including all the optional fields", func() {
 			payload := []byte(`{
 				"cluster": {
-					"name": "cluster1",
-					"cluster_id": "test1-cluster1-fake-uuid-1"
+					"name": "managed1",
+					"cluster_id": "test1-managed1-fake-uuid-1"
 				},
 				"parent_policy": {
 					"name": "etcd-encryption1",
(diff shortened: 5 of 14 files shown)


controllers/complianceeventsapi/server.go
```diff
@@ -671,6 +702,20 @@ OFFSET %d ROWS;`,
...
+			log.Error(err, "Failed to get Read Auth for "+ce.Cluster.Name)
+			writeErrMsgJSON(w, "Internal Error", http.StatusInternalServerError)
+
+			break
+		}
+
+		if !isAuth {
+			log.Error(err, "Do not have the Read authorization for "+ce.Cluster.Name)
+
+			continue
```
yiraeChristineKim, line 716: Continue or Break?

go.mod
```diff
@@ -15,6 +15,7 @@ require (
 	github.com/stolostron/go-log-utils v0.1.2
 	github.com/stolostron/go-template-utils/v4 v4.0.1-0.20231212190701-4dc096ec1b40
 	github.com/stolostron/kubernetes-dependency-watches v0.5.2-0.20231212185913-628ab39622b8
+	gopkg.in/square/go-jose.v2 v2.6.0
```
yiraeChristineKim, line 18: jwt

controllers/complianceeventsapi/server.go
```diff
@@ -738,6 +789,20 @@ func postComplianceEvent(db *sql.DB, w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
+	authorized, err := getWriteAuth(config, r.Header.Get("Authorization"), reqEvent.Cluster.Name)
```
mprahl, line 792: Could you pass the context from `r.Context()` to this method for API requests it makes instead of using `context.TODO()`? Same for the get API request.

controllers/complianceeventsapi/server.go
```diff
@@ -613,7 +638,13 @@ func getWhereClause(options *queryOptions) (string, []any) {
 }
 
 // getComplianceEvents handles the list API endpoint for compliance events.
-func getComplianceEvents(db *sql.DB, w http.ResponseWriter, r *http.Request) {
+func getComplianceEvents(db *sql.DB, w http.ResponseWriter, r *http.Request, config *rest.Config) {
+	if r.Header.Get("Authorization") == "" {
+		writeErrMsgJSON(w, "Could not read Authorization", http.StatusForbidden)
```
mprahl, line 643:
```suggestion
		writeErrMsgJSON(w, "The Authorization header is not set", http.StatusForbidden)
```

controllers/complianceeventsapi/server.go
```diff
@@ -671,6 +702,20 @@ OFFSET %d ROWS;`,
 			return
 		}
 
+		isAuth, err := getReadAuth(config, r.Header.Get("Authorization"), ce.Cluster.Name)
```
mprahl, line 705: This isn't quite right. On the "list" API endpoint, you'll need to find the clusters the user has access to and modify the `where` clause so that it's like `WHERE clusters.name  IN ('cluster1', 'cluster2', ...)` unless the user has `*` access in `resourceNames` for the `ManagedCluster` objects, then there is no filter. You'll also need to validate any `cluster.name` filters.

The thing I'm unsure about is how to handle `cluster.cluster_id` filters. I'm guessing for now, we just don't allow filtering by `cluster_id` unless the user has `*` access as mentioned above.

yiraeChristineKim: Oh oh I see!!

mprahl: For example, you'd call this:
```
curl -k -s -XPOST --data-binary '{"kind": "SelfSubjectRulesReview", "apiVersion": "authorization.k8s.io/v1", "spec": {"namespace": "does-not-exist-but-is-requred"}}'  -H "Content-Type: application/json" -H "Accept: application/json" -H "Authorization: Bearer $(oc whoami --show-token)" "$(oc whoami --show-server)/apis/authorization.k8s.io/v1/selfsubjectrulesreviews" | jq .
```

Then you would do similar filters to https://github.com/stolostron/rbac-api-utils/blob/159deac7d398c3470ae730c3fbb8ffbbcca1ac5a/pkg/rbac/rbac.go#L187.

controllers/complianceeventsapi/auth.go
```diff
@@ -0,0 +1,162 @@
...
+
+func getWriteAuth(ctx context.Context, config *rest.Config, token, managedClusterName string,
+) (bool, error) {
+	saName := getServiceAccountName(token)
+	if saName == "" {
+		return false, fmt.Errorf("failed to parse token")
+	}
+
+	ssar := &authorizationv1.SubjectAccessReview{
+		Spec: authorizationv1.SubjectAccessReviewSpec{
```
yiraeChristineKim, line 125: Is this correct? do we want to check replicated policies in managedcluster namespace?

.github/workflows/kind.yml
```diff
@@ -108,7 +108,7 @@ jobs:
     - name: E2E Tests for Compliance Events API
       run: |
         make postgres
-        KUBECONFIG=${PWD}/kubeconfig_hub make e2e-test-coverage-compliance-events-api
+        make e2e-test-coverage-compliance-events-api
```
yiraeChristineKim, line 111: I remove KUBECONFIG=${PWD}/kubeconfig_hub this. Because I use impersonate for auth. With KUBECONFIG=${PWD}/kubeconfig_hub, e2e-test is not working properly

yiraeChristineKim: But..! is there any other way?

controllers/complianceeventsapi/utils.go
```diff
@@ -0,0 +1,11 @@
+package complianceeventsapi
+
+func ContainRules(rules []string, input string) bool {
```
yiraeChristineKim, line 3: I wanted to use `slice.contains` but our go version is 1.20

controllers/complianceeventsapi/auth.go
```diff
@@ -0,0 +1,166 @@
...
+	}
+
+	kclient, err := kubernetes.NewForConfig(config)
+	if err != nil {
+		log.Error(err, "Failed to get clientSet")
+
+		return false, err
+	}
+
+	response, err := kclient.AuthorizationV1().SelfSubjectAccessReviews().Create(
```
yiraeChristineKim, line 149: FYI, SelfSubjectAccessReviews doesn't need roles but SubjectAccessReviews need `CREATE` verb role

.github/workflows/kind.yml
JustinKuli: It seems like github.com/stolostron/rbac-api-utils might also need `impersonate` permissions, so it might be hard to remove that...

controllers/complianceeventsapi/utils.go
JustinKuli: Just bumped to 1.21, does that help?

JustinKuli (changes requested):
I think we need to figure out the service account stuff for the tests.

yiraeChristineKim:
> I think we need to figure out the service account stuff for the tests.

SubjectAccessReview needs clusterrole or role so if the user doesn't have SubjectAccessReview **GET** role. That user will fail. I am not sure which one is better

yiraeChristineKim:
Hi all, since I opened https://github.com/open-cluster-management-io/governance-policy-propagator/pull/167 after 167 merged that I will work on this 164 PR Thanks

JustinKuli:
> > I think we need to figure out the service account stuff for the tests.
> 
> SubjectAccessReview needs clusterrole or role so if the user doesn't have SubjectAccessReview **GET** role. That user will fail. I am not sure which one is better. Also, I should use `SelfSubjectRulesReview`(github.com/stolostron/rbac-api-utils) to get all roles at once. That needs to impersonate. Otherwise, like using SubjectAccessReview, I have to ping kubernetes API per a line. Do you have a good idea?

Limiting the number of API calls is a good reason to favor `SelfSubjectRulesReview`, I wasn't thinking about that.

controllers/complianceeventsapi/auth.go
```diff
@@ -0,0 +1,165 @@
...
+
+func getManagedClusterRules(config *rest.Config,
+	token string, managedClusterNames []string,
+) (map[string][]string, error) {
+	saName := getServiceAccountName(token)
+	if saName == "" {
+		return nil, fmt.Errorf("failed to parse token")
+	}
+	// Token include Bearer prefix
+	config.Impersonate = rest.ImpersonationConfig{
```
mprahl, line 70: We don't want to use impersonation because then group membership does not get taken into account.

We want to instead use the user token to make an SelfSubjectAccessReview.

For example, see how the Kubernetes client is generated using the input token:
https://github.com/stolostron/rbac-api-utils/blob/159deac7d398c3470ae730c3fbb8ffbbcca1ac5a/pkg/rbac/rbac.go#L91-L104

Then you can call `GetResourceAccess` with the Kubernetes client.

yiraeChristineKim: Thanks!

controllers/complianceeventsapi/auth.go
```diff
@@ -0,0 +1,165 @@
...
+
+	return false, nil
+}
+
+func getAllManagedClusterRules(ctx context.Context, config *rest.Config,
+	token string, client client.Client,
+) (map[string][]string, error) {
+	mangedClusterList := &clusterv1.ManagedClusterList{}
+
+	err := client.List(ctx, mangedClusterList)
```
mprahl, line 47: This is not required. It's better to call `GetResourceAccess` without specifying a value for `resourcenames`. Then if the user has access to a subset of clusters, then a list of cluster names is returned. If the user is an admin/has access to all clusters, then `*` is returned. In the case of `*` you don't need to set any filters in the where clause.

mprahl: On second thought, if the user provides a cluster.name or cluster.cluster_id filter, then you'd set `resourcenames` to the filter. Then the returned list will be the subset that they requested.

controllers/complianceeventsapi/server.go
```diff
@@ -528,17 +558,82 @@ func getSingleComplianceEvent(db *sql.DB, w http.ResponseWriter, r *http.Request
...
+
+				rules, ok := allManagedClusterRules[clusterName]
+				if !ok {
+					continue
+				}
+
+				if slices.Contains(rules, "get") || slices.Contains(rules, "*") {
+					filterClusterNames = append(filterClusterNames, "'"+clusterName+"'")
+				} else {
+					log.Info("The service account does not have GET authorization for " + clusterName)
```
mprahl, line 608: Could you make this a debug log? This could be quite noisy.

controllers/complianceeventsapi/server.go
```diff
@@ -528,17 +558,82 @@ func getSingleComplianceEvent(db *sql.DB, w http.ResponseWriter, r *http.Request
...
+		// "cluster" change to "clusters" at queryOptionsToSQL
+		if sqlColumn == "clusters.name" {
+			for _, value := range values {
+				rules, ok := allManagedClusterRules[value]
+				if !ok {
+					continue
+				}
+
+				if slices.Contains(rules, "get") || slices.Contains(rules, "*") {
+					filterClusterNames = append(filterClusterNames, "'"+value+"'")
```
mprahl, line 626: This could be susceptible to SQL injection. It's unlikely now because your current code does validate the cluster names and I don't think Kubernetes would allow such special characters in cluster names, but this check will be removed after my review comments are addressed. Could you please use the prepared statement style of using `$1, $2` and etc like the rest of the code? Then we take out any risk.

controllers/complianceeventsapi/server.go
```diff
@@ -528,17 +558,82 @@ func getSingleComplianceEvent(db *sql.DB, w http.ResponseWriter, r *http.Request
...
+	// "True" means the user added "cluster_id" or "cluster.name" as url filters
+	hasClusterFilter := false
+
 	for sqlColumn, values := range options.Filters {
 		if len(values) == 0 {
 			continue
 		}
 
+		// "cluster" change to "clusters" at queryOptionsToSQL
+		if sqlColumn == "clusters.cluster_id" {
```
mprahl, line 596: I'm thinking that it'd be cleaner if `parseQueryArgs` did the auth check and set the `clusters.cluster_id` or `clusters.name` filter in `options.Filters` with what the user has access to.

Then `getWhereClause` would not need any changes.

controllers/complianceeventsapi/auth.go
```diff
@@ -0,0 +1,165 @@
...
+				Resource:    "policies",
+				Verb:        "patch",
+				Namespace:   managedClusterName,
+				Subresource: "status",
+			},
+		},
+	}
+
+	// Token include Bearer prefix
+	config.Impersonate = rest.ImpersonationConfig{
```
mprahl, line 137: Like my other comment, we just want to use the token directly instead of impersonating.

controllers/complianceeventsapi/server.go
yiraeChristineKim: @mprahl  When the user has only partial auth for clusters,  should we send a message for that?

controllers/complianceeventsapi/server.go
```diff
@@ -145,6 +168,7 @@ func (s *ComplianceAPIServer) Start(ctx context.Context, serverContext *Complian
 		ReadTimeout:  15 * time.Second,
 		WriteTimeout: 15 * time.Second,
 		IdleTimeout:  15 * time.Second,
+		TLSConfig:    &tls.Config{ClientAuth: tls.VerifyClientCertIfGiven, ClientCAs: s.clientAuthCAs},
```
yiraeChristineKim, line 171: Should I remove this too?

mprahl:
@yiraeChristineKim could you please change the PR description since you are only handling the auth/authorization on the get API endpoint in this PR now?

controllers/complianceeventsapi/auth.go
```diff
@@ -0,0 +1,44 @@
...
+		Group:    "cluster.open-cluster-management.io",
+		Resource: "managedclusters",
+	}
+
+	// key is managedClusterName and value is verbs.
+	// ex: {"managed1": ["get"], "managed2": ["list","create"], "managed3": []}
+	return rbac.GetResourceAccess(kclient, managedClusterGVK, managedClusterNames, "")
+}
+
+func getSingleManagedClusterRules(userChangedConfig *rest.Config, managedClusterName string,
```
mprahl, line 31:
```suggestion
func canGetManagedCluster(userChangedConfig *rest.Config, managedClusterName string,
```

controllers/complianceeventsapi/server.go
```diff
@@ -162,7 +169,12 @@ func (s *ComplianceAPIServer) Start(ctx context.Context, serverContext *Complian
 
 		switch r.Method {
 		case http.MethodGet:
-			getComplianceEvents(serverContext.DB, w, r)
+			// To verify each request independently
+			localConfig, err := replaceGetTokenToConfig(s.config, w, r)
```
mprahl, line 173:
```suggestion
			userConfig, err := replaceGetTokenToConfig(s.config, w, r)
```

controllers/complianceeventsapi/server.go
```diff
@@ -223,6 +241,33 @@ func (s *ComplianceAPIServer) Start(ctx context.Context, serverContext *Complian
...
+
+	if r.Header.Get("Authorization") == "" {
+		log.V(1).Error(ErrNotAuthorized, "Token is empty")
+
+		writeErrMsgJSON(w, "The Authorization header is not set", http.StatusForbidden)
+
+		return nil, ErrNotAuthorized
+	}
+
+	localConfig.BearerToken = strings.TrimPrefix(r.Header.Get("Authorization"), "Bearer ")
```
mprahl, line 266:
```suggestion
	localConfig.BearerToken = strings.TrimSpace(strings.TrimPrefix(req.Header.Get("Authorization"), "Bearer "))
```

controllers/complianceeventsapi/server.go
```diff
@@ -223,6 +241,33 @@ func (s *ComplianceAPIServer) Start(ctx context.Context, serverContext *Complian
 	}
 }
 
+func replaceGetTokenToConfig(config *rest.Config, w http.ResponseWriter, r *http.Request) (*rest.Config, error) {
+	// Clean pod's token
+	localConfig := &rest.Config{
```
mprahl, line 246:
```suggestion
	userConfig := &rest.Config{
```

controllers/complianceeventsapi/server.go
```diff
@@ -223,6 +241,33 @@ func (s *ComplianceAPIServer) Start(ctx context.Context, serverContext *Complian
 	}
 }
 
+func replaceGetTokenToConfig(config *rest.Config, w http.ResponseWriter, r *http.Request) (*rest.Config, error) {
+	// Clean pod's token
```
mprahl, line 245:
```suggestion
```

controllers/complianceeventsapi/server.go
```diff
@@ -223,6 +241,33 @@ func (s *ComplianceAPIServer) Start(ctx context.Context, serverContext *Complian
...
+			CAFile:     config.TLSClientConfig.CAFile,
+			CAData:     config.TLSClientConfig.CAData,
+			ServerName: config.TLSClientConfig.ServerName,
+			// For testing
+			Insecure: config.TLSClientConfig.Insecure,
+		},
+	}
+
+	if r.Header.Get("Authorization") == "" {
+		log.V(1).Error(ErrNotAuthorized, "Token is empty")
```
mprahl, line 259: We don't need to log this.
```suggestion
```

controllers/complianceeventsapi/server.go
```diff
@@ -173,6 +185,12 @@ func (s *ComplianceAPIServer) Start(ctx context.Context, serverContext *Complian
 	mux.HandleFunc("/api/v1/compliance-events/", func(w http.ResponseWriter, r *http.Request) {
 		w.Header().Set("Content-Type", "application/json")
 
+		// To verify each request independently
+		localConfig, err := replaceGetTokenToConfig(s.config, w, r)
```
mprahl, line 189:
```suggestion
		userConfig, err := replaceGetTokenToConfig(s.config, w, r)
```

controllers/complianceeventsapi/server.go
```diff
@@ -380,9 +426,108 @@ func parseQueryArgs(queryArgs url.Values) (*queryOptions, error) {
...
+		}
+
+		if slices.Contains(allRules["*"], "get") {
+			authorizedClusterNames = append(authorizedClusterNames, clusterNames...)
+		} else {
+			for _, cn := range clusterNames {
+				if slices.Contains(allRules[cn], "get") || slices.Contains(allRules[cn], "*") {
+					authorizedClusterNames = append(authorizedClusterNames, cn)
+				} else {
+					log.V(1).Info("The service account does not have GET authorization for " + cn)
```
mprahl, line 482:
```suggestion
					log.V(2).Info("The service account does not have GET authorization for " + cn)
```

mprahl, line 482:
```suggestion
					log.V(1).Info("The service account does not have GET authorization for the cluster", "cluster", cn)
```

controllers/complianceeventsapi/server.go
```diff
@@ -539,6 +701,17 @@ func getSingleComplianceEvent(db *sql.DB, w http.ResponseWriter, r *http.Request
 	}
 }
 
+func getClusterNameFromID(ctx context.Context, db *sql.DB, clusterID string) (name string) {
```
mprahl, line 704: This needs to return an error.

controllers/complianceeventsapi/server.go
```diff
@@ -380,9 +426,108 @@ func parseQueryArgs(queryArgs url.Values) (*queryOptions, error) {
...
+
+	parsed.Filters["clusters.name"] = authorizedClusterNames
+
 	return parsed, nil
 }
 
+// Return authorized cluster names or, If the url query doesn't have specific clusters, it verifies
+// all managedCluster Authorizations and it returns true if the user has all authorization for manageClusters
+func getAuthorizedClusterNames(ctx context.Context, db *sql.DB, parsed *queryOptions,
+	userChangedConfig *rest.Config,
```
mprahl, line 458:
```suggestion
	userConfig *rest.Config,
```

controllers/complianceeventsapi/server.go
```diff
@@ -380,9 +426,108 @@ func parseQueryArgs(queryArgs url.Values) (*queryOptions, error) {
...
+	}
+
+	clusterNames, okClusterNames := parsed.Filters["clusters.name"]
+	if okClusterNames {
+		allRules, err := getManagedClusterRules(userChangedConfig, clusterNames)
+		if err != nil {
+			return nil, false, err
+		}
+
+		if slices.Contains(allRules["*"], "get") {
```
mprahl, line 495: This is repeating the logic from above. Could you populate a `clusterNames` slice with all the clusters from the `clusters.cluster_id` and `clusters.name` filter and then check `allRules` against it?

controllers/complianceeventsapi/server.go
```diff
@@ -380,9 +426,108 @@ func parseQueryArgs(queryArgs url.Values) (*queryOptions, error) {
...
+
+	if len(authorizedClusterNames) == 0 {
+		log.V(1).Info("The service account does not have GET authorization for all")
+
+		return nil, ErrNotAuthorized
+	}
+
+	delete(parsed.Filters, "clusters.cluster_id")
+
+	parsed.Filters["clusters.name"] = authorizedClusterNames
```
mprahl, line 450: After talking it over with QE, I think it'd be better if return unauthorized if any of the cluster filters is for a cluster the user does not have access to rather than filter on the subset of what was requested and what they are authorized for.

controllers/complianceeventsapi/server.go
```diff
@@ -380,9 +426,108 @@ func parseQueryArgs(queryArgs url.Values) (*queryOptions, error) {
...
+		return parsed, nil
+	}
+
+	if len(authorizedClusterNames) == 0 {
+		log.V(1).Info("The service account does not have GET authorization for all")
+
+		return nil, ErrNotAuthorized
+	}
+
+	delete(parsed.Filters, "clusters.cluster_id")
```
mprahl, line 448: We actually don't want to do this because even though the authorization has to convert this to a cluster name, the filter should still be on the cluster ID. This is because there could be multiple clusters with the same name but different IDs.

controllers/complianceeventsapi/server.go
```diff
@@ -526,6 +673,21 @@ func getSingleComplianceEvent(db *sql.DB, w http.ResponseWriter, r *http.Request
 		return
 	}
 
+	// Check auth for managedCluster GET verb
+	isAuth, err := getSingleManagedClusterRules(config, complianceEvent.Cluster.Name)
+	if err != nil {
+		log.Error(err, "Failed to get the Read authorization for "+complianceEvent.Cluster.Name)
```
mprahl, line 679:
```suggestion
		log.Error(err, "Failed to get the Read authorization for the cluster", "cluster", complianceEvent.Cluster.Name)
```

controllers/complianceeventsapi/server.go
```diff
@@ -526,6 +673,21 @@ func getSingleComplianceEvent(db *sql.DB, w http.ResponseWriter, r *http.Request
...
+		writeErrMsgJSON(w, "Cannot get read Authorization", http.StatusInternalServerError)
```
mprahl, line 680: We don't want to expose this to the user.
```suggestion
		writeErrMsgJSON(w, "Internal Error", http.StatusInternalServerError)
```

controllers/complianceeventsapi/server.go
```diff
@@ -526,6 +673,21 @@ func getSingleComplianceEvent(db *sql.DB, w http.ResponseWriter, r *http.Request
...
+
+		return
+	}
+
+	if !isAuth {
+		writeErrMsgJSON(w, "Not Authorized", http.StatusForbidden)
```
mprahl, line 686:
```suggestion
		writeErrMsgJSON(w, "Forbidden", http.StatusForbidden)
```

controllers/complianceeventsapi/server.go
yiraeChristineKim: Should we tell users why? "you don't have auths for a subset of clusters" like this? That could make users confused.

test/e2e/case19_rep_policy_placement_test.go
```diff
@@ -230,12 +230,15 @@ var _ = Describe("Test replicated_policy controller and propagation", Ordered, S
...
 		Expect(plc2).ToNot(BeNil())
-		afterString = utils.GetMetrics("controller_runtime_reconcile_total",
-			`controller=\"replicated-policy\"`,
-			`,result=\"success\"`,
-		)[0]
-		afterTotal, err = strconv.Atoi(afterString)
-		Expect(err).ShouldNot(HaveOccurred())
-		Expect(afterTotal - beforeTotal).Should(Equal(2))
+
+		Eventually(func(g Gomega) {
```
yiraeChristineKim, line 234: Flacky issue

controllers/complianceeventsapi/auth.go
```diff
@@ -7,17 +7,54 @@ import (
...
 	"k8s.io/apiserver/pkg/server/dynamiccertificates"
 	"k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/rest"
 )
 
+func getManagedClusterRules(userChangedConfig *rest.Config, managedClusterNames []string,
+) (map[string][]string, error) {
+	kclient, err := kubernetes.NewForConfig(userChangedConfig)
+	if err != nil {
+		log.Error(err, "Failed to get clientSet")
```
mprahl, line 28:
```suggestion
		log.Error(err, "Failed to create a Kubernetes client with the user token")
```

controllers/complianceeventsapi/auth.go
```diff
@@ -7,17 +7,54 @@ import (
...
+
+		return nil, err
+	}
+
+	managedClusterGVK := schema.GroupResource{
```
mprahl, line 33:
```suggestion
	managedClusterGR := schema.GroupResource{
```

controllers/complianceeventsapi/server.go
```diff
@@ -115,6 +120,8 @@ func init() {
 	ErrInvalidQueryArg = fmt.Errorf(
 		"an invalid query argument was provided, choose from: %s", strings.Join(validQueryArgs, ", "),
 	)
+
+	utilruntime.Must(clusterv1.AddToScheme(scheme.Scheme))
```
mprahl, line 124: You need to put this in an `init` like in `main.go`, or else if this starts failing, it would cause the pod to crash during the execution instead of at startup.

controllers/complianceeventsapi/server.go
```diff
@@ -221,6 +240,12 @@ func (s *ComplianceAPIServer) Start(ctx context.Context, serverContext *Complian
 	mux.HandleFunc("/api/v1/compliance-events/", func(w http.ResponseWriter, r *http.Request) {
 		w.Header().Set("Content-Type", "application/json")
 
+		// To verify each request independently
+		userConfig, err := getUserKubeConfig(s.cfg, w, r)
```
mprahl, line 244: Could you move this to right before `getSingleComplianceEvent`? I'd rather avoid doing work if the database is offline or the method is invalid.

controllers/complianceeventsapi/server.go
```diff
@@ -284,6 +315,30 @@ func (s *ComplianceAPIServer) Start(ctx context.Context, serverContext *Complian
...
+		TLSClientConfig: rest.TLSClientConfig{
+			CAFile:     config.TLSClientConfig.CAFile,
+			CAData:     config.TLSClientConfig.CAData,
+			ServerName: config.TLSClientConfig.ServerName,
+			// For testing
+			Insecure: config.TLSClientConfig.Insecure,
+		},
+	}
+
+	if r.Header.Get("Authorization") == "" {
```
mprahl, line 331: Could you check this on the result of `parseToken`?

This is to return an unauthorized error in the event the user passes `Authorization: Bearer `.

controllers/complianceeventsapi/server.go
```diff
@@ -446,9 +503,145 @@ func parseQueryArgs(queryArgs url.Values, isCSV bool) (*queryOptions, error) {
...
+		return authorizedClusterIds, authorizedClusterNames, unAuthorizedClusters, false, err
+	}
+
+	clusterIDs, okClusterIds := parsed.Filters["clusters.cluster_id"]
+	if okClusterIds {
+		// Convert id to name
+		for _, id := range clusterIDs {
+			clusterName, err := getClusterNameFromID(ctx, db, id)
+			if err != nil {
+				log.Error(err, "Failed to get cluster name from cluster id :"+id)
```
mprahl, line 558:
```suggestion
				log.Error(err, "Failed to get cluster name from cluster ID", "ID", id)
```

controllers/complianceeventsapi/server.go
```diff
@@ -446,9 +503,145 @@ func parseQueryArgs(queryArgs url.Values, isCSV bool) (*queryOptions, error) {
...
+			// includeAllManagedClusters is true
+			return authorizedClusterIds, authorizedClusterNames, unAuthorizedClusters, true, nil
+		}
+	}
+
+	return authorizedClusterIds, authorizedClusterNames, unAuthorizedClusters, false, nil
+}
+
+func getAccessByClusterName(allManagedClusterRules map[string][]string, clusterName string) bool {
+	// User requested all managed cluster and has all managedCluster Auth
```
mprahl, line 615: This seems like an out of date comment.

controllers/complianceeventsapi/server.go
```diff
@@ -446,9 +503,145 @@ func parseQueryArgs(queryArgs url.Values, isCSV bool) (*queryOptions, error) {
...
+
+	// There is no cluster.id or clusterName arg from the url query
+	// In other words, the user requests all
+	if !okClusterIds && !okClusterNames {
+		startRules, ok := allRules["*"]
+		if ok && slices.Contains(startRules, "get") || slices.Contains(startRules, "*") {
+			return authorizedClusterIds, authorizedClusterNames, unAuthorizedClusters, true, nil
+		}
+
+		managedClusterNames, err := getAllManagedClusterNames(ctx, c)
```
mprahl, line 593: If the user doesn't have `get` access on `*` managed clusters, then just set `parsed.Filters["clusters.name"]` to the managed clusters the user does have explicit `get` access to in `allRules`.

controllers/complianceeventsapi/server.go
```diff
@@ -446,9 +503,145 @@ func parseQueryArgs(queryArgs url.Values, isCSV bool) (*queryOptions, error) {
...
+
+	rules, ok := allManagedClusterRules[clusterName]
+	if ok && slices.Contains(rules, "get") || slices.Contains(rules, "*") {
+		return true
+	}
+
+	return false
+}
+
+func getAllManagedClusterNames(ctx context.Context, c client.Client) ([]string, error) {
```
mprahl, line 629: After following this comment https://github.com/open-cluster-management-io/governance-policy-propagator/pull/164/files#r1491133119, you'll be able to remove this function.

controllers/complianceeventsapi/server.go
```diff
@@ -446,9 +503,145 @@ func parseQueryArgs(queryArgs url.Values, isCSV bool) (*queryOptions, error) {
...
+				authorizedClusterNames = append(authorizedClusterNames, clusterName)
+			} else {
+				unAuthorizedClusters = append(unAuthorizedClusters, clusterName)
+			}
+		}
+	}
+
+	// There is no cluster.id or clusterName arg from the url query
+	// In other words, the user requests all
+	if !okClusterIds && !okClusterNames {
```
mprahl, line 587: Optional: instead of this indent, you could just have this:

```go
if okClusterIds || okClusterNames {
return authorizedClusterIds, authorizedClusterNames, unAuthorizedClusters, false, nil
}
```

yiraeChristineKim: I am using `return authorizedClusterIds, authorizedClusterNames, unAuthorizedClusters, false, nil`, result of this block so

controllers/complianceeventsapi/server.go
yiraeChristineKim: This block is when the user doesn't have parsed.Filters["clusters.name"]

controllers/complianceeventsapi/server.go
```diff
@@ -446,9 +503,145 @@ func parseQueryArgs(queryArgs url.Values, isCSV bool) (*queryOptions, error) {
 		}
 	}
 
+	authorizedClusterIds, authorizedClusterNames,
+		unAuthorizedClusters, includeAllManagedClusters, err := getAuthorizedClusters(ctx,
```
mprahl, line 507: Rather than return all these values, I think it'd be cleaner if `getAuthorizedClusters` directly set the filter values for the clusters and returned an error that included the unauthorized clusters in the text which could be directly used by the calling function of `parseQueryArgs`.

controllers/complianceeventsapi/server.go
```diff
@@ -446,9 +503,145 @@ func parseQueryArgs(queryArgs url.Values, isCSV bool) (*queryOptions, error) {
...
+	}
+
+	if includeAllManagedClusters {
+		// parsed.Filters["clusters.name"] and parsed.Filters["clusters.cluster_id"] are empty
+		return parsed, nil
+	}
+
+	// If there is any unAuthorized Cluster, it should return ErrNotAuthorized
+	if len(unAuthorizedClusters) > 0 {
+		log.V(2).Info("need GET access", "managedClusters", strings.Join(unAuthorizedClusters, ", "))
```
mprahl, line 520: This should be a normal info log and not hidden under the verbosity level of 2 since this is used for audting.

controllers/complianceeventsapi/server.go
```diff
@@ -446,9 +503,145 @@ func parseQueryArgs(queryArgs url.Values, isCSV bool) (*queryOptions, error) {
...
+
+		return nil, fmt.Errorf("need GET access to these managedClusters: %s, error: %w",
```
mprahl, line 522:
```suggestion
		return nil, fmt.Errorf("%w: the following cluster filters are not authorized: %s",
```

controllers/complianceeventsapi/server.go
yiraeChristineKim: `	utilruntime.Must(clusterv1.AddToScheme(scheme.Scheme))` in init function... should I use main scheme?

controllers/complianceeventsapi/server.go
yiraeChristineKim: Is` GET "/api/v1/compliance-events` for all managedcluster ? I thought we needed to verify all `managedcluster` auths. For example: `managedclusters`: m1, m2, m3
but the user has only m1 "GET". Then I thought I should return UnAuthorized

test/e2e/case19_rep_policy_placement_test.go
dhaiducek: I don't think an `Eventually()` resolves this flaky test. Isn't the Flake where it returns `3` instead of `2`?

yiraeChristineKim: yes yes but it resolved after this

yiraeChristineKim: Never seen after this

controllers/complianceeventsapi/auth.go
```diff
@@ -7,17 +7,54 @@ import (
...
+}
+
+func canGetManagedCluster(userChangedConfig *rest.Config, managedClusterName string,
+) (bool, error) {
+	allRules, err := getManagedClusterRules(userChangedConfig, []string{managedClusterName})
+	if err != nil {
+		return false, err
+	}
+
+	if slices.Contains(allRules["*"], "get") ||
```
mprahl, line 50: Do you also need to check `slices.Contains(allRules["*"], "*")`?

controllers/complianceeventsapi/server.go
```diff
@@ -309,6 +329,30 @@ func (s *ComplianceAPIServer) Start(ctx context.Context, serverContext *Complian
 	}
 }
 
+func getUserKubeConfig(config *rest.Config, w http.ResponseWriter, r *http.Request) (*rest.Config, error) {
```
mprahl, line 332: This duplicates `getClientFromToken` a bit. Could you delete `getClientFromToken`  and refactor `canRecordComplianceEvent` to use this new function?

controllers/complianceeventsapi/server.go
```diff
@@ -309,6 +329,30 @@ func (s *ComplianceAPIServer) Start(ctx context.Context, serverContext *Complian
...
+			ServerName: config.TLSClientConfig.ServerName,
+			// For testing
+			Insecure: config.TLSClientConfig.Insecure,
+		},
+	}
+
+	userConfig.BearerToken = parseToken(r)
+
+	if userConfig.BearerToken == "" {
+		writeErrMsgJSON(w, "The Authorization header is not set", http.StatusForbidden)
```
mprahl, line 348: Could you let the calling function do this instead? That would allow you to use this function in `canRecordComplianceEvent`.

controllers/complianceeventsapi/server.go
```diff
@@ -471,9 +517,122 @@ func parseQueryArgs(queryArgs url.Values, isCSV bool) (*queryOptions, error) {
...
+	return modifiedFilters, unAuthorizedClusters, false, nil
+}
+
+func getAccessByClusterName(allManagedClusterRules map[string][]string, clusterName string) bool {
+	startRules, ok := allManagedClusterRules["*"]
+	if ok && slices.Contains(startRules, "get") || slices.Contains(startRules, "*") {
+		return true
+	}
+
+	rules, ok := allManagedClusterRules[clusterName]
```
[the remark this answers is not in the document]
yiraeChristineKim: But other case it loops explicit and "*" both

controllers/complianceeventsapi/server.go
```diff
@@ -471,6 +505,87 @@ func parseQueryArgs(queryArgs url.Values, isCSV bool) (*queryOptions, error) {
...
+	if err != nil {
+		return parsed, err
+	}
+
+	clusterIDs, okClusterIds := parsed.Filters["clusters.cluster_id"]
+	if okClusterIds {
+		// Convert id to name
+		for _, id := range clusterIDs {
+			clusterName, err := getClusterNameFromID(ctx, db, id)
+			if err != nil {
```
mprahl, line 536: This is too generic of a check. I think `getClusterNameFromID` should return an empty string when it can't find the cluster name from the cluster ID. I think the scan will return something like row not found. Then if clusterName is an empty string, it'll come back as unauthorized. If there is an error returned, then it's an internal error (e.g. Database query failed).

controllers/complianceeventsapi/server.go
```diff
@@ -471,6 +505,87 @@ func parseQueryArgs(queryArgs url.Values, isCSV bool) (*queryOptions, error) {
...
+				unAuthorizedClusters = append(unAuthorizedClusters, id)
+
+				continue
+			}
+
+			if getAccessByClusterName(allRules, clusterName) {
+				parsed.Filters["clusters.cluster_id"] = append(parsed.Filters["clusters.cluster_id"], id)
+			} else {
+				unAuthorizedClusters = append(unAuthorizedClusters, id)
+			}
```
mprahl, line 549:
```suggestion
			if !getAccessByClusterName(allRules, clusterName) {
				unAuthorizedClusters = append(unAuthorizedClusters, id)
			}
```

controllers/complianceeventsapi/server.go
```diff
@@ -471,6 +505,87 @@ func parseQueryArgs(queryArgs url.Values, isCSV bool) (*queryOptions, error) {
...
+	}
+
+	parsedClusterNames, okClusterNames := parsed.Filters["clusters.name"]
+	if okClusterNames {
+		for _, clusterName := range parsedClusterNames {
+			if getAccessByClusterName(allRules, clusterName) {
+				parsed.Filters["clusters.name"] = append(parsed.Filters["clusters.name"], clusterName)
+			} else {
+				unAuthorizedClusters = append(unAuthorizedClusters, clusterName)
+			}
```
mprahl, line 560:
```suggestion
			if !getAccessByClusterName(allRules, clusterName) {
				unAuthorizedClusters = append(unAuthorizedClusters, clusterName)
			}
```

test/e2e/case18_compliance_api_test.go
```diff
@@ -1728,12 +1746,15 @@ var _ = Describe("Test the compliance events API", Label("compliance-events-api"
...
+			csvEndpoints := "http://localhost:8385/api/v1/reports/compliance-events"
 			It("should send CSV file in http response", func(ctx context.Context) {
-				endpoints := "http://localhost:8385/api/v1/reports/compliance-events"
-
-				req, err := http.NewRequestWithContext(ctx, http.MethodGet, endpoints, nil)
+				req, err := http.NewRequestWithContext(ctx, http.MethodGet, csvEndpoints, nil)
 				Expect(err).ShouldNot(HaveOccurred())
 
+				// Set auth token
+				token := getToken(ctx)
```
mprahl, line 1755: Could you cache this at the top of the describe so it can be reused for all the tests? It's expensive to keep calling the API.

test/e2e/case18_compliance_api_test.go
```diff
@@ -1962,6 +2051,256 @@ var _ = Describe("Test the compliance events API", Label("compliance-events-api"
 			Expect(err).To(MatchError(ContainSubstring("The compliance event already exists")))
 		})
 	})
+
+	Describe("Test authorization", func() {
+		Describe("Test method Get", func() {
+			It("Should return StatusForbidden when it is empty Authorization", func(ctx context.Context) {
```
mprahl, line 2057: Shouldn't it be unauthorized when no authentication info is provided?

mprahl, line 2057: Could you also test the CSV endpoint?

yiraeChristineKim: I did line 1802

yiraeChristineKim: It is empty token.. maybe I should change the title

test/e2e/case18_compliance_api_test.go
```diff
@@ -1962,6 +2051,256 @@ var _ = Describe("Test the compliance events API", Label("compliance-events-api"
...
+					err = json.Unmarshal(body, &respJSON)
+					Expect(err).ShouldNot(HaveOccurred())
+
+					rows, ok := respJSON["data"].([]interface{})
+					Expect(ok).To(BeTrue())
+
+					By("Should return 0 rows")
+					Expect(rows).Should(BeEmpty())
+				})
+			It("Should return a forbidden error when SA has only managed1 auth",
```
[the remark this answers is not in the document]
yiraeChristineKim: It returns http.StatusForbidden with "not authorized: the following cluster filters are not authorized: managed2, managed3"

controllers/complianceeventsapi/server.go
yiraeChristineKim: How about this 
```
if errors.Is(err, sql.ErrNoRows) {
					// This leads to send forbidden error to the user
					unAuthorizedClusters = append(unAuthorizedClusters, id)
				}
```

controllers/complianceeventsapi/server.go
```diff
@@ -471,6 +505,93 @@ func parseQueryArgs(queryArgs url.Values, isCSV bool) (*queryOptions, error) {
...
+	parsed, err := setAuthorizedClusters(ctx, db, parsed, userConfig)
+	if err != nil {
+		// ErrNoAccess needs queryOptions
+		return parsed, err
+	}
+
+	return parsed, nil
+}
+
+// getAuthorizedClusterNames verifies that if a cluster filter is provided,
```
mprahl, line 517:
```suggestion
// setAuthorizedClusters verifies that if a cluster filter is provided,
```

controllers/complianceeventsapi/server.go
```diff
@@ -471,6 +505,93 @@ func parseQueryArgs(queryArgs url.Values, isCSV bool) (*queryOptions, error) {
...
+// the user has access to this filter. If no cluster filter is provided,
+// it sets the cluster filter to all managed clusters the user has access to.
```
mprahl, line 519:
```suggestion
// it sets the cluster filter to all managed clusters the user has access to.
// If the user has no access, then ErrNoAccess is returned.
```

controllers/complianceeventsapi/server.go
```diff
@@ -471,6 +505,93 @@ func parseQueryArgs(queryArgs url.Values, isCSV bool) (*queryOptions, error) {
...
+	// Deep copy parsed.Filters["clusters.cluster_id"]
+	copy(clusterIDs, tempClusterIDs)
+
+	if okClusterIds {
+		// Convert id to name
+		for index, id := range clusterIDs {
+			clusterName, err := getClusterNameFromID(ctx, db, id)
+			if err != nil {
+				if errors.Is(err, sql.ErrNoRows) {
+					// This leads to send forbidden error to the user
```
mprahl, line 543:
```suggestion
					// Filter out invalid cluster IDs from the query
```

controllers/complianceeventsapi/server.go
```diff
@@ -471,6 +505,93 @@ func parseQueryArgs(queryArgs url.Values, isCSV bool) (*queryOptions, error) {
...
+					parsed.Filters["clusters.cluster_id"] = append(parsed.Filters["clusters.cluster_id"][:index],
```
mprahl, line 544: This wouldn't work if there was more than one invalid cluster ID since index no longer matches. My suggestion is traverse `clusterIDs` in reverse so that removing any cluster ID does not impact the current index.

controllers/complianceeventsapi/server.go
```diff
@@ -471,6 +505,93 @@ func parseQueryArgs(queryArgs url.Values, isCSV bool) (*queryOptions, error) {
...
+						parsed.Filters["clusters.cluster_id"][index+1:]...)
+
+					continue
+				}
+
+				log.Error(err, "Failed to get cluster name from cluster ID", "ID", id)
```
mprahl, line 550:
```suggestion
				log.Error(err, "Failed to get cluster name from cluster ID", "clusterID", id)
```

controllers/complianceeventsapi/server.go
```diff
@@ -471,6 +505,93 @@ func parseQueryArgs(queryArgs url.Values, isCSV bool) (*queryOptions, error) {
...
+		return parsed, err
+	}
+
+	tempClusterIDs, okClusterIds := parsed.Filters["clusters.cluster_id"]
+	clusterIDs := make([]string, len(tempClusterIDs))
+
+	// Deep copy parsed.Filters["clusters.cluster_id"]
+	copy(clusterIDs, tempClusterIDs)
+
+	if okClusterIds {
```
mprahl, line 537: This if statement is not needed since ranging over an empty/uninitialized slice will just be skipped.

controllers/complianceeventsapi/server.go
```diff
@@ -471,6 +505,93 @@ func parseQueryArgs(queryArgs url.Values, isCSV bool) (*queryOptions, error) {
...
+			}
+
+			if !getAccessByClusterName(allRules, clusterName) {
+				unAuthorizedClusters = append(unAuthorizedClusters, id)
+			}
+		}
+	}
+
+	parsedClusterNames, okClusterNames := parsed.Filters["clusters.name"]
+	if okClusterNames {
```
mprahl, line 562: This if statement is not needed since ranging over an empty/uninitialized slice will just be skipped.

controllers/complianceeventsapi/server.go
```diff
@@ -471,6 +505,93 @@ func parseQueryArgs(queryArgs url.Values, isCSV bool) (*queryOptions, error) {
...
+		for _, clusterName := range parsedClusterNames {
+			if !getAccessByClusterName(allRules, clusterName) {
+				unAuthorizedClusters = append(unAuthorizedClusters, clusterName)
+			}
+		}
+	}
+
+	// There is no cluster.id or clusterName arg from the url query
+	// In other words, the user requests all
```
mprahl, line 571:
```suggestion
	// In other words, the user requests all they have access to.
```

controllers/complianceeventsapi/server.go
```diff
@@ -471,6 +505,93 @@ func parseQueryArgs(queryArgs url.Values, isCSV bool) (*queryOptions, error) {
...
+func setAuthorizedClusters(ctx context.Context, db *sql.DB, parsed *queryOptions,
+	userConfig *rest.Config,
+) (*queryOptions, error) {
+	unAuthorizedClusters := []string{}
+
+	// Get all managedCluster rules
+	allRules, err := getManagedClusterRules(userConfig, nil)
+	if err != nil {
+		return parsed, err
+	}
```
mprahl, line 529: Here is an optimization for admin users (common):
```suggestion
	}

	if slices.Contains(allRules["*"], "get") || slices.Contains(allRules["*"], "*") {
		return parsed, nil
	}
```

controllers/complianceeventsapi/server.go
yiraeChristineKim: isn't better we reset parsed.Filters["clusters.cluster_id"] then append?

controllers/complianceeventsapi/server.go
```diff
@@ -235,7 +238,16 @@ func (s *ComplianceAPIServer) Start(ctx context.Context, serverContext *Complian
 
 		switch r.Method {
 		case http.MethodGet:
-			getComplianceEvents(serverContext.DB, w, r)
+			// To verify each request independently
+			userConfig, err := getUserKubeConfig(s.cfg, r)
+			if err != nil {
+				if errors.Is(err, ErrNotAuthorized) {
+					writeErrMsgJSON(w, "The Authorization header is not set", http.StatusForbidden)
```
mprahl, line 245:
```suggestion
					writeErrMsgJSON(w, "The Authorization header is not set", http.StatusUnauthorized)
```

controllers/complianceeventsapi/server.go
```diff
@@ -261,20 +273,40 @@ func (s *ComplianceAPIServer) Start(ctx context.Context, serverContext *Complian
 			return
 		}
 
-		getSingleComplianceEvent(serverContext.DB, w, r)
+		// To verify each request independently
+		userConfig, err := getUserKubeConfig(s.cfg, r)
+		if err != nil {
+			if errors.Is(err, ErrNotAuthorized) {
+				writeErrMsgJSON(w, "The Authorization header is not set", http.StatusForbidden)
```
mprahl, line 280:
```suggestion
				writeErrMsgJSON(w, "The Authorization header is not set", http.StatusUnauthorized)
```

controllers/complianceeventsapi/server.go
```diff
@@ -261,20 +273,40 @@ func (s *ComplianceAPIServer) Start(ctx context.Context, serverContext *Complian
...
 
 	mux.HandleFunc("/api/v1/reports/compliance-events", func(w http.ResponseWriter, r *http.Request) {
 		// This header is for error writings
 		w.Header().Set("Content-Type", "application/json")
 
+		// To verify each request independently
+		userConfig, err := getUserKubeConfig(s.cfg, r)
+		if err != nil {
+			if errors.Is(err, ErrNotAuthorized) {
+				writeErrMsgJSON(w, "The Authorization header is not set", http.StatusForbidden)
```
mprahl, line 297:
```suggestion
				writeErrMsgJSON(w, "The Authorization header is not set", http.StatusUnauthorized)
```

controllers/complianceeventsapi/server.go
```diff
@@ -471,6 +505,95 @@ func parseQueryArgs(queryArgs url.Values, isCSV bool) (*queryOptions, error) {
...
+	}
+
+	parsedClusterNames := parsed.Filters["clusters.name"]
+	for _, clusterName := range parsedClusterNames {
+		if !getAccessByClusterName(allRules, clusterName) {
+			unAuthorizedClusters = append(unAuthorizedClusters, clusterName)
+		}
+	}
+
+	// There is no cluster.id or clusterName arg from the url query
```
mprahl, line 572:
```suggestion
	// There is no cluster.cluster_id or cluster.name arg from the url query
```

controllers/complianceeventsapi/server.go
```diff
@@ -471,6 +505,95 @@ func parseQueryArgs(queryArgs url.Values, isCSV bool) (*queryOptions, error) {
...
+	}
+
+	tempClusterIDs := parsed.Filters["clusters.cluster_id"]
+	clusterIDs := make([]string, len(tempClusterIDs))
+
+	// Deep copy parsed.Filters["clusters.cluster_id"]
+	copy(clusterIDs, tempClusterIDs)
+
+	// Reset clusters.cluster_id
+	parsed.Filters["clusters.cluster_id"] = []string{}
```
mprahl, line 543: I think this would work and be more efficient:
```suggestion
	clusterIDs := parsed.Filters["clusters.cluster_id"]

	// Reset clusters.cluster_id
	parsed.Filters["clusters.cluster_id"] = []string{}
```

controllers/complianceeventsapi/server.go
```diff
@@ -471,6 +505,95 @@ func parseQueryArgs(queryArgs url.Values, isCSV bool) (*queryOptions, error) {
...
+		}
+	}
+
+	// There is no cluster.id or clusterName arg from the url query
+	// In other words, the user requests all they have access to.
+	if len(clusterIDs) == 0 && len(parsedClusterNames) == 0 {
+		starRules, ok := allRules["*"]
+		if ok && slices.Contains(starRules, "get") || slices.Contains(starRules, "*") {
+			return parsed, nil
+		}
```
mprahl, line 578: The beginning of the function already checks this.

```suggestion
```

controllers/complianceeventsapi/server.go
```diff
@@ -471,6 +505,95 @@ func parseQueryArgs(queryArgs url.Values, isCSV bool) (*queryOptions, error) {
...
+
+		if len(parsed.Filters["clusters.name"]) == 0 {
+			return parsed, ErrNoAccess
+		}
+	}
+
+	if len(unAuthorizedClusters) > 0 {
+		return parsed, fmt.Errorf("%w: the following cluster filters are not authorized: %s",
+			ErrNotAuthorized, strings.Join(unAuthorizedClusters, ", "))
```
mprahl, line 594: I think we should have a new error called `ErrForbidden` to match the HTTP status codes we return. It's getting a bit confusing to use this for both cases.

mprahl (approved):
All my remaining comments are minor. I'll address them in a separate PR.

controllers/complianceeventsapi/server.go
```diff
@@ -471,6 +505,95 @@ func parseQueryArgs(queryArgs url.Values, isCSV bool) (*queryOptions, error) {
...
+
+	// Reset clusters.cluster_id
+	parsed.Filters["clusters.cluster_id"] = []string{}
+	// Convert id to name
+	for _, id := range clusterIDs {
+		clusterName, err := getClusterNameFromID(ctx, db, id)
+		if err != nil {
+			if errors.Is(err, sql.ErrNoRows) {
+				// Filter out invalid cluster IDs from the query
+				continue
```
mprahl, line 550: After thinking about this, I think this allows a user to bypass all authorization checks and have everything returned if they provide an invalid cluster ID.

I'll fix this in an upcoming PR.

PR #165, opened 2024-01-25
Set the database ID annotations on replicated policies

mprahl:
Note that the second commit enables the compliance API to exposed from the KinD cluster.

When the compliance events API is enabled, the replicated policies will
contain the annotation
policy.open-cluster-management.io/parent-policy-compliance-db-id and
each of its policy-templates entries will have the
policy.open-cluster-management.io/policy-compliance-db-id annotation.

As part of this, resilience to database connection losses and changes
was added. This is done by monitoring the database connection. If the
database connection is down, the replicated policy controller will queue
up reconcile requests to add the database specific annotations if the
answer isn't already cached. Once the database connection is restored,
the queued up reconciles will be triggered.

Note that now everytime a database connection changes, a database
migration is run. This is an idempotent action and it's to catch the
case where the database server has been swapped out or restored to an
older backup which may not have the latest database schema.

Relates:
https://issues.redhat.com/browse/ACM-6889


mprahl:
This is now ready for review.

Makefile
```diff
@@ -280,6 +284,7 @@ e2e-stop-instrumented:
 
 .PHONY: e2e-test-coverage
 e2e-test-coverage: E2E_TEST_ARGS = --json-report=report_e2e.json --output-dir=.
+e2e-test-coverage: E2E_TEST_CODE_ARGS = --compliance-api-port=8385
```
yiraeChristineKim, line 287: what is diff 8385 vs 8384

mprahl: So if you have the Kind cluster running, it exposes the 8384 port so it can't be reused when running `e2e-test-coverage` since the controller is run outside of the cluster.

controllers/common/common.go
```diff
@@ -35,6 +37,35 @@ var (
 	ErrInvalidLabelValue = errors.New("unexpected format of label value")
 )
 
+type GuttedObject struct {
```
mprahl, line 40: FYI: I had to move GuttedObject to prevent an import loop.

controllers/propagator/replicatedpolicy_controller.go
```diff
@@ -317,6 +330,158 @@ func (r *ReplicatedPolicyReconciler) Reconcile(ctx context.Context, request ctrl
...
+				parentPolicyID = dbParentPolicy.KeyID
+
+				r.parentPolicyToID.Store(dbParentPolicy.Key(), dbParentPolicy.KeyID)
+			} else {
+				log.Error(
+					err,
+					"Failed to get the database ID of the parent policy",
+					"namespace", replicatedPolicy.Namespace,
+					"name", replicatedPolicy.Name,
+				)
```
JustinKuli, line 368: I was trying to imagine what the rest of the function does after this, because I was trying to think if it could just queue up the thing, and then return early, which might be cleaner/faster?

It seems like it would skip trying to connect to the database again, and just just uses whatever it might already have in its cache. But when its cache is missing something, it deletes that annotation from the policy. So in a case where the cache is empty because the controller just restarted, and it (for some reason) can't connect to the database yet, it seems like it removes all those annotations? Is that right, is that fully intentional?

JustinKuli (commented):
Does anything reset the cached IDs when the database connection changes? It seems like a migration might lead to those IDs being changed, but maybe I missed something.

controllers/complianceeventsapi/complianceeventsapi_controller.go
```diff
@@ -44,106 +47,285 @@ const (
...
+		if err != nil {
+			log.Error(err, "The cached database secret could not be converted to a typed secret")
+
+			return reconcile.Result{}, nil
 		}
 
-		return reconcile.Result{}, err
+		parsedConnectionURL, err = ParseDBSecret(&typedDBSecret, r.TempDir)
+		if errors.Is(err, ErrInvalidDBSecret) {
+			log.Error(err, "Will retry once the invalid secret is updated")
```
yiraeChristineKim, line 156: where is `return`? it says retry

controllers/complianceeventsapi/complianceeventsapi_controller.go
```diff
@@ -44,106 +47,285 @@ const (
...
-	// r.connectionURL is only ever set if the database migration succeeded previously so if the URL is the same then
-	// skip any further work.
-	if r.connectionURL == parsedConnectionURL {
-		err := r.compAPIServer.Start(r.connectionURL)
-		if err != nil {
-			log.Error(err, "Unable to start dbServer")
+	return reconcile.Result{}, nil
+}
+
+// MonitorDatabaseConnection will check the database connection health every 20 seconds. If healthy, it will migrate
```
yiraeChristineKim, line 204: Curious why 20 secs? not 30 or 1 min

main.go
```diff
@@ -436,21 +426,198 @@ func main() {
...
+		}
+	}
+
+	complianceServerCtx, err := complianceeventsapi.NewComplianceServerCtx(dbConnectionURL)
+	if err == nil {
+		// If the migration failed, MigrateDB will log it and MonitorDatabaseConnection will fix it.
+		err := complianceServerCtx.MigrateDB(ctx, client, controllerNamespace)
 		if err != nil {
-			log.Error(err, "Ignoring since this is technical preview")
+			log.Info("Will periodically retry the migration until it is successful")
```
yiraeChristineKim, line 547: Why it is no` log.Error(err`

controllers/propagator/replicatedpolicy_controller.go
mprahl: I was originally thinking you can't guarantee accuracy if the DB is down and there is no cache entry, but you're right, we can verify the replicated policy didn't change based on the DB unique constraints and reuse the ID from the replicated policy.

I just added this.

mprahl:
@JustinKuli good catch! I'm now clearing the cache after a database migration.

main.go
JustinKuli: In a call, Matt mentioned that if this used `log.Error`, it would give a stack trace kind of thing every time. If the DB is down, this is expected to be called a lot, so then the logs would be extremely noisy with less useful information. Using `log.Info` keeps it shorter.

yiraeChristineKim (approved):

PR #167, opened 2024-02-01
Add an API endpoint to generate a CSV compliance report

yiraeChristineKim:
Generating a report as described in the [design](https://github.com/open-cluster-management-io/enhancements/tree/main/enhancements/sig-policy/98-long-term-compliance-history#generating-a-report).

Ref: https://issues.redhat.com/browse/ACM-6884

### DIFF
--- a/controllers/complianceeventsapi/server.go
+++ b/controllers/complianceeventsapi/server.go
@@ -5,6 +5,7 @@ import (
 	"crypto/tls"
 	"crypto/x509"
 	"database/sql"
+	"encoding/csv"
 	"encoding/json"
 	"errors"
 	"fmt"
@@ -238,6 +239,19 @@ func (s *ComplianceAPIServer) Start(ctx context.Context, serverContext *Complian
 		getSingleComplianceEvent(serverContext.DB, w, r)
 	})
 
+	mux.HandleFunc("/api/v1/reports/compliance-events", func(w http.ResponseWriter, r *http.Request) {
+		// This header is for error writings
+		w.Header().Set("Content-Type", "application/json")
+
+		if r.Method != http.MethodGet {
+			writeErrMsgJSON(w, "Method not allowed", http.StatusMethodNotAllowed)
+
+			return
+		}
+
+		getComplianceEventsCSV(serverContext.DB, w, r)
+	})
+
 	serveErr := make(chan error)
 
 	go func() {
@@ -309,7 +323,7 @@ func splitQueryValue(value string) []string {
 
 // parseQueryArgs will parse the HTTP request's query arguments and convert them to a usable format for constructing
 // the SQL query. All defaults are set and any invalid query arguments result in an error being returned.
-func parseQueryArgs(queryArgs url.Values) (*queryOptions, error) {
+func parseQueryArgs(queryArgs url.Values, isCSV bool) (*queryOptions, error) {
 	parsed := &queryOptions{
 		Direction:    "desc",
 		Page:         1,
@@ -318,6 +332,11 @@ func parseQueryArgs(queryArgs url.Values) (*queryOptions, error) {
 		ArrayFilters: map[string][]string{},
 		Filters:      map[string][]string{},
 		NullFilters:  []string{},
+	}
+
+	// Case return CSV file, default PerPage is 0. Unlimited
+	if isCSV {
+		parsed.PerPage = 0
 	}
 
 	for arg := range queryArgs {
@@ -433,6 +452,17 @@ func parseQueryArgs(queryArgs url.Values) (*queryOptions, error) {
 // generateGetComplianceEventsQuery will return a SELECT query with results ready to be parsed by
 // scanIntoComplianceEvent. The caller is responsible for adding filters to the query.
 func generateGetComplianceEventsQuery(includeSpec bool) string {
+	return fmt.Sprintf(`SELECT %s
+FROM
+  compliance_events
+  LEFT JOIN clusters ON compliance_events.cluster_id = clusters.id
+  LEFT JOIN parent_policies ON compliance_events.parent_policy_id = parent_policies.id
+  LEFT JOIN policies ON compliance_events.policy_id = policies.id`,
+		strings.Join(generateSelectedArgs(includeSpec), ", "),
+	)
+}
+
+func generateSelectedArgs(includeSpec bool) []string {
 	selectArgs := []string{
 		"compliance_events.id",
 		"compliance_events.compliance",
@@ -460,14 +490,19 @@ func generateGetComplianceEventsQuery(includeSpec bool) string {
 		selectArgs = append(selectArgs, "policies.spec")
 	}
 
-	return fmt.Sprintf(`SELECT %s
-FROM
-  compliance_events
-  LEFT JOIN clusters ON compliance_events.cluster_id = clusters.id
-  LEFT JOIN parent_policies ON compliance_events.parent_policy_id = parent_policies.id
-  LEFT JOIN policies ON compliance_events.policy_id = policies.id`,
-		strings.Join(selectArgs, ", "),
-	)
+	return selectArgs
+}
+
+// generate Headers for CSV. "." replace by "_"
+// Example: parent_policies.namespace -> parent_policies_namespace
+func getCsvHeader(includeSpec bool) []string {
+	localSelectArgs := generateSelectedArgs(includeSpec)
+
+	for i, arg := range localSelectArgs {
+		localSelectArgs[i] = strings.ReplaceAll(arg, ".", "_")
+	}
+
+	return localSelectArgs
 }
 
 type Scannable interface {
@@ -672,7 +707,7 @@ func getWhereClause(options *queryOptions) (string, []any) {
 
 // getComplianceEvents handles the list API endpoint for compliance events.
 func getComplianceEvents(db *sql.DB, w http.ResponseWriter, r *http.Request) {
-	queryArgs, err := parseQueryArgs(r.URL.Query())
+	queryArgs, err := parseQueryArgs(r.URL.Query(), false)
 	if err != nil {
 		writeErrMsgJSON(w, err.Error(), http.StatusBadRequest)
 
@@ -682,27 +717,7 @@ func getComplianceEvents(db *sql.DB, w http.ResponseWriter, r *http.Request) {
 	// Note that the where clause could be an empty string if not filters were passed in the query arguments.
 	whereClause, filterValues := getWhereClause(queryArgs)
 
-	// Example query:
-	//   SELECT compliance_events.id, compliance_events.compliance, ...
-	//     FROM compliance_events
-	//   LEFT JOIN clusters ON compliance_events.cluster_id = clusters.id
-	//   LEFT JOIN parent_policies ON compliance_events.parent_policy_id = parent_policies.id
-	//   LEFT JOIN policies ON compliance_events.policy_id = policies.id
-	//   WHERE (policies.name=$1 OR policies.name=$2) AND (policies.kind=$3)
-	//   ORDER BY compliance_events.timestamp desc
-	//   LIMIT 20
-	//   OFFSET 0 ROWS;
-	query := fmt.Sprintf(`%s%s
-ORDER BY %s %s
-LIMIT %d
-OFFSET %d ROWS;`,
-		generateGetComplianceEventsQuery(queryArgs.IncludeSpec),
-		whereClause,
-		strings.Join(queryArgs.Sort, ", "),
-		queryArgs.Direction,
-		queryArgs.PerPage,
-		(queryArgs.Page-1)*queryArgs.PerPage,
-	)
+	query := getComplianceEventsQuery(whereClause, queryArgs)
 
 	rows, err := db.QueryContext(r.Context(), query, filterValues...)
 	if err == nil {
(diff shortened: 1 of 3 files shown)


controllers/complianceeventsapi/server.go
```diff
@@ -802,6 +795,205 @@ func postComplianceEvent(db *sql.DB, w http.ResponseWriter, r *http.Request) {
 	}
 }
 
+func getComplianceEventsQuery(whereClause string, queryArgs *queryOptions) string {
```
yiraeChristineKim, line 798: copy paste

controllers/complianceeventsapi/server.go
```diff
@@ -802,6 +795,204 @@ func postComplianceEvent(db *sql.DB, w http.ResponseWriter, r *http.Request) {
...
+	return fmt.Sprintf(`%s%s
+	ORDER BY %s %s
+	LIMIT %d
+	OFFSET %d ROWS;`,
+		generateGetComplianceEventsQuery(queryArgs.IncludeSpec),
+		whereClause,
+		strings.Join(queryArgs.Sort, ", "),
+		queryArgs.Direction,
+		queryArgs.PerPage,
+		(queryArgs.Page-1)*queryArgs.PerPage,
```
JustinKuli, line 818: I think these will need adjustments for the CSV, because it shouldn't do any pagination for that. I don't know if skipping the LIMIT and OFFSET would be right in all cases, maybe it should still limit it to some number (just so it's not an absolutely massive amount of data coming from the database all at once), but be sure to go through all of the pages.

JustinKuli (commented):
Looks pretty good overall! Easy to follow.

controllers/complianceeventsapi/server.go
```diff
@@ -802,6 +795,204 @@ func postComplianceEvent(db *sql.DB, w http.ResponseWriter, r *http.Request) {
...
+		ce.Event.ReportedBy = &nilString
+	}
+
+	if ce.Policy.Severity == nil {
+		ce.Policy.Severity = &nilString
+	}
+
+	if ce.Policy.Namespace == nil {
+		ce.Policy.Namespace = &nilString
+	}
```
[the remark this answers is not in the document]
yiraeChristineKim: *string cause error in `values := []interface{}{` block... umm is there any better way.. to me it was my best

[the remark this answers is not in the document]
yiraeChristineKim: OH should I loop and check type is *string????!!

controllers/complianceeventsapi/server.go
```diff
@@ -802,6 +795,204 @@ func postComplianceEvent(db *sql.DB, w http.ResponseWriter, r *http.Request) {
...
+	// Example query:
+	//   SELECT compliance_events.id, compliance_events.compliance, ...
+	//     FROM compliance_events
+	//   LEFT JOIN clusters ON compliance_events.cluster_id = clusters.id
+	//   LEFT JOIN parent_policies ON compliance_events.parent_policy_id = parent_policies.id
+	//   LEFT JOIN policies ON compliance_events.policy_id = policies.id
+	//   WHERE (policies.name=$1 OR policies.name=$2) AND (policies.kind=$3)
+	//   ORDER BY compliance_events.timestamp desc
+	//   LIMIT 20
+	//   OFFSET 0 ROWS;
```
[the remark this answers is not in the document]
yiraeChristineKim: Is doc-string readme?

controllers/complianceeventsapi/server.go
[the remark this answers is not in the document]
yiraeChristineKim: ohhooh How about this

controllers/complianceeventsapi/server.go
yiraeChristineKim: page can be omitted by users if they want. but some users maybe need csv file just like `GET` endpoint result?!!  personally, i like this because we give more options to users. I don't know what @mprahl opinion. We can discuss this!

controllers/complianceeventsapi/server.go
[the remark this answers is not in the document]
yiraeChristineKim: He added this comment on purpose........

controllers/complianceeventsapi/server.go
mprahl: @yiraeChristineKim, it's a good point. I think we can accept pagination arguments but default `per_page` to unlimited if not explicitly set by the user for CSV reports where as the list endpoint will still require a maximum `per_page`.

controllers/complianceeventsapi/server.go
```diff
@@ -180,6 +181,18 @@ func (s *complianceAPIServer) Start(dbURL string) error {
 		getSingleComplianceEvent(db, w, r)
 	})
 
+	mux.HandleFunc("/api/v1/reports/compliance-events", func(w http.ResponseWriter, r *http.Request) {
+		w.Header().Set("Content-Type", "application/json")
```
mprahl, line 185: You aren't returning JSON. This can be removed since `getComplianceEventsCSV` sets the correct header.

yiraeChristineKim: This for err, I am setting again near csv writing

controllers/complianceeventsapi/server.go
```diff
@@ -802,6 +795,204 @@ func postComplianceEvent(db *sql.DB, w http.ResponseWriter, r *http.Request) {
...
+		"parent_policies_categories",
+		"parent_policies_controls",
+		"parent_policies_standards",
+		"policies_id",
+		"policies_api_group",
+		"policies_kind",
+		"policies_name",
+		"policies_namespace",
+		"policies_severity",
+		"policies_spec",
```
mprahl, line 870: Does `policies_spec` need to be conditional based on if `queryArgs.IncludeSpec` is set or does the library handle this automatically by dropping this header?

controllers/complianceeventsapi/server.go
```diff
@@ -802,6 +795,204 @@ func postComplianceEvent(db *sql.DB, w http.ResponseWriter, r *http.Request) {
...
+	if err != nil {
+		log.Error(err, "Failed to query for compliance events")
+		writeErrMsgJSON(w, "Internal Error", http.StatusInternalServerError)
+
+		return
+	}
+
+	defer rows.Close()
+
+	titles := []string{
```
mprahl, line 849: Rather than duplicate this, could you move `selectArgs` in `generateGetComplianceEventsQuery` to a separate function? Then in this function, you'd call it and then replace all `.` with `_`. That would make it easier for when we add a new column.

controllers/complianceeventsapi/server.go
yiraeChristineKim: it is going to be nil

yiraeChristineKim: It is going to be "nil"

yiraeChristineKim: Do you want drop the header?

mprahl: Yeah, let's drop the header if `queryArgs.IncludeSpec` is not true.

yiraeChristineKim:
@JustinKuli @mprahl  Ready again Thanks!!!

JustinKuli (approved):
/hold for Matt

LGTM overall! I just had one (almost inconsequential) comment. I didn't check if Matt had any remaining requests.

controllers/complianceeventsapi/server.go
```diff
@@ -409,18 +439,28 @@ func generateGetComplianceEventsQuery(includeSpec bool) string {
...
 
-	return fmt.Sprintf(`SELECT %s
-FROM
-  compliance_events
-  LEFT JOIN clusters ON compliance_events.cluster_id = clusters.id
-  LEFT JOIN parent_policies ON compliance_events.parent_policy_id = parent_policies.id
-  LEFT JOIN policies ON compliance_events.policy_id = policies.id`,
-		strings.Join(selectArgs, ", "),
-	)
+	return localSelectArgs
```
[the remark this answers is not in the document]
mprahl: I agree with @JustinKuli as well. Could you please put it back to the way it was?

controllers/complianceeventsapi/server.go
```diff
@@ -409,18 +439,28 @@ func generateGetComplianceEventsQuery(includeSpec bool) string {
...
+}
+
+// generate Headers for CSV. "." replace by "_"
+// Example: parent_policies.namespace -> parent_policies_namespace
+func getCsvHeader(includeSpec bool) []string {
+	localSelectArgs := generateSelectedArgs(includeSpec)
+
+	for i, arg := range localSelectArgs {
+		localSelectArgs[i] = strings.Replace(arg, ".", "_", -1)
```
mprahl, line 460: Optional:
```suggestion
		localSelectArgs[i] = strings.ReplaceAll(arg, ".", "_")
```

controllers/complianceeventsapi/server_test.go
```diff
@@ -35,3 +36,95 @@ func TestSplitQueryValue(t *testing.T) {
...
+			Metadata:   nil,
+			ReportedBy: &reportBy,
+			Timestamp:  theTime,
+		},
+		Cluster: Cluster{
+			ClusterID: "1111",
+			Name:      "cluster1",
+		},
+		Policy: Policy{
+			KeyID:    0,
```
mprahl, line 61:
```suggestion
			KeyID:    5,
```

controllers/complianceeventsapi/server_test.go
```diff
@@ -35,3 +36,95 @@ func TestSplitQueryValue(t *testing.T) {
...
+			Message:    "event1 message",
+			Metadata:   nil,
+			ReportedBy: &reportBy,
+			Timestamp:  theTime,
+		},
+		Cluster: Cluster{
+			ClusterID: "1111",
+			Name:      "cluster1",
+		},
+		Policy: Policy{
```
mprahl, line 60: Could you not use null/empty values here?

controllers/complianceeventsapi/server_test.go
```diff
@@ -35,3 +36,95 @@ func TestSplitQueryValue(t *testing.T) {
...
+	// Should follow this order
+	// 	"compliance_events_id",
+	// "compliance_events_compliance",
+	// "compliance_events_message",
+	// "compliance_events_metadata",
+	// "compliance_events_reported_by",
+	// "compliance_events_timestamp",
+	// "clusters_cluster_id",
+	// "clusters_name",
+	// "parent_policies_id",
```
mprahl, line 87: This should not return 0 but instead no value if it's not set.

controllers/complianceeventsapi/server.go
```diff
@@ -813,6 +828,201 @@ func postComplianceEvent(db *sql.DB, w http.ResponseWriter, r *http.Request) {
...
+		if vv == nil {
+			return ""
+		}
+
+		return *vv
+	case string:
+		return vv
+	case int32:
+		// All int32 related id
+		if int(vv) == 0 {
```
yiraeChristineKim, line 1002: 0 mean id is empty

yiraeChristineKim:
@mprahl  @JustinKuli  Ready Again! Thanks all

controllers/complianceeventsapi/server.go
```diff
@@ -813,6 +828,201 @@ func postComplianceEvent(db *sql.DB, w http.ResponseWriter, r *http.Request) {
...
+			return "nil"
```
mprahl, line 1003: I think no value is more readable in a spreadsheet.
```suggestion
			return ""
```

controllers/complianceeventsapi/server.go
```diff
@@ -813,6 +828,201 @@ func postComplianceEvent(db *sql.DB, w http.ResponseWriter, r *http.Request) {
...
+		}
+
+		return strconv.Itoa(int(vv))
+	case time.Time:
+		return vv.String()
+	case pq.StringArray:
+		// nil will be []
+		return fmt.Sprintf("%v", vv)
```
mprahl, line 1011: Could you do a strings.Join on this so that you get `val1, val2`. This is more friendly than an array like syntax.

controllers/complianceeventsapi/server.go
```diff
@@ -813,6 +828,201 @@ func postComplianceEvent(db *sql.DB, w http.ResponseWriter, r *http.Request) {
...
+		return strconv.FormatBool(vv)
+	case JSONMap:
+		if vv == nil {
+			return "nil"
+		}
+
+		return fmt.Sprintf("%v", vv)
+	default:
+		// case nil:
+		return "nil"
```
mprahl, line 1022: Shouldn't this be a programming error if we encounter a type we do not know how to convert?

mprahl: I think at least, you could return this:
```suggestion
		return fmt.Sprintf("%v", vv)
```

mprahl (commented):
Very close!

controllers/complianceeventsapi/server.go
yiraeChristineKim: How about 
```
 	case JSONMap:
		if vv == nil {
			return "nil"
		}
```
Since most nil cases change to ""

controllers/complianceeventsapi/server.go
mprahl: Does the CSV writer not escape the commas for you?

controllers/complianceeventsapi/server.go
mprahl: Yeah, I think so.

controllers/complianceeventsapi/server.go
```diff
@@ -813,6 +828,201 @@ func postComplianceEvent(db *sql.DB, w http.ResponseWriter, r *http.Request) {
...
+		// nil will be []
+		return fmt.Sprintf("%v", vv)
+	case bool:
+		return strconv.FormatBool(vv)
+	case JSONMap:
+		if vv == nil {
+			return "nil"
+		}
+
+		return fmt.Sprintf("%v", vv)
```
mprahl, line 1019: Could you make this marshalled JSON instead? I don't think this output would be programatically parsable.

controllers/complianceeventsapi/server.go
yiraeChristineKim: jsonByte, err := json.MarshalIndent(vv, "", "  ") output like this:
json```
"{
  ""test"": true
}"
```

File: controllers/complianceeventsapi/server.go (hunk 15 above)
Comment (yiraeChristineKim) [reply controllers/complianceeventsapi/server.go:1011]: `strings.join` output is like this:
`"cat-3, cat-4"`
which looks good to me

File: controllers/complianceeventsapi/server.go (hunk 16 above)
Comment (yiraeChristineKim) [reply controllers/complianceeventsapi/server.go:1022]: `Error` looks like this:
`,configmaps [common] found in namespace default,`

File: controllers/complianceeventsapi/server.go (hunk 14 above)
Comment (yiraeChristineKim) [reply controllers/complianceeventsapi/server.go]: But in Mac spreadsheet and slack spreadsheet looks fine 
```
{
  "severity": "low",
  "test": "one"
}
```

mprahl (approved):

PR #171, opened 2024-02-05
Remove app decision

dhaiducek:
Remove `appsv1.PlacementDecision` where possible


JustinKuli (changes requested):
I think it's risky to remove a field from a CRD. If there is an instance that has that field set, and then the CRD is updated, I'm not sure if the kubernetes API will like it. Personally, I'd prefer to just mark that field as deprecated, and we can continue to ignore it.

Have you tested updating the CRD (ideally when there is an instance with this field set), to see what happens? If it is handled gracefully, then we might be ok.

dhaiducek:
> I think it's risky to remove a field from a CRD.

Yeah, you're probably right. I've removed the commit.

controllers/propagator/replication.go
```diff
@@ -50,7 +50,7 @@ func (r *ReplicatedPolicyReconciler) buildReplicatedPolicy(ctx context.Context,
 	root.ObjectMeta.DeepCopyInto(&replicated.ObjectMeta)
 	root.Spec.DeepCopyInto(&replicated.Spec)
 	replicated.SetName(replicatedName)
-	replicated.SetNamespace(decision.ClusterNamespace)
+	replicated.SetNamespace(clusterName)
```
gparvin, line 53: Can I in theory use a different namespace than the cluster name with hosted mode?

dhaiducek: I don't think so? Hosted mode still has the concept of ManagedClusters that have the same name and namespace. The difference is that in hosted mode, the config-policy-controller is pointing to a different cluster, and the addon-controller deploys the framework to a custom namespace. The hub cluster (i.e. the propagator) isn't all that affected by hosted mode.

gparvin (approved):

dhaiducek:
@JustinKuli Would you re-review this, please? Your "requested changes" review is currently blocking the merge.

JustinKuli (approved):

PR #174, opened 2024-02-06
Fix spec fetching for ComplianceAPI

dhaiducek:
`NestedStringMap` fetches `map[string]string`. We want `NestMap`, which will fetch `map[string]interface{}`.


mprahl:
This was my mistake, so I'll take over this PR to fix the tests.

dhaiducek:
> This was my mistake, so I'll take over this PR to fix the tests.

Okay--I think I'll put a pause on investigating the flaky E2E until this is merged just in case it is somehow the cause.

PR #175, opened 2024-02-07
Fix spec fetching for Compliance API

mprahl:
This was also incorrectly setting the replicated policy annotations on the policy template's annotations.

Built from #174

### DIFF
--- a/test/e2e/case3_mutation_recovery_test.go
+++ b/test/e2e/case3_mutation_recovery_test.go
@@ -136,10 +136,13 @@ var _ = Describe("Test unexpected policy mutation", func() {
 		rootPlc := utils.GetWithTimeout(
 			clientHubDynamic, gvrPolicy, case3PolicyName, testNamespace, true, defaultTimeoutSeconds,
 		)
-		Eventually(func() interface{} {
+		Eventually(func(g Gomega) interface{} {
 			plc = utils.GetWithTimeout(
 				clientHubDynamic, gvrPolicy, testNamespace+"."+case3PolicyName, "managed2", true, defaultTimeoutSeconds,
 			)
+
+			err := utils.RemovePolicyTemplateDBAnnotations(plc)
+			g.Expect(err).ToNot(HaveOccurred())
 
 			return plc.Object["spec"]
 		}, defaultTimeoutSeconds, 1).Should(utils.SemanticEqual(rootPlc.Object["spec"]))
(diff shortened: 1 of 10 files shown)


test/e2e/case3_mutation_recovery_test.go
```diff
@@ -136,11 +136,13 @@ var _ = Describe("Test unexpected policy mutation", func() {
...
-		Eventually(func() interface{} {
+		Eventually(func(g Gomega) interface{} {
 			plc = utils.GetWithTimeout(
 				clientHubDynamic, gvrPolicy, testNamespace+"."+case3PolicyName, "managed2", true, defaultTimeoutSeconds,
 			)
 
+			utils.RemovePolicyTemplateDBAnnotations(g, plc)
+
 			return plc.Object["spec"]
 		}, defaultTimeoutSeconds, 1).Should(utils.SemanticEqual(rootPlc.Object["spec"]))
```
dhaiducek, line 147: (Optional) This seems so common it could be part of a new `util` function that wraps both functions and returns the spec, something like:
```golang
		Eventually(
			utils.GetCleanPolicy(
				clientHubDynamic, 
				gvrPolicy,
				testNamespace+"."+case3PolicyName,
				"managed2", 
				true, 
				defaultTimeoutSeconds,
			), defaultTimeoutSeconds, 1).Should(utils.SemanticEqual(rootPlc.Object["spec"]))
```
With that, you can return spec and the error and Ginkgo will fail the test accordingly on error: https://onsi.github.io/gomega/#category-2-making-codeeventuallycode-assertions-on-functions

test/utils/utils.go
```diff
@@ -54,6 +56,50 @@ func GeneratePldStatus(
 	return &clusterv1beta1.PlacementDecisionStatus{Decisions: plrDecision}
 }
 
+func RemovePolicyTemplateDBAnnotations(g Gomega, plc *unstructured.Unstructured) {
```
dhaiducek, line 59: (Optional) This should return the error rather than assert on it.
```suggestion
func RemovePolicyTemplateDBAnnotations(g Gomega, plc *unstructured.Unstructured) error {
```

dhaiducek (commented):
LGTM! I have some comments to consider about the test.

test/e2e/case3_mutation_recovery_test.go
mprahl: I decided to skip this suggestion since time is tight.

test/utils/utils.go
mprahl: Fixed in the latest push!

dhaiducek (approved):

PR #177, opened 2024-02-12
Add certificate and token auth and authz on the POST API endpoint

mprahl:
If a certificate is provided, the HTTP server will run over HTTPS which enables certificate authentication.

The authorization requires a user to have patch access to the policy status on the managed cluster namespace of the managed cluster that the compliance event is for.

Relates:
https://issues.redhat.com/browse/ACM-6866

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -174,6 +174,11 @@ postgres: cert-manager
 		--from-literal="host=$(POSTGRES_HOST)" \
 		--from-literal="dbname=ocm-compliance-history" \
 		--from-literal="ca=$$(kubectl -n $(KIND_NAMESPACE) get secret postgres-cert -o json | jq -r '.data["ca.crt"]' | base64 -d)"
+	
+	@echo "Copying the compliance API certificates locally"
+	kubectl -n $(KIND_NAMESPACE) get secret compliance-api-cert -o json | jq -r '.data["tls.crt"]' | base64 -d > dev-tls.crt
+	kubectl -n $(KIND_NAMESPACE) get secret compliance-api-cert -o json | jq -r '.data["ca.crt"]' | base64 -d >> dev-ca.crt
+	kubectl -n $(KIND_NAMESPACE) get secret compliance-api-cert -o json | jq -r '.data["tls.key"]' | base64 -d > dev-tls.key
 
 webhook: cert-manager
 	-kubectl create ns $(KIND_NAMESPACE)
--- a/controllers/complianceeventsapi/auth.go
+++ b/controllers/complianceeventsapi/auth.go
@@ -1,0 +1,190 @@
+// Copyright Contributors to the Open Cluster Management project
+package complianceeventsapi
+
+import (
+	"crypto/x509"
+	"encoding/base64"
+	"encoding/json"
+	"errors"
+	"net/http"
+	"strings"
+
+	authzv1 "k8s.io/api/authorization/v1"
+	k8serrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	apiserverx509 "k8s.io/apiserver/pkg/authentication/request/x509"
+	"k8s.io/apiserver/pkg/server/dynamiccertificates"
+	"k8s.io/client-go/kubernetes"
+	"k8s.io/client-go/rest"
+)
+
+// parseToken will return the token string in the Authorization header.
+func parseToken(req *http.Request) string {
+	return strings.TrimSpace(strings.TrimPrefix(req.Header.Get("Authorization"), "Bearer"))
+}
+
+// getClientFromToken will generate a Kubernetes client using the input config and token. No authentication data from
+// the input config is used in the generated client.
+func getClientFromToken(cfg *rest.Config, token string) (*kubernetes.Clientset, error) {
+	userConfig := &rest.Config{
+		Host:    cfg.Host,
+		APIPath: cfg.APIPath,
+		TLSClientConfig: rest.TLSClientConfig{
+			CAFile:     cfg.TLSClientConfig.CAFile,
+			CAData:     cfg.TLSClientConfig.CAData,
+			ServerName: cfg.TLSClientConfig.ServerName,
+			Insecure:   cfg.TLSClientConfig.Insecure,
+		},
+		BearerToken: token,
+	}
+
+	return kubernetes.NewForConfig(userConfig)
+}
+
+// canRecordComplianceEvent will perform certificate or token authentication and perform a subject access review to
+// ensure the input user has patch access to patch the policy status in the managed cluster namespace. An error is
+// returned if the authorization could not be determined. Note that authenticatedClient and authenticator can be nil
+// if certificate authentication isn't used. If both certificate and token authentication is present, certificate takes
+// precedence.
+func canRecordComplianceEvent(
+	cfg *rest.Config,
+	authenticatedClient *kubernetes.Clientset,
+	authenticator *apiserverx509.Authenticator,
+	clusterName string,
+	req *http.Request,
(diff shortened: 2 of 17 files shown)


Makefile
```diff
@@ -174,6 +174,11 @@ postgres: cert-manager
 		--from-literal="host=$(POSTGRES_HOST)" \
 		--from-literal="dbname=ocm-compliance-history" \
 		--from-literal="ca=$$(kubectl -n $(KIND_NAMESPACE) get secret postgres-cert -o json | jq -r '.data["ca.crt"]' | base64 -d)"
+	
+	@echo "Copying the compliance API certificates locally"
+	kubectl -n $(KIND_NAMESPACE) get secret compliance-api-cert -o json | jq -r '.data["tls.crt"]' | base64 -d > dev-tls.crt
+	kubectl -n $(KIND_NAMESPACE) get secret compliance-api-cert -o json | jq -r '.data["ca.crt"]' | base64 -d >> dev-ca.crt
+	kubectl -n $(KIND_NAMESPACE) get secret compliance-api-cert -o json | jq -r '.data["tls.key"]' | base64 -d > dev-tls.key
```
yiraeChristineKim, line 181: Where this secret is created?

controllers/complianceeventsapi/auth.go
```diff
@@ -0,0 +1,186 @@
...
+			return false, ErrNotAuthorized
+		}
+	} else {
+		allowed = result.Status.Allowed
+	}
+
+	if !allowed {
+		var username string
+
+		userInfo, err := getTokenUserInfo(req.Context(), authenticatedClient, token)
```
yiraeChristineKim, line 139: This is just for log.. do we really need?

controllers/complianceeventsapi/server.go
```diff
@@ -749,6 +802,27 @@ func postComplianceEvent(db *sql.DB, w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
+	allowed, err := canRecordComplianceEvent(cfg, authenticatedClient, authenticator, reqEvent.Cluster.Name, r)
+	if err != nil {
+		if errors.Is(err, ErrNotAuthorized) {
+			writeErrMsgJSON(w, "Unauthorized", http.StatusUnauthorized)
```
yiraeChristineKim, line 808: This error is propagator SA doesn't have "`CREATE`" for SubjectAccessReviews.  I think it is `StatusInternalServerError`.

yiraeChristineKim:
```
There's a problem with 401 Unauthorized, the HTTP status code for authentication errors. And that’s just it: it’s for authentication, not authorization. Receiving a 401 response is the server telling you, “you aren’t authenticated–either not authenticated at all or authenticated incorrectly–but please reauthenticate and try again.” To help you out, it will always include a WWW-Authenticate header that describes how to authenticate.
```
This is what I found. Here isn't about `authentication`. Is it?!!!

Makefile
mprahl: It's defined in build/kind/postgres.yaml using CertManager.

controllers/complianceeventsapi/auth.go
mprahl: I changed it to try parsing the token locally as a best effort. It's important to log unauthorized access attempts for auditing.

controllers/complianceeventsapi/server.go
mprahl: I think it's right. If the user provided credential (token or cert) does not allow us to get to the subject access review call, then it's unauthorized (bad authentication provided). If the subject access review says the user is not allowed then it should be a forbidden status code.

controllers/complianceeventsapi/auth.go
```diff
@@ -0,0 +1,194 @@
...
+		Group:       "policy.open-cluster-management.io",
+		Version:     "v1",
+		Resource:    "policies",
+		Verb:        "patch",
+		Namespace:   clusterName,
+		Subresource: "status",
+	}
+
+	// req.TLS.PeerCertificates will be empty if certificate authentication is not enabled (e.g. endpoint is not HTTPS)
+	if req.TLS != nil && len(req.TLS.PeerCertificates) > 0 {
```
gparvin, line 66: If not HTTPS will an error be returned?

mprahl: If not HTTPS then `req.TLS` will be `nil` so then only token based authentication is supported.

controllers/complianceeventsapi/auth.go
```diff
@@ -0,0 +1,194 @@
...
+
+		return review.Status.Allowed, nil
+	}
+
+	token := parseToken(req)
+	if token == "" {
+		return false, ErrNotAuthorized
+	}
+
+	userClient, err := getClientFromToken(cfg, token)
```
gparvin, line 112: Is the previous HTTPS check for certificate auth and this is handling token auth -- but both are HTTPS connections?

gparvin: Does anything need the HTTP connection?  Can we limit it to HTTPS only?

main.go
```diff
@@ -154,12 +159,21 @@ func main() {
...
+		"The port that the compliance history API will listen on",
+	)
+	pflag.StringVar(
+		&complianceAPICert, "compliance-history-api-cert", "",
+		"The path to the certificate the compliance history API will use for HTTPS (CA cert, if any, concatenated "+
+			"after server cert). If not set, HTTP will be used.",
+	)
+	pflag.StringVar(
+		&complianceAPIKey, "compliance-history-api-key", "",
+		"The path to the private key the compliance history API will use for HTTPS. If not set, HTTP will be used.",
```
gparvin, line 176: Can we make these required?  Basically requiring HTTPS

controllers/complianceeventsapi/server.go
yiraeChristineKim: Also you already wrote `//+kubebuilder:rbac:groups=authorization.k8s.io,resources=subjectaccessreviews,verbs=create` This won't fall into here

controllers/complianceeventsapi/auth.go
```diff
@@ -0,0 +1,190 @@
...
+	"k8s.io/client-go/kubernetes"
+	"k8s.io/client-go/rest"
+)
+
+// parseToken will return the token string in the Authorization header.
+func parseToken(req *http.Request) string {
+	return strings.TrimSpace(strings.TrimPrefix(req.Header.Get("Authorization"), "Bearer"))
+}
+
+// getClientFromToken will generate a Kubernetes client using the input config and token. No authentication data from
```
yiraeChristineKim, line 26: Should we merge this PR first then merge mine? I want to reuse this func

controllers/complianceeventsapi/auth.go
```diff
@@ -0,0 +1,190 @@
...
+			"cluster", clusterName,
+			"user", getTokenUsername(token),
+		)
+	}
+
+	return result.Status.Allowed, nil
+}
+
+// getTokenUsername will parse the token and return the username. If the token is invalid, an empty string is returned.
+func getTokenUsername(token string) string {
```
yiraeChristineKim, line 146: Oh?! This was changed?! it was token obj create?. am I dreaming?

controllers/complianceeventsapi/auth.go
mprahl: I think it's good to have an HTTP option in case you want a proxy to handle the TLS. In ACM, we never will though.

yiraeChristineKim: I agree to both `http` and `https`. We have CSV endpoint too. I expect individual users would use CSV endpoint using Curl

yiraeChristineKim (approved):

PR #182, opened 2024-02-15
Stop trying to connect to the DB when the connection URL is not set

mprahl:
This resulted in a log message every 30 seconds when the DB connection URL was unset.

The second commit filters out connection reset logs from the OpenShift router.


gparvin (approved):
Nice!

PR #185, opened 2024-02-15
Upgrade `controller-gen`

dhaiducek:


dhaiducek:
@mprahl I don't see anything with the update that would create incompatibility with OCP v3.11?

dhaiducek:
CI is passing now.

mprahl (approved):

mprahl:
> @mprahl I don't see anything with the update that would create incompatibility with OCP v3.11?

It looks good to me and this component doesn't need to run on OCP 3.11 since it's hub only. I'm going to merge this since it fixes things for Go 1.22.

PR #186, opened 2024-02-16
Use `HaveOccurred()` for E2E error checking

dhaiducek:
Also adjusts `KubectlWithOutput()` so that the `stderr` message is returned with the error, rather than the unhelpful "exit status 1" system error.

### DIFF
--- a/test/utils/utils.go
+++ b/test/utils/utils.go
@@ -298,11 +298,14 @@ func Kubectl(args ...string) {
 
 // KubectlWithOutput execute kubectl cli and return output and error
 func KubectlWithOutput(args ...string) (string, error) {
-	output, err := exec.Command("kubectl", args...).CombinedOutput()
-	//nolint:forbidigo
-	fmt.Println(string(output))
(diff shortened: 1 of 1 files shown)


JustinKuli (approved):

test/utils/utils.go
```diff
@@ -298,11 +298,14 @@ func Kubectl(args ...string) {
...
 // KubectlWithOutput execute kubectl cli and return output and error
 func KubectlWithOutput(args ...string) (string, error) {
-	output, err := exec.Command("kubectl", args...).CombinedOutput()
-	//nolint:forbidigo
-	fmt.Println(string(output))
+	kubectlCmd := exec.Command("kubectl", args...)
+
+	output, err := kubectlCmd.CombinedOutput()
+	if err != nil {
+		return string(output), fmt.Errorf("error running kubectl command: %s: %s", output, err.Error())
```
dhaiducek, line 305: @JustinKuli Do you think we should put the command being run also or just return the stderr message?

PR #187, opened 2024-02-19
Include command in error from `kubectl` commands

dhaiducek:
Followup to:
- #186

### DIFF
--- a/test/utils/utils.go
+++ b/test/utils/utils.go
@@ -302,10 +302,16 @@ func KubectlWithOutput(args ...string) (string, error) {
 
 	output, err := kubectlCmd.CombinedOutput()
 	if err != nil {
-		return string(output), fmt.Errorf("error running kubectl command: %s: %s", output, err.Error())
+		// Reformat error to include kubectl command and stderr output
+		err = fmt.Errorf(
+			"error running command '%s':\n %s: %s",
+			strings.Join(kubectlCmd.Args, " "),
+			output,
+			err.Error(),
+		)
 	}
 
-	return string(output), nil
+	return string(output), err
 }
 
 // GetMetrics execs into the propagator pod and curls the metrics endpoint, filters


test/utils/utils.go
```diff
@@ -302,10 +302,16 @@ func KubectlWithOutput(args ...string) (string, error) {
 
 	output, err := kubectlCmd.CombinedOutput()
 	if err != nil {
-		return string(output), fmt.Errorf("error running kubectl command: %s: %s", output, err.Error())
+		// Reformat error to include kubectl command and stderr output
+		err = fmt.Errorf(
+			"error running command '%s':\n %s: %s",
+			strings.Join(kubectlCmd.Args, " "),
```
dhaiducek, line 308: Two notes:
- There's a line break in the format string because the output also has a line break, so it separates the message out nicely:
  ```
  error running kubectl command 'kubectl get namespaces':
   Error from server (Forbidden): namespaces is forbidden: User "acm-grc-security" cannot list resource "namespaces" in API group "" at the cluster scope
  : exit status 1
  ```
- I chose `strings.Join()` because `kubectlCmd.String()` includes the path of the binary (among possibly other variations), which would be noisier than necessary and could be different from what was passed in (like if something there also expanded paths in the arguments).

JustinKuli: Can the error be wrapped with `%w` instead of using `%s` with `err.Error()`? It's unlikely to matter much, I'd bet the error type is something generic, but it's just generally what I prefer with `fmt.Errorf`

dhaiducek: I generally prefer that also, but in this case the system errors come with a lot of fluff that I don't think is helpful:
```
    <*exec.ExitError | 0xc000081080>: 
    exit status 1
    {
        ProcessState: {
            pid: 6025,
            status: 256,
            rusage: {
                Utime: {Sec: 0, Usec: 513835},
                Stime: {Sec: 0, Usec: 173975},
                Maxrss: 287568,
                Ixrss: 0,
                Idrss: 0,
                Isrss: 0,
                Minflt: 17688,
                Majflt: 0,
                Nswap: 0,
                Inblock: 0,
                Oublock: 67216,
                Msgsnd: 0,
                Msgrcv: 0,
                Nsignals: 0,
                Nvcsw: 1246,
                Nivcsw: 1683,
            },
        },
        Stderr: nil,
    }
```

dhaiducek: I could be mistaken, but I seem to recall that this always occurs for `exec.ExitError`...

JustinKuli: I think you're right, but I was thinking that was the output of `err.Error()` anyway. If that fluff gets skipped by what you have, then I'm all for it!

JustinKuli (approved):

PR #191, opened 2024-02-20
Remove client certificate authentication

mprahl:
This will not be used and it didn't work on OpenShift because the CAs included in the Kubeconfig don't include the CA that signed the client certificate. This would have required additional code to pull the correct CA.


gparvin (approved):
Thanks for getting this cleaned up and for figuring out all the certificate issues.  This looks good to me.

PR #198, opened 2024-02-26
Reset the database ID cache when applicable

mprahl:
This resets the database ID cache whenever the database connection URL changes or an invalid cache entry is found. This also combines the database ID caches on the propagator reconciler and the compliance history API goroutines.

Relates:
https://issues.redhat.com/browse/ACM-10109

### DIFF
--- a/controllers/complianceeventsapi/server.go
+++ b/controllers/complianceeventsapi/server.go
@@ -116,10 +116,12 @@ func init() {
 	)
 }
 
+const (
+	postgresForeignKeyViolationCode = "23503"
+)
+
 var (
 	clusterKeyCache         sync.Map
-	parentPolicyKeyCache    sync.Map
-	policyKeyCache          sync.Map
 	queryOptionsToSQL       map[string]string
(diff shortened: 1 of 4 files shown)


controllers/complianceeventsapi/server.go
```diff
@@ -1051,15 +1050,37 @@ func postComplianceEvent(db *sql.DB, cfg *rest.Config, w http.ResponseWriter, r
...
 	if err != nil {
 		if errors.Is(err, errDuplicateComplianceEvent) {
 			writeErrMsgJSON(w, "The compliance event already exists", http.StatusConflict)
 
 			return
 		}
 
-		log.Error(err, "error inserting compliance event")
+		pqErr, ok := err.(*pq.Error) //nolint:errorlint
+		if ok && pqErr.Code == "23503" {
```
JustinKuli, line 1062: Would it be helpful to make this magic number a const? It seems something like `postgresForeignKeyViolationCode` might be descriptive.

JustinKuli (approved):

PR #201, opened 2024-02-28
Add additional log details for database errors

mprahl:
Relates:
https://issues.redhat.com/browse/ACM-10180

### DIFF
--- a/controllers/complianceeventsapi/server.go
+++ b/controllers/complianceeventsapi/server.go
@@ -521,7 +521,7 @@ func setAuthorizedClusters(ctx context.Context, db *sql.DB, parsed *queryOptions
 				continue
 			}
 
-			log.Error(err, "Failed to get cluster name from cluster ID", "ID", id)
+			log.Error(err, "Failed to get cluster name from cluster ID", getPqErrKeyVals(err, "ID", id)...)
 
 			return parsed, err
 		}
@@ -719,7 +719,7 @@ func getSingleComplianceEvent(db *sql.DB, w http.ResponseWriter,
 			return
 		}
 
-		log.Error(err, "Failed to unmarshal the database results")
+		log.Error(err, "Failed to unmarshal the database results", getPqErrKeyVals(err)...)
 		writeErrMsgJSON(w, "Internal Error", http.StatusInternalServerError)
 
 		return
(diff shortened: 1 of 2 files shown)


controllers/complianceeventsapi/server.go
```diff
@@ -754,6 +754,26 @@ func getSingleComplianceEvent(db *sql.DB, w http.ResponseWriter,
...
+
+	for unwrappedErr != nil {
+		if pqErr, ok := unwrappedErr.(*pq.Error); ok { //nolint: errorlint
+			return append(
+				[]interface{}{"dbMessage", pqErr.Message, "dbDetail", pqErr.Detail, "dbCode", pqErr.Code},
+				additionalKeyVals...,
+			)
+		}
+
+		unwrappedErr = errors.Unwrap(unwrappedErr)
```
JustinKuli, line 771: This seems like the kind of thing that errors.Is or errors.As would do... but I readily admit being confused by those often. Do neither of those work here?

controllers/complianceeventsapi/server.go
```diff
@@ -754,6 +754,26 @@ func getSingleComplianceEvent(db *sql.DB, w http.ResponseWriter,
 	}
 }
 
+// getPqErrKeyVals is a helper to add additional database error details to a log message. additionalKeyVals is provided
+// as a convenience so that the keys don't need to be explicitly set to interface{} types when using the
+// `getPqErrKeyVals(err, "key1", "val1")...“ syntax.
+func getPqErrKeyVals(err error, additionalKeyVals ...interface{}) []interface{} {
+	unwrappedErr := err
+
+	for unwrappedErr != nil {
```
yiraeChristineKim, line 763: Why this is `for`?

mprahl: The idea is to keep unwrapping until either a pq.Error is found or we've unwrapped all that we could and thefore, nil is returned.

yiraeChristineKim: oh I see so unwrap one by one then log db errors each

yiraeChristineKim (approved):

controllers/complianceeventsapi/server.go
JustinKuli: I think `errors.As` could work as a type conversion: [errors.As](https://pkg.go.dev/errors) "examines the tree of its first argument looking for an error that can be assigned to its second argument, which must be a pointer. If it succeeds, it performs the assignment and returns true."

But I mention it more out of curiosity than necessity.

mprahl: Oh yeah, you're right. I didn't catch that the second argument is a pointer to the type that gets set if the type matches:
https://github.com/open-cluster-management-io/governance-policy-propagator/pull/202

PR #203, opened 2024-02-29
Replace unmaintained release action

dhaiducek:
The `curl` command looks pretty much the same as the action! 😄


dhaiducek:
> I don't know why we had to switch to curl, but I like that we don't rely on creating a temporary file for the changelog. Thanks

We were using an unmaintained action that was on a deprecated version of NodeJS. I think I could have written a raw JavaScript snippet instead here, but I went with `curl` since it was straightforward enough.

PR #219, opened 2024-04-25
Update controller-runtime to 0.17.3

JeffeyL:
Changes required to update `controller-runtime` to 0.17.3 

Ref: https://issues.redhat.com/browse/ACM-10835


go.mod
```diff
@@ -70,12 +70,12 @@ require (
 	go.uber.org/atomic v1.11.0 // indirect
 	go.uber.org/multierr v1.11.0 // indirect
 	go.uber.org/zap v1.26.0 // indirect
-	golang.org/x/crypto v0.14.0 // indirect
+	golang.org/x/crypto v0.21.0 // indirect
 	golang.org/x/exp v0.0.0-20231006140011-7918f672742d // indirect
 	golang.org/x/net v0.22.0 // indirect
```
gparvin, line 75: Can we do a replace on x/net to use at least v0.23.0?

JeffeyL: Should I do this in the other upgrades as well?

gparvin: I am mainly focused on the propagator for that update.  Anywhere else would be a nice to have.

PR #222, opened 2024-05-16
Fix flaky propagator test (case19_rep_policy_placement_test)

yiraeChristineKim:


yiraeChristineKim:
Or I can reset Reconciler before each `It`

mprahl (dismissed):
I'm good with removing this test as I don't see much value in it.

test/e2e/case19_rep_policy_placement_test.go
```diff
@@ -1,241 +0,0 @@
...
-			clientHubDynamic, gvrPolicy, testNamespace+"."+case19PolicyName, "managed2", true, defaultTimeoutSeconds,
-		)
-		Expect(plc2).ToNot(BeNil())
-		afterString = utils.GetMetrics("controller_runtime_reconcile_total",
-			`controller=\"replicated-policy\"`,
-			`,result=\"success\"`,
-		)[0]
-		afterTotal, err = strconv.Atoi(afterString)
-		Expect(err).ShouldNot(HaveOccurred())
-		Expect(afterTotal - beforeTotal).Should(Equal(2))
```
dhaiducek, line 239: This is the one I was thinking could alternatively just be:
```
Expect(afterTotal - beforeTotal).Should(Equal(Or(2,3)))
```

test/e2e/case19_rep_policy_placement_test.go
```diff
@@ -1,241 +0,0 @@
...
-		Expect(plc2).To(BeNil())
-		afterString := utils.GetMetrics("controller_runtime_reconcile_total",
-			`controller=\"replicated-policy\"`,
-			`,result=\"success\"`,
-		)[0]
-		afterTotal, err := strconv.Atoi(afterString)
-		Expect(err).ShouldNot(HaveOccurred())
-		Expect(afterTotal - beforeTotal).Should(Equal(2))
-	})
-	It("should reconcile twice when managed1 and managed2 are created by placementRule", func() {
```
dhaiducek, line 166: I don't have a strong opinion, but should it just be this test node that gets removed?

.gitignore
```diff
@@ -102,3 +102,6 @@ testbin/*
 kubeconfig*
 dev-tls.crt
 dev-tls.key
+dev-ca.crt
+
+__debug_bin*
```
dhaiducek, line 107: nit: trailling newline
```suggestion
__debug_bin*

```

dhaiducek (commented):
I'm also fine with taking it out since the metrics haven't really gotten a lot of attention. It could be interesting if the metrics really take off because of some change and this test detects it, but I think that seems unlikely at this point.

test/e2e/case19_rep_policy_placement_test.go
yiraeChristineKim: It should be 2. because managed1, managed2 added. This happens only "minimum"

test/e2e/case19_rep_policy_placement_test.go
yiraeChristineKim: You mean just remove the test after line 166?

dhaiducek: As an alternative, yes, but @mprahl indicated the tests don't have a lot of value, which I'm fine with, so removing it is also fine with me.

mprahl (approved):

PR #232, opened 2024-07-02
Add flag to disable PlacementRule watches

JustinKuli:
Some deployments will not want to install the PlacementRule CRD, since it is an older API which may eventually be removed. This adds an argument to the propagator which will disable all watches on PlacementRules, allowing the controllers to run in those environments.

In addition, if the flag is not set, the controller will check for the PlacementRule CRD on the cluster, and if not found, disable those watches (which would otherwise fail and prevent the controller from working).

Refs:
 - https://issues.redhat.com/browse/ACM-2609

### DIFF
--- a/test/e2e/case4_unexpected_policy_test.go
+++ b/test/e2e/case4_unexpected_policy_test.go
@@ -11,7 +11,7 @@ import (
 	"open-cluster-management.io/governance-policy-propagator/test/utils"
 )
 
-var _ = Describe("Test unexpect policy handling", func() {
+var _ = Describe("Test unexpect policy handling", Label("non-placement-rule"), func() {
 	const (
 		case4PolicyName string = "case4-test-policy"
 		case4PolicyYaml string = "../resources/case4_unexpected_policy/case4-test-policy.yaml"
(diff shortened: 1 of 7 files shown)


test/e2e/case4_unexpected_policy_test.go
```diff
@@ -11,7 +11,7 @@ import (
 	"open-cluster-management.io/governance-policy-propagator/test/utils"
 )
 
-var _ = Describe("Test unexpect policy handling", func() {
+var _ = Describe("Test unexpect policy handling", Label("non-placement-rule"), func() {
```
yiraeChristineKim, line 14: Everything is good! but  Why is this test in non-placementRule?

JustinKuli: This is the one test I found that didn't use PlacementRule. In the future, I'm hoping we have others (maybe we already do, I didn't look super hard).

For this PR, I just needed at least one test for that suite to make sure that the controller didn't crash when the PlacementRule CRD was no longer present.

mprahl (approved):

PR #236, opened 2024-08-01
Add support for a custom service account for hub templates

mprahl:
A new field of spec.hubTemplateOptions.serviceAccountName can be set to leverage a service account in the root policy namespace to resolve hub templates instead of the restricted hub templating of referencing only objects in the root policy namespace.

This leverages the TokenRequest API to get short-lived tokens of the service accounts and periodically refreshes them before expiration.

Relates:
https://issues.redhat.com/browse/ACM-12790


api/v1/policy_types.go
```diff
@@ -98,6 +102,9 @@ type PolicySpec struct {
 	// PolicyDependencies is a list of dependency objects detailed with extra considerations for
 	// compliance that should be fulfilled before applying the policies to the managed clusters.
 	Dependencies []PolicyDependency `json:"dependencies,omitempty"`
+
+	// HubTemplateOptions changes the default behavior of hub templates.
+	HubTemplateOptions *HubTemplateOptions `json:"hubTemplateOptions,omitempty"`
```
mprahl, line 107: This is a pointer so that the `omitempty` has effect rather than returning an empty struct when `hubTemplateOptions` is unset.

yiraeChristineKim: Do we need to create HubTemplateOptions.ServiceAccountName? because it is only one property.

mprahl: We don't need it but I chose it in case we make other things customizable and I thought a `hubTemplateServiceAccountName` field was too long lol.

yiraeChristineKim: Grouping more pretty and this is good for future maybe

controllers/propagator/replicatedpolicy_controller.go
```diff
@@ -219,45 +227,40 @@ func (r *ReplicatedPolicyReconciler) Reconcile(ctx context.Context, request ctrl
...
-		if watcherErr != nil {
-			log.Error(watcherErr, "Failed to uncache objects related to the replicated policy's templates")
-		}
+	// Any errors to expose to the user are logged and recorded in the processTemplates method. Only retry
+	// the request if it's determined to be a retryable error (i.e. don't retry syntax errors).
+	tmplErr := r.processTemplates(ctx, desiredReplicatedPolicy, decision.Cluster, rootPolicy)
+	if errors.Is(tmplErr, ErrRetryable) {
+		// Return the error if it's retryable, which will utilize controller-runtime's exponential backoff.
+		return reconcile.Result{}, tmplErr
+	} else if errors.Is(tmplErr, ErrSAMissing) {
```
yiraeChristineKim, line 253: `else` is not needed?

mprahl: It's not because template errors that are not retryable must continue so that the replicated policy has the hub template error annotation set.

mprahl:
I'm not sure what's going on with the test. It's failing every time in CI but working every time locally...

controllers/propagator/template_utils.go
```diff
@@ -0,0 +1,641 @@
...
+	// A channel to trigger replicated policy reconciles if an asynchronous error occurs such as refreshing the token
+	// failing.
+	replicatedPolicyUpdates chan event.GenericEvent
+	// globalLock has a write lock if a template resolver is being instantiated or getting cleaned up. All other
+	// operations use a read lock.
+	globalLock sync.RWMutex
+	// The default template resolver when a replicated policy does not specify a service account.
+	defaultTemplateResolver *templates.TemplateResolver
+	// A map of service acconts to template resolvers. Access to this must leverage globalLock.
+	TemplateResolvers map[types.NamespacedName]*templateResolverWithCancel
```
yiraeChristineKim, line 104: types.NamespacedName namespace is always rootPlc's namespace right?

mprahl: For this map, the key is the service account the template resolver uses.

yiraeChristineKim: Oh now I got it rootPlc namespace could vary

test/resources/case9_templates/case9-token-sa.yaml
```diff
@@ -0,0 +1,5 @@
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  name: test-user
+  namespace: default
```
yiraeChristineKim, line 5: Isn't it better this service account has diff namespace from the policy?

yiraeChristineKim: Oh nvm I saw `testNamespace`

test/e2e/case9_templates_test.go
```diff
@@ -517,4 +527,298 @@ var _ = Describe("Test policy templates", func() {
 			utils.ListWithTimeout(clientHubDynamic, gvrPolicy, opt, 0, false, defaultTimeoutSeconds)
 		})
 	})
+
+	Describe("Test a custom service account", Ordered, func() {
+		AfterAll(func(ctx context.Context) {
```
yiraeChristineKim, line 532: Do we need to clean clusterrole, role, SA, and so on in afterAll?

yiraeChristineKim, line 532: Do we need to clean case9SATokenYaml?

controllers/propagator/template_utils.go
mprahl: Yes, exactly.

test/e2e/case9_templates_test.go
mprahl: I think the BeforeAll's DeferCleanUp in "Test token issuance and refresh" covers this. Did I miss something?

yiraeChristineKim: oh no worry I was confused with token-sa

controllers/propagator/template_utils.go
yiraeChristineKim: One more question, should the ServiceAccount be created in the same namespace as the root policy??

mprahl: @yiraeChristineKim that's right. Only service accounts of the root policy namespace can be used.

test/e2e/case9_templates_test.go
```diff
@@ -517,4 +529,308 @@ var _ = Describe("Test policy templates", func() {
...
+			Expect(token).ToNot(BeNil())
+
+			By("Monitoring the token file for a change: " + tokenPath)
+
+			Eventually(func(g Gomega) {
+				refreshedToken, err := os.ReadFile(tokenPath)
+				g.Expect(err).ToNot(HaveOccurred())
+				g.Expect(refreshedToken).ToNot(BeNil())
+				g.Expect(bytes.Equal(token, refreshedToken)).To(BeFalse(), "Expected the token to have been refreshed")
+			}, 70, 3).Should(Succeed())
```
yiraeChristineKim, line 794: Why 70 not 60?

mprahl: Just because it could be exactly 60 seconds before the token refreshes, so I gave it a bit of a buffer. I could have made it 65 seconds though.

yiraeChristineKim: I don't think it is a big deal as long as they change it. I don't think no one complain that it is over 1min

controllers/propagator/template_utils.go
```diff
@@ -0,0 +1,645 @@
...
+		cancel:         resolverCtxCancel,
+		referenceCount: &atomic.Uint32{},
+	}
+
+	t.wg.Add(1)
+
+	go func() {
+		err := dynamicWatcher.Start(resolverCtx)
+
+		rv.canceled.Store(true)
```
yiraeChristineKim, line 215: Why this is located before `err != nil`?

mprahl: `Start` is blocking so if the method returns, it means the dynamic watcher is no longer running regardless of if stopping it was an error or intentional.

controllers/propagator/template_utils.go
```diff
@@ -0,0 +1,645 @@
...
+		}
+
+		t.replicatedPolicyUpdates <- event.GenericEvent{Object: &replicatedPolicy}
+	}
+
+	return nil
+}
+
+// RemovePolicy will clean up watches on the current template resolver (service account used in the last call to
+// GetResolver) and if this was the last policy using this template resolver, the template resolver will be cleaned up.
```
yiraeChristineKim, line 299: root policy or replicated policy? It is confusing. Base on referenceCount description it is a replicated policy right?

mprahl: Yup, it's a replicated policy. I can rename the method or clarify the description. Any preference?

yiraeChristineKim: RemoveRePolicy? I don't know Dale is good at this lol

controllers/propagator/template_utils.go
```diff
@@ -0,0 +1,645 @@
...
+	wg *sync.WaitGroup
+	// A channel to trigger replicated policy reconciles if an asynchronous error occurs such as refreshing the token
+	// failing.
+	replicatedPolicyUpdates chan event.GenericEvent
+	// globalLock has a write lock if a template resolver is being instantiated or getting cleaned up. All other
+	// operations use a read lock.
+	globalLock sync.RWMutex
+	// The default template resolver when a replicated policy does not specify a service account.
+	defaultTemplateResolver *templates.TemplateResolver
+	// A map of service acconts to template resolvers. Access to this must leverage globalLock.
```
yiraeChristineKim, line 103:
```suggestion
	// A map of service acconts to templateResolverWithCancel. Access to this must leverage globalLock.
```
??

yiraeChristineKim: or `template resolver` single

yiraeChristineKim:
This is amazing! Yare are the best expert Go concurrency

controllers/propagator/propagation.go
```diff
@@ -408,7 +432,5 @@ func (r *ReplicatedPolicyReconciler) processTemplates(
 		}
 	}
 
-	log.V(1).Info("Successfully processed templates")
```
mprahl, line 411: This log message is removed because `processTemplates` is now always called and when no templates are present, it does nothing. This is to clean up watches if hub templates used to be present but are now removed.

controllers/propagator/template_utils.go
mprahl: I added a comment to clarify this.

controllers/propagator/template_utils.go
```diff
@@ -0,0 +1,647 @@
...
+	// cancel will stop the TemplateResolver and the automatic token refresh.
+	cancel context.CancelFunc
+	// canceled will be set if resolver failed to start or was stopped. The atomic.Bool type is used to avoid using
+	// a write lock.
+	canceled atomic.Bool
+	// referenceCount is a counter for how many replicated policies rely on this template resolver. Once it is 0,
+	// the template resolver will be cleaned up. A pointer is used since the reference count can be copied to a new
+	// templateResolverWithCancel instance if the TemplateResolver instance was unexpectedly canceled and was
+	// restarted.
+	referenceCount *atomic.Uint32
```
JustinKuli, line 71: I can't shake the feeling that us doing our own reference counting means we're missing some other way to do this

controllers/propagator/template_utils.go
```diff
@@ -0,0 +1,647 @@
...
+func getUserKubeConfig(config *rest.Config, tokenFile string) *rest.Config {
+	userConfig := &rest.Config{
+		Host:    config.Host,
+		APIPath: config.APIPath,
+		TLSClientConfig: rest.TLSClientConfig{
+			CAFile:     config.TLSClientConfig.CAFile,
+			CAData:     config.TLSClientConfig.CAData,
+			ServerName: config.TLSClientConfig.ServerName,
+			// For testing
+			Insecure: config.TLSClientConfig.Insecure,
```
JustinKuli, line 611: This comment makes me think this piece should be removed, but I don't think that's right - this just means that this user config's Insecure setting matches the input rest Config's setting, which seems good even outside of testing.

JustinKuli (commented):
Seems about right :sweat_smile:

controllers/propagator/rootpolicy_setup.go
```diff
@@ -24,6 +24,7 @@ import (
 //+kubebuilder:rbac:groups=cluster.open-cluster-management.io,resources=managedclusters;placementdecisions;placements,verbs=get;list;watch
 //+kubebuilder:rbac:groups=apps.open-cluster-management.io,resources=placementrules,verbs=get;list;watch
 //+kubebuilder:rbac:groups=core,resources=events,verbs=get;list;watch;create;update;patch;delete
+//+kubebuilder:rbac:groups=core,resources=serviceaccounts/token,verbs=create
```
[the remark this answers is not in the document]
mprahl: It's `serviceaccounts/token` to create service account tokens but not service accounts.

controllers/propagator/template_utils.go
```diff
@@ -0,0 +1,647 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/event"
+	"sigs.k8s.io/controller-runtime/pkg/source"
+
+	policiesv1 "open-cluster-management.io/governance-policy-propagator/api/v1"
+	"open-cluster-management.io/governance-policy-propagator/controllers/common"
+)
+
+var (
+	// A special NamespacedName to indicate the default service account is used
+	defaultSANamespacedName types.NamespacedName = types.NamespacedName{Name: "<default>"}
```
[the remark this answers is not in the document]
mprahl: The Go compiler says no :laughing:

controllers/propagator/template_utils.go
mprahl: The only other way I could think of that would work other than periodically checking to see if any root policy still references this service account was to have a map of service accounts to replicated policies. I thought the reference count was better to avoid locking around a shared map. It also saves a bit of memory.

Another way that came to mind was using a read and write lock for each template resolver and all replicated policies have a read lock and then a separate goroutine waits for a write lock, which would indicate there are no more readers left to then clean up the template resolver. The problem is that the read and write lock in the Go standard library doesn't work like that as it prioritizes giving the write lock when it can, which makes sense.

JustinKuli (approved):
Seems good to me.

PR #243, opened 2024-09-20
Update go-template-utils for better permission handling

mprahl:
This will stop the RetryWatcher if the service account used for templates no longer has access to watch a resource.

This required additional changes due to controller-runtime and client-go needing updates.


dhaiducek (approved):
LGTM!

PR #246, opened 2024-10-02
Remove requirement for `eventHook` in automation

dhaiducek:
ref: https://issues.redhat.com/browse/ACM-14657


gparvin (approved):
Thanks! I think this was likely part of an original design that we intended to expand on.

PR #228, opened 2024-06-11
Use controller-runtime to protect metrics endpoint

zyjjay:
Use the new `FilterProvider` and `SecureServing` fields to authenticate and authorize requests to the metrics endpoint

ref: https://issues.redhat.com/browse/ACM-8346

PR #324, opened 2026-06-26
fix: update permissions

dhaiducek:
Followup to:
- #228 
- #301

## Summary by CodeRabbit

* **Bug Fixes**
  * Expanded Kubernetes RBAC permissions for handling events to cover both the core API group (`""`) and `events.k8s.io`.
  * Updated operator, role, and leader-election RBAC so the same event actions remain available (get/list/watch, create, update/patch, and delete).

### DIFF
--- a/controllers/propagator/metric.go
+++ b/controllers/propagator/metric.go
@@ -8,7 +8,8 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/metrics"
 )
 
-//+kubebuilder:rbac:groups=authorization.k8s.io,resources=tokenreviews,verbs=create
+//+kubebuilder:rbac:groups=authentication.k8s.io,resources=tokenreviews,verbs=create
+//+kubebuilder:rbac:groups=authorization.k8s.io,resources=subjectaccessreviews,verbs=create
 
 var (
 	hubTemplateActiveWatchesMetric = prometheus.NewGauge(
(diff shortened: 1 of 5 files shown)


controllers/propagator/metric.go
```diff
@@ -8,7 +8,7 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/metrics"
 )
 
-//+kubebuilder:rbac:groups=authorization.k8s.io,resources=tokenreviews,verbs=create
+//+kubebuilder:rbac:groups=authentication.k8s.io,resources=tokenreviews,verbs=create
```
jan-law, line 11: Does the propagator need `subjectaccessreviews:create` too?

dhaiducek: Hmmm. Good question. Let me look into that.

dhaiducek: Ah, looks like you're right--good catch!

dhaiducek:
Looks like MulticlusterHub already had these permissions from when the `kube-rbac-proxy` container was in place, so this PR just puts them in sync and doesn't need to be backported:
```
- apiGroups:
  - 'authentication.k8s.io'
  resources:
  - 'tokenreviews'
  verbs:
  - create
- apiGroups:
  - authorization.k8s.io
  resources:
  - subjectaccessreviews
  verbs:
  - create
```
https://github.com/stolostron/multiclusterhub-operator/blob/4cf1e08a2de6ea8f4a106559de882e998d773862/pkg/templates/charts/toggle/grc/templates/grc-clusterrole.yaml#L69-L80

jan-law (approved):
LGTM

PR #331, opened 2026-07-29
feat: exclude policies in a policyset

jan-law:
ref: https://redhat.atlassian.net/browse/ACM-37779

Some notes about 5.0 UI compatibility:
- The yaml editor can edit the policyset exclusions - no changes required
- The policyset webhook will silently block invalid edits (invalid policy name, or more than one exclusion entry for the same policyname), even though the UI appears to successfully "Submit" the edit with no errors

Screenshot From 2026-07-30 11-55-30

Design considerations
- If a policyset excludes a cluster with a non-existent name, the Policy and PolicySet status still shows an exclusion for that cluster. I wasn't sure how to verify if a temporarily unreachable cluster was valid. This also means typos in cluster names are not validated.
- If the webhook is not working, the controller will silently merge any duplicate exclusion entries for the same policy, and silently ignore exclusions for a policyName that is not part of the set

Screenshot From 2026-07-30 11-57-11

## Summary by CodeRabbit

* **New Features**
  * Added policy-set exclusions for selected clusters, with optional reasons.
  * Added status reporting for excluded clusters and remaining applicable bindings.
  * Added validation to reject duplicate or unknown policy exclusions.
  * Updated propagation and reconciliation to honor exclusions, including direct policy bindings.
  * Added clearer status messages reflecting active exclusions.

* **Tests**
  * Added unit and end-to-end coverage for exclusion validation, propagation, status reporting, and multi-binding scenarios.


jan-law:
@coderabbitai review

jan-law:
@coderabbitai review

api/v1beta1/policyset_webhook.go
```diff
@@ -0,0 +1,130 @@
...
+			if req != nil {
+				log = log.WithValues("kind", req.Kind, "namespace", req.Namespace, "name", req.Name)
+			}
+
+			return log
+		}).
+		Complete()
+}
+
+// +kubebuilder:webhook:path=/validate-policy-open-cluster-management-io-v1beta1-policyset,mutating=false,failurePolicy=Ignore,sideEffects=None,groups=policy.open-cluster-management.io,resources=policysets,verbs=create;update,versions=v1beta1,name=policyset.open-cluster-management.io.webhook,admissionReviewVersions=v1,serviceName=propagator-webhook-service,serviceNamespace=open-cluster-management
```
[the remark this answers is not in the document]
jan-law: Similar to the policy webhook's failurePolicy of Ignore, there is low operational risk when the webhook fails to validate the policyset exclusions. At most, duplicate exclusion entries for the same policy will be ignored, preferring the first entry. The status may look a little confusing when the exclusions report a policy name that is not in the policy set. Defer to human reviewers.

api/v1beta1/policyset_webhook.go
```diff
@@ -0,0 +1,130 @@
...
+	if len(duplicatePolicyNames) > 0 {
+		errs = append(errs, fmt.Errorf("%w: %s", errDuplicateExclusionPolicy, strings.Join(duplicatePolicyNames, ", ")))
+	}
+
+	if len(invalidPolicyNames) > 0 {
+		errs = append(errs, fmt.Errorf("%w: %s", errInvalidExclusionPolicy, strings.Join(invalidPolicyNames, ", ")))
+	}
+
+	err := errors.Join(errs...)
+	log.Info("Invalid policy set exclusions: " + err.Error())
```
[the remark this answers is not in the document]
jan-law: fixed

jan-law:
@coderabbitai review

jan-law:
@coderabbitai review

controllers/common/common.go
```diff
@@ -410,3 +412,165 @@ func LogConstructor(controllerName string, kind string, req *reconcile.Request)
...
+
+		slices.SortFunc(remainingBindings, func(a, b policiesv1.RemainingBinding) int {
+			return cmp.Compare(a.PlacementBinding, b.PlacementBinding)
+		})
+
+		result[clusterName] = remainingBindings
+	}
+
+	return result, nil
+}
```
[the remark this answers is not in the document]
jan-law: please re-evaluate latest commit

test/e2e/case18_policyset_exclusions_test.go
```diff
@@ -0,0 +1,566 @@
...
+	)
+
+	BeforeAll(func(ctx SpecContext) {
+		bootstrapCase18PolicySet(ctx, case18Yaml, case18PolicySet, case18PolicySetPLD)
+	})
+
+	AfterAll(func(ctx SpecContext) {
+		_, _ = utils.KubectlWithOutput(ctx, "delete",
+			"-f", case18Yaml, "-n", testNamespace, "--kubeconfig="+kubeconfigHub, "--ignore-not-found")
+	})
```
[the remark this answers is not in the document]
jan-law: Don't the webhook tests run at a different time than the regular e2e tests? See https://github.com/open-cluster-management-io/governance-policy-propagator/blob/main/.github/workflows/kind.yml 
```
    - name: E2E Tests
      run: |
        export GOPATH=$(go env GOPATH)
        KUBECONFIG=${PWD}/kubeconfig_hub make e2e-test-coverage
...
    - name: Verify Deployment Configuration
      run: |
        make webhook
        make build-images
        KUBECONFIG=${PWD}/kubeconfig_hub_e2e make kind-deploy-controller-dev
    
    - name: E2E Tests for Webhook
      run: |
        KUBECONFIG=${PWD}/kubeconfig_hub make e2e-test-webhook
```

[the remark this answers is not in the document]
jan-law: See Makefile: https://github.com/open-cluster-management-io/governance-policy-propagator/blob/main/Makefile
```
E2E_LABEL_FILTER = --label-filter="!webhook && !policyautomation"
E2E_JSON = --json-report=report_e2e.json
E2E_JUNIT = --junit-report=report_e2e.xml

.PHONY: e2e-test
e2e-test: e2e-dependencies
	$(GINKGO) -v $(E2E_LABEL_FILTER) --output-dir=. $(E2E_JSON) $(E2E_JUNIT) test/e2e

.PHONY: e2e-test-webhook
e2e-test-webhook: E2E_LABEL_FILTER = --label-filter="webhook"
e2e-test-webhook: E2E_JSON = --json-report=report_e2e_webhook.json
e2e-test-webhook: E2E_JUNIT = --junit-report=report_e2e_webhook.xml
e2e-test-webhook: e2e-test
```

api/v1beta1/policyset_webhook.go
```diff
@@ -0,0 +1,135 @@
...
+	if err != nil {
+		log.Info(
+			"Invalid policy set exclusions",
+			"duplicatePolicyCount", len(duplicatePolicyNames),
+			"invalidPolicyCount", len(invalidPolicyNames),
+		)
+
+		// return the last error for readability
+		return err
+	}
```
JustinKuli, line 133: If both problems happen, we only return info about the invalid ones, is that right? Should we use `errors.Join` (or something similar) to return both? Or is there a reason to only return one?

test/e2e/case18_policyset_exclusions_test.go
```diff
@@ -0,0 +1,566 @@
...
+			g.Expect(clusterStatuses).ToNot(BeEmpty())
+
+			for _, clusterStatus := range clusterStatuses {
+				clusterStatusMap, ok := clusterStatus.(map[string]any)
+				g.Expect(ok).To(BeTrue())
+
+				_, hasRemainingBindings := clusterStatusMap["remainingBindings"]
+				g.Expect(hasRemainingBindings).To(BeFalse())
+			}
+		}, defaultTimeoutSeconds, 1).Should(Succeed())
```
JustinKuli, line 70: From what I see, this verifies that the excluded policy is noted as "excluded" in the root policy status. But I don't think it asserts anything about the compliance state at that point. From what I see, the `managed2` cluster would still show up in `status.status`, but with a blank/unknown compliance.

Is that deliberate? If true, I think it would prevent this policy's root compliance from becoming Compliant. That might be fine, I don't see where in the design we specified what would happen here. But whatever happens (compliant or unknown root status), I think we want to check for it here.

JustinKuli (commented):
Looks great overall! One design consideration came up while I was looking at the tests, other than clarifying that piece, I think this will be good to go

api/v1beta1/policyset_webhook.go
jan-law: Originally I used `Join`, but CodeRabbit suggested to return only one error :sweat_smile: ... I'll put the `errors.Join` back

test/e2e/case18_policyset_exclusions_test.go
jan-law: I keep changing my mind about whether the policy status should be Compliant or unknown. Now I'm thinking excluded clusters should report Compliant so that the overall compliance is green

JustinKuli: I can see why the root policy being overall Compliant would be nice. I don't know if that means that the excluded clusters should appear as Compliant in the root policy, I think that could be more confusing. I'd rather the excluded cluster stay blank/unknown, or have a special new state like `Excluded` (we wouldn't want that possible to use in other places, so I think this would require a new enum type).

I don't think it's too much of a problem as blank/unknown like it is right now. If a user saw that in the overall status, then dug into and saw the excluded clusters, I think they would understand. As long as we documented it.

It might also be possible to consider excluded clusters when doing `CalculateRootCompliance` without any changes inside the `status.status` list.

jan-law: Leaving status as Unknown. Will add documentation to the readme in a follow up PR

JustinKuli (approved):

PR #334, opened 2026-08-11
Check if replicated policy cache is stale

JustinKuli:
When a replicated policy is deleted, `ResourceVersions` is marked "deleted" so the resulting watch event can be skipped. Under rapid PlacementDecision churn, a follow-up reconcile can still get the deleted object from the controller-runtime cache, and conclude the policy already matches and skip recreating it.

Now, if ResourceVersions says the object was deleted, treat a cache hit as stale and try to recreate the policy. If it then finds that the policy already / still exists, clear the deleted marker and requeue so the next reconcile can handle it properly.

Refs:
 - https://issues.redhat.com/browse/ACM-37326

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of replicated policies that are deleted and recreated, ensuring stale cached data is refreshed correctly.
  * Prevented policies from remaining incorrectly marked as missing after recreation.
  * Improved detection of version-specific deletion events and recovery when policies are recreated.
  * Ensured deletion and recreation state is recorded consistently for more reliable policy propagation.

* **Tests**
  * Added coverage for stale cached policy state, deletion markers, missing entries, and valid resource versions.

### DIFF
--- a/controllers/propagator/replicatedpolicy_controller.go
+++ b/controllers/propagator/replicatedpolicy_controller.go
@@ -45,8 +45,10 @@ func (r *ReplicatedPolicyReconciler) Reconcile(ctx context.Context, request ctrl
 		hubTemplateActiveWatchesMetric.Set(float64(watchCount))
 	}()
 
+	staleCache := false
 	replicatedExists := true
 	replicatedPolicy := &policiesv1.Policy{}
+	rsrcVersKey := request.Namespace + "/" + request.Name
 
 	if err := r.Get(ctx, request.NamespacedName, replicatedPolicy); err != nil {
 		if !k8serrors.IsNotFound(err) {
@@ -55,6 +57,12 @@ func (r *ReplicatedPolicyReconciler) Reconcile(ctx context.Context, request ctrl
 			return reconcile.Result{}, err
 		}
 
+		replicatedExists = false
+	} else if cachedReplicatedPolicyIsStale(r.ResourceVersions, rsrcVersKey, replicatedPolicy.GetResourceVersion()) {
+		// The policy we got from the cache is the same one we deleted
+		log.V(1).Info("Cached replicated policy is stale after delete; treating as missing")
+
+		staleCache = true
 		replicatedExists = false
 	}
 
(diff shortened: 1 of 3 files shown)


controllers/propagator/replicatedpolicy_controller.go
```diff
@@ -283,7 +287,15 @@ func (r *ReplicatedPolicyReconciler) Reconcile(ctx context.Context, request ctrl
 		defer version.Unlock()
 
 		err = r.Create(ctx, desiredReplicatedPolicy)
-		if err != nil {
+		if k8serrors.IsAlreadyExists(err) {
+			// Another actor may have recreated the policy while ResourceVersions still said
+			// "deleted". Clear the marker so the next reconcile can trust a cache hit.
```
jan-law, line 276: Could you help me understand the cleanup paths in the reconciler? 
When the cache is stale with `replicatedExists = false`, and the root policy is disabled, does that mean we can safely assume the replicated policy is deleting and not recreated by another actor?

https://github.com/JustinKuli/governance-policy-propagator-1/blob/03dea2ae99bae739f1f8e4581045ad798f1d4e38/controllers/propagator/replicatedpolicy_controller.go#L138

```
if rootPolicy.Spec.Disabled {
		if replicatedExists {
			if err := r.cleanUpReplicated(ctx, replicatedPolicy); err != nil {
```

JustinKuli: That's a good point - I think I was only really considering a specific situation, and not thinking much about other "actors" getting in the way here.

If the root policy is disabled, and this controller deletes the replicated policy, but then it is re-created by *something else*, then I think the controller would incorrectly trust the `resourceVersions` cache, and would miss that it needs to re-delete that errant replicated policy.

I think we'll need something more robust to handle this - thanks for noticing that!

JustinKuli: The updated "deleted" marker, which now saves the resourceVersion that was deleted, should resolve the issue where *something else* created the replicated policy.

controllers/propagator/replicatedpolicy_setup.go
```diff
@@ -91,7 +92,7 @@ func replicatedPolicyPredicates(resourceVersions *sync.Map) predicate.Funcs {
 			version, loaded := safeReadLoad(resourceVersions, key)
 			defer version.RUnlock()
 
-			return !loaded || version.resourceVersion != "deleted"
+			return !loaded || !strings.HasPrefix(version.resourceVersion, "deleted")
```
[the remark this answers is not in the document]
JustinKuli: No, I think we want to skip a reconcile in this case whether it is `deleted` or `deleted-*`. There isn't a way for the value to be something invalid like `deleted123`.

jan-law (approved):
LGTM! Code coverage is sufficient

